PR #105, opened 2015-02-05
build_dependencies: added libsigscan to LIBYAL_LIBRARIES

ge0rg:
This adds libsigscan to the libraries that are built by build_dependencies, to fulfil dfvfs' requirements.


joachimmetz:
George thanks for the patch. Note that normally we use this process:
https://sites.google.com/a/kiddaland.net/plaso/developer/style-guide

Since the patch is small enough I'll merge it in.

PR #167, opened 2015-04-16
Update dependencies.py

Splinks:
Temporary fix that makes it installable, need to figure out which package is what on pypi in order to uncomment each dependency but this should be fine for installing the 'bleeding edge' version from the GIFT ppa atleast

### DIFF
--- a/plaso/dependencies.py
+++ b/plaso/dependencies.py
@@ -8,48 +8,48 @@ import urllib2
 # The tuple values are:
 # module_name, minimum_version
 LIBYAL_DEPENDENCIES = [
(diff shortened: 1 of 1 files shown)


joachimmetz:
Thanks for the contribution but FYI github PRs are not plaso coding practice. For more information see: https://github.com/log2timeline/plaso/wiki/Developers-Guide.

BTW Also it breaks the automated test and I just fixed the issue ;)
See: https://github.com/log2timeline/plaso/commit/994abf1ed4d26949535f86417068f9b20f766740

PR #328, opened 2015-09-09
Update winevt_rc.py

wdrone:
To fix crash issue related to line 90. When a condition uses a unicode character a sqlite3.OperationalError is raised crashing psort.py. I added the try/except block to keep things moving. In my version i printed the sql query to the terminal, however I think the gurus can come up with a more graceful error mechanism.

### DIFF
--- a/plaso/formatters/winevt_rc.py
+++ b/plaso/formatters/winevt_rc.py
@@ -87,7 +87,13 @@ class Sqlite3DatabaseFile(object):
     sql_query = u'SELECT {1:s} FROM {0:s}{2:s}'.format(
         u', '.join(table_names), u', '.join(column_names), condition)
 
-    self._cursor.execute(sql_query)
+    # TODO: Current code errors out when condition contains unicode char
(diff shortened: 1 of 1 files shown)


joachimmetz:
@wdrone thanks for the pull request but we don't use them for code merging purposes. Also see:
https://github.com/log2timeline/plaso/wiki/Codereview

Also by crash I assume that you mean the code raises an exception. I opt to file an issue instead and we'll have a look at the error. Seeing that you're not too confident about your solution.

joachimmetz:
I've filed https://github.com/log2timeline/plaso/issues/329 and will close this PR

wdrone:
I was unaware of your code merging procedures. As noted in my original request, the code raises a specific exception that is unhandled and ultimately causes the script to terminate prematurely. My solution was not really a solution but more of a field hot fix that allowed me to continue working. The purpose of my pull request was to convey the error and provide a temporary "keep it running" fix that could be reviewed by those such as yourself. Your process is different than I expected and thus filing an issue would have been a better course of action.

PR #742, opened 2016-06-24
301050043: Make the worker produce an event source, not a pathspec. Fixes crash in worker.

Onager:
[Code review: 301050043: Make the worker produce an event source, not a pathspec. Fixes crash in worker.](https://codereview.appspot.com/301050043/)


joachimmetz:
Was merged.

Issue #771, opened 2016-07-05
Improve task-based processing


joachimmetz:
Task-based processing changes to do after 1.5.0 release:

* [ ] implement task-based processing in psort
* [ ] improve performance
  * if there are no new event sources workers are idle. prioritize processing of directories first?

Unsorted:

* [ ] handle sigterm of main process as signal abort
* [ ] Make tracking of tasks persistent
* [ ] Fix https://github.com/log2timeline/plaso/issues/724
* [ ] Other
  * move remainder of collector, filter file, to worker?
  * Storage object change `SetEnableProfiling` to `EnableProfiling` and `DisableProfiling`
* [ ] Write source path spec information (mount points?)
* [ ] find a solution for the multi processing logger issue https://github.com/log2timeline/plaso/issues/460
* [ ] change event object to use event source
* [ ] look into making engine aware of worker abort
* [ ] worker: look into replacing "file_entry = None" with explicit invocation of GC
* [ ] revisit `_CalculateNTFSTimeHash`, is this approach sane? switch MD5 to SHA-256?
* [ ] clean up `PsortAnalysisReportZIPStorageFileWriter`
* [ ] merge `StorageFile` and `ZIPStorageFile`
* [ ] Clean up
  * ~~rename `AddEventObject` to `AddEvent`~~
  * ~~Rename `StorageObject` to `BaseStorage`~~
* [ ] Update docstrings according to https://github.com/log2timeline/plaso/issues/733
* [ ] Clean up ProcessSources method (Also see: https://codereview.appspot.com/301940043/diff/20001/plaso/frontend/extraction_frontend.py?context=10&column_width=80)
* [ ] Remove fake storage from preg (Also see: https://codereview.appspot.com/301940043/diff/20001/plaso/frontend/preg.py?context=10&column_width=80)
* [ ] Look at (https://codereview.appspot.com/301940043/diff/40001/tests/multi_processing/multi_process.py?context=10&column_width=80)

Also address:

* extraction frontend: add multi volume support #109
* Improve preprocess #241
* Provide a cleaner way to analyze a storage file #244 
* Modify the Parser Manager to allow for "always-on" parsers #269

Aftercare:

* [ ] fix newly spawned worker remaining idle after other working being killed
* [ ] prioritize to process directories first?
* [ ] clean up unit tests, make unit tests are testing units and move more end-to-end tests to the end-to-end testing 
* [ ] clean up event buffer, and different event heaps
[...]

PR #901, opened 2016-08-08
303250043: Worked on attribute container identifiers #771

joachimmetz:
[Code review: 303250043: Worked on attribute container identifiers #771](https://codereview.appspot.com/303250043/)


joachimmetz:
changes moved into other CLs

PR #1068, opened 2016-10-25
310600043: Add systemd journal parser

rgayon:
[Code review: 310600043: Add systemd journal parser](https://codereview.appspot.com/310600043/)


rgayon:
Currently generating subclasses of TextEvent, direclty using the dict passed from the messages in the journal. (hence using '_HOSTNAME' as a key in the formatter, for example).
This will need to be changed when the great event data refactor happens.

PR #1860, opened 2018-05-01
343850043: Added utility scripts to plot tasks profiling data

joachimmetz:
[Code review: 343850043: Added utility scripts to plot tasks profiling data](https://codereview.appspot.com/343850043/)

### DIFF
--- a/utils/plot-tasks.py
+++ b/utils/plot-tasks.py
@@ -1,0 +1,207 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# pylint: disable=invalid-name
+"""Script to plot tasks from profiling data.
+
+This script requires the matplotlib and numpy Python modules.
+"""
+
+from __future__ import print_function
+# mathplotlib does not support Unicode strings a column names.
+# from __future__ import unicode_literals
+
+import argparse
+import glob
+import math
+import os
+import sys
+
+import numpy
+
+from matplotlib import pyplot  # pylint: disable=import-error
+from numpy import genfromtxt  # pylint: disable=import-error
+
+
+class TaskMeasurements(object):
+  """Measurements of a task.
+
+  Attributes:
+    completed_time (float): time when the task was completed by the foreman.
+    created_time (float): time when the task was created by the foreman.
+    merging_duration (float): time it took the foreman to merge the task.
+    merging_time (float): time when the task was started to be merged by
+        the foreman.
+    pending_merge (float): time when the task was scheduled to be merged by
+        the foreman.
+    processed_time (float): time when the task was processed according to
+        the foreman.
+    processing_duration (float): time it took the worker to process the task.
+    processing_time (float): time when the task started to be processed by
+        the worker.
+    scheduled_time (float): time when the task was scheduled onto the task
+        queue by the foreman.
+  """
+
+  def __init__(self):
+    """Initializes a task measurement."""
+    super(TaskMeasurements, self).__init__()
+    self.completed_time = None
+    self.created_time = None
+    self.merging_duration = None
+    self.merging_time = None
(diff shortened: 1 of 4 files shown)


utils/plot-tasks.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# pylint: disable=invalid-name
```
Onager: Please rename to plot_tasks, and remove this pylint override:
https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming 
https://google.github.io/styleguide/pyguide.html?showone=Main#Main
"Even a file meant to be used as a script should be importable"

Yes, we have other files that aren't compliant at present, but we should fix those up.

utils/plot-tasks.py
```diff
@@ -0,0 +1,203 @@
...
+# mathplotlib does not support Unicode strings a column names.
+# from __future__ import unicode_literals
+
+import argparse
+import glob
+import os
+import sys
+
+from matplotlib import pyplot  # pylint: disable=import-error
+from numpy import genfromtxt  # pylint: disable=import-error
```
Onager: genfromtxt is a function, import the module or package instead: https://google.github.io/styleguide/pyguide.html?showone=Imports#Imports

utils/plot-tasks.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# pylint: disable=invalid-name
+"""Script to plot tasks from profiling data.
+
+This script requires the matplotlib and numpy Python modules.
+"""
+
+from __future__ import print_function
+# mathplotlib does not support Unicode strings a column names.
```
Onager: Isn't this just a problem when setting the column names in pyplot then? Using byte strings here is going to cause compatibility problems in the near future. Please reconsider this.

utils/plot-tasks.py
```diff
@@ -0,0 +1,203 @@
...
+    print('No such directory: {0:s}'.format(options.profile_path))
+    return False
+
+  names = ['time', 'identifier', 'status']
+
+  measurements = {}
+
+  glob_expression = os.path.join(options.profile_path, 'tasks-*.csv.gz')
+  for csv_file in glob.glob(glob_expression):
+    data = genfromtxt(
```
Onager: Please set the encoding here as well

utils/plot-tasks.py
```diff
@@ -0,0 +1,203 @@
...
+  if not os.path.isdir(options.profile_path):
+    print('No such directory: {0:s}'.format(options.profile_path))
+    return False
+
+  names = ['time', 'identifier', 'status']
+
+  measurements = {}
+
+  glob_expression = os.path.join(options.profile_path, 'tasks-*.csv.gz')
+  for csv_file in glob.glob(glob_expression):
```
Onager: rename to csv_file_name

utils/plot-tasks.py
joachimmetz: done

utils/plot-tasks.py
joachimmetz: ack

utils/plot-tasks.py
joachimmetz: done

utils/plot-tasks.py
joachimmetz: done

utils/plot-tasks.py
joachimmetz: if I recall correctly, bytes column names cause issues for python3 and unicode string for python2. For now it is not problematic if these scripts only support 1 python version.

Issue #1893, opened 2018-05-19
Migrate construct-based parsers to use dtfabric


joachimmetz:
Migrate construct-based parsers to use dtfabric. This will improve Python 3 compatibility.

Also see: https://github.com/log2timeline/plaso/projects/14
* [x] ~~chrome_cache~~
  * ~~generate specimens (https://github.com/dfirlabs/crhome-specimens)~~
  * ~~https://github.com/log2timeline/plaso/pull/2075~~
* [x] ~~mac_keychain~~
  * ~~construct parser has limited format support~~
  * ~~analyse format: https://github.com/libyal/dtformats/blob/master/documentation/MacOS%20keychain%20database%20file%20format.asciidoc~~
* [x] ~~systemd_journal~~
* [x] ~~asl~~
  * ~~https://codereview.appspot.com/347900043/~~
  * ~~also see: https://github.com/log2timeline/plaso/issues/530~~
* [x] ~~bsm~~
  * ~~https://codereview.appspot.com/359740043/~~
  * ~~dtfabric update https://github.com/log2timeline/l2tdevtools/issues/419~~
  * https://github.com/log2timeline/plaso/pull/1972
* [x] ~~cups_ipp~~
  * ~~https://codereview.appspot.com/347800043/~~
* [x] ~~firefox_cache~~
  * ~~generate specimens (https://github.com/dfirlabs/firefox-specimens)~~
  * ~~also address https://github.com/log2timeline/plaso/issues/2066~~
  * https://github.com/log2timeline/plaso/pull/2070
* [x] ~~fseventsd~~
  * ~~https://github.com/log2timeline/plaso/issues/1925~~
  * ~~https://github.com/log2timeline/plaso/pull/2032~~
* [x] ~~java_idx~~
  * ~~https://codereview.appspot.com/346090043~~
* [x] ~~ntfs~~
  * ~~https://codereview.appspot.com/341360044~~
* [x] ~~pls_recall~~
  * ~~https://codereview.appspot.com/347890043~~
* [x] ~~recycler~~
  * ~~https://codereview.appspot.com/348940043/~~
  * ~~current parser has a bug? that unsets the ANSI original filename: https://github.com/log2timeline/plaso/blob/12454e448cd807b464ba1e6dbcf7c476c4558839/plaso/parsers/recycler.py#L190~~
* [x] ~~safari_cookies~~
    * ~~https://codereview.appspot.com/348840043/~~
* [x] ~~utmp~~
[...]

PR #1915, opened 2018-05-27
348840043: Migrated Safari cookies construct-based parser to use dtfabric #1893

joachimmetz:
[Code review: 348840043: Migrated Safari cookies construct-based parser to use dtfabric #1893](https://codereview.appspot.com/348840043/)

### DIFF
--- a/plaso/parsers/safari_cookies.py
+++ b/plaso/parsers/safari_cookies.py
@@ -93,12 +93,11 @@ class BinaryCookieParser(data_formats.DataFormatParser):
     self._cookie_plugins = (
         cookie_plugins_manager.CookiePluginsManager.GetPlugins())
 
-  def _ParsePage(self, parser_mediator, page_number, page_data):
+  def _ParsePage(self, parser_mediator, page_data):
     """Parses a page.
 
     Args:
       parser_mediator (ParserMediator): parser mediator.
-      page_number (int): page number.
       page_data (bytes): page data.
 
     Raises:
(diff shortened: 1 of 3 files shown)


plaso/parsers/safari_cookies.py
```diff
@@ -43,104 +43,127 @@ def __init__(self):
...
-  _PAGE_HEADER = construct.Struct(
-      'page_header',
-      construct.Bytes('header', 4),
-      construct.ULInt32('number_of_records'),
-      construct.Array(
-          lambda ctx: ctx.number_of_records,
-          construct.ULInt32('offsets')))
+  _DEFINITION_FILE = 'safari_cookies.yaml'
+
+  _SIGNATURE = b'cook'
```
Onager: Add the null byte here as well, to match the format specification

plaso/parsers/safari_cookies.yaml
```diff
@@ -0,0 +1,121 @@
...
+  number_of_elements: 4
+- name: number_of_pages
+  data_type: uint32
+---
+name: binarycookies_page_sizes
+type: sequence
+element_data_type: uint32be
+number_of_elements: binarycookies_file_header.number_of_pages
+---
+name: binarycookies_file_footer
```
Onager: Please remove the footer, as it's not used

plaso/parsers/safari_cookies.py
joachimmetz: no the signature is 4 bytes the 0-byte in the format spec is there to reduce false positives

plaso/parsers/safari_cookies.yaml
joachimmetz: ack

PR #1934, opened 2018-06-08
Improved python3 compatibility of lib tests #1932

Onager:
Improved python3 compatibility of lib tests #1932

### DIFF
--- a/plaso/lib/binary.py
+++ b/plaso/lib/binary.py
@@ -78,7 +78,7 @@ def ReadUTF16Stream(file_object, offset=None, byte_size=0):
   """Reads an UTF-16 formatted stream from a file-like object.
 
   Reads an UTF-16 formatted stream that's terminated by
-  an end-of-string terminator (\x00\x00) or up to the byte size.
+  an end-of-string character (\x00\x00) or up to the byte size.
 
   Args:
     file_object (file): file-like object to read the data from.
@@ -239,16 +239,16 @@ def ReadUTF16(string_buffer):
       use_buffer, 'utf-16', errors='ignore').replace('\x00', '')
 
 
-def HexifyBuffer(byte_stream):
-  """Returns an hexadecimal representation of a byte stream.
+def HexifyBuffer(byte_sequence):
+  """Returns an hexadecimal representation of a byte sequence.
 
   Args:
-    string_buffer (bytes): byte string.
+    byte_sequence (bytes): byte sequence.
 
   Returns:
     str: hexadecimal representation of the byte stream.
   """
-  hex_bytes = codecs.encode(byte_stream, 'hex')
+  hex_bytes = codecs.encode(byte_sequence, 'hex')
   output_string = codecs.decode(hex_bytes, 'utf-8')
   string_iterators = [iter(output_string)] * 2
 
@@ -279,8 +279,8 @@ def _StreamContainsUTF16NullTerminator(byte_stream, offset):
   """
   byte_1 = byte_stream[offset]
   byte_2 = byte_stream[offset + 1]
-  if py2to3.PY_2 and (byte_1 == b'\x00' and byte_2 == b'\x00'):
+  if py2to3.PY_2 and byte_1 == b'\x00' and byte_2 == b'\x00':
     return True
-  if py2to3.PY_3 and (byte_1 == 0 and byte_2 == 0):
+  if py2to3.PY_3 and byte_1 == 0 and byte_2 == 0:
     return True
   return False
--- a/plaso/lib/binary.py
+++ b/plaso/lib/binary.py
@@ -3,7 +3,8 @@
 
 from __future__ import unicode_literals
 
-import binascii
+import codecs
+import itertools
 import logging
 import os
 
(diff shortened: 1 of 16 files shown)


plaso/lib/objectfilter.py
```diff
@@ -112,8 +113,15 @@ def __init__(self, brand, code):
 
 def GetUnicodeString(string):
   """Converts the string to Unicode if necessary."""
```
joachimmetz: add args section, also maybe rename string argument to value?

plaso/lib/utils.py
```diff
@@ -21,7 +21,7 @@ def IsText(bytes_in, encoding=None):
   sequence is not text, but a byte sequence.
 
   Args:
-    bytes_in (bytes): byte stream to examine.
+    bytes_in (bytes|str): byte stream to examine.
```
joachimmetz: why add str?

plaso/lib/utils.py
```diff
@@ -32,19 +32,21 @@ def IsText(bytes_in, encoding=None):
   # Start with the assumption we are dealing with a text.
   is_ascii = True
 
+  if isinstance(bytes_in, py2to3.UNICODE_TYPE):
+    return is_ascii
```
joachimmetz: lets rename `is_ascii` to `is_text` seeing an ASCII Unicode string does not make sense to me

tests/lib/binary.py
```diff
@@ -172,26 +172,26 @@ def testArrayOfUTF16StreamCopyToStringTable(self):
       self.assertEqual(string_table, expected_string_table)
 
   def testStringParsing(self):
```
joachimmetz: name of function and docstring don't seem to match what is being tested

plaso/lib/utils.py
joachimmetz: or define `is_ascii` after this check and return True

plaso/lib/binary.py
```diff
@@ -28,44 +29,44 @@ def ByteStreamCopyToString(byte_stream, codepage='utf-8'):
...
   try:
-    string = byte_stream.decode(codepage)
+    string = codecs.decode(byte_stream, codepage)
   except UnicodeDecodeError:
     logging.warning(
         'Unable to decode {0:s} formatted byte stream.'.format(codepage))
-    string = byte_stream.decode(codepage, errors='ignore')
+    string = codecs.decode(byte_stream, codepage, errors='ignore')
 
   string, _, _ = string.partition('\x00')
```
joachimmetz: Wasn't split faster? Maybe change to `string = string.split('\x00')[0]`

plaso/lib/objectfilter.py
```diff
@@ -110,11 +111,25 @@ def __init__(self, brand, code):
...
-def GetUnicodeString(string):
-  """Converts the string to Unicode if necessary."""
-  if not isinstance(string, py2to3.UNICODE_TYPE):
-    return str(string).decode('utf8', errors='ignore')
-  return string
+def GetUnicodeString(value):
+  """Attempts to convert the argument to a Unicode string.
+  
+  Args:
+    value (list|int|bytes): value to convert.
```
joachimmetz: add `|str`

plaso/lib/utils.py
joachimmetz: Ack, keep it as-is. I think `bytes_in` is what initially confused me.

joachimmetz:
Some linter issues as well: https://travis-ci.org/log2timeline/plaso/jobs/390859521

joachimmetz:
> "byte stream" is still strange to me - see https://en.wikipedia.org/wiki/Stream_(computing), which
> mentions in part that "a stream is a sequence of data elements made available over time". This matches
> my understanding, whereas the data in the "streams" mentioned in binary.py aren't made available over
> time, but all at once.

A different Wikipedia article says: https://en.wikipedia.org/wiki/Bitstream

```
A bytestream is a sequence of bytes. 
```

This says nothing about the availability of the bytes. Then again https://en.wikipedia.org/wiki/Reliable_byte_stream says:

```
A reliable byte stream is a common service paradigm
```

Based on that I would say yeah to more ambiguity ;)

If I look at:
https://docs.oracle.com/javase/tutorial/essential/io/bytestreams.html
https://msdn.microsoft.com/en-us/library/system.io.memorystream(v=vs.110).aspx

I would conclude that the term seems to be generally used for a sequence of bytes, backed by what ever form of medium or availability.

> I'd prefer to remove the usages of "stream" here, but you seem to feel strongly the other way.

I opt to do this in a different CL if you feel strongly about this. As I indicated, this code is likely to go away anyway with dtFabric. But please use something that is not `string` ;)

plaso/lib/objectfilter.py
Onager: Done

plaso/lib/binary.py
```diff
@@ -211,31 +210,76 @@ def ArrayOfUTF16StreamCopyToStringTable(byte_stream, byte_stream_size=None):
...
+  # pylint: disable=no-member
+  if py2to3.PY_2:
+    iterators = itertools.izip_longest(*string_iterators)
+  else:
+    iterators = itertools.zip_longest(*string_iterators)
+  groups = list(iterators)
+  output_string = ''.join(
+      ['\\x{0:s}{1:s}'.format(group[0], group[1]) for group in groups])
+  return output_string
+
```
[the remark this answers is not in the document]
Onager: Done

plaso/lib/binary.py
Onager: Split() is only faster for large strings, I think this is better as partition()

joachimmetz:
Feel free to merge once tests pass

Issue #1947, opened 2018-06-13
Ensure engine tests pass under Python 3


Onager:

Onager:
Fixed some initial incompatibilities, but it's currently hard to see what's broken due to the tests in worker.py relying on parser tests working. This is blocked on #1952

PR #1955, opened 2018-06-13
Improved compatibility of engine tests with Python 3 #1947

Onager:
Improved compatibility of engine tests with Python 3 #1947

### DIFF
--- a/plaso/engine/profilers.py
+++ b/plaso/engine/profilers.py
@@ -291,4 +291,3 @@ class TasksProfiler(SampleFileProfiler):
     sample = '{0:f}\t{1:s}\t{2:s}\n'.format(
         sample_time, task.identifier, status)
     self._Write(sample)
-
--- a/plaso/engine/profilers.py
+++ b/plaso/engine/profilers.py
@@ -44,9 +44,12 @@ class CPUTimeMeasurement(object):
 class SampleFileProfiler(object):
   """Shared functionality for sample file-based profilers."""
 
-  _FILENAME_PREFIX = ''
 
-  _FILE_HEADER = ''
+  # This constant must be overriden by subclasses.
+  _FILENAME_PREFIX = None
+
+  # This constant must be overriden by subclasses.
+  _FILE_HEADER = None
 
   def __init__(self, identifier, configuration):
     """Initializes a sample file profiler.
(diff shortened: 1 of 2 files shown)


plaso/engine/profilers.py
```diff
@@ -73,13 +85,13 @@ def IsSupported(cls):
 
   def Start(self):
     """Starts the profiler."""
-    filename = '{0:s}-{1:s}.csv.gz'.format(
+    filename = '{0!s}-{1:s}.csv.gz'.format(
```
joachimmetz: Why support anything else than a string here?

plaso/engine/profilers.py
```diff
@@ -62,6 +65,15 @@ def __init__(self, identifier, configuration):
     self._sample_file = None
     self._start_time = None
 
+  def _Write(self, content):
```
joachimmetz: Unclear from the method name what is written, maybe rename to `_WriteSample` or `_WriteString`

Onager: Done

plaso/engine/profilers.py
Onager: Changed the initialization of _FILENAME_PREFIX instead

plaso/engine/profilers.py
```diff
@@ -43,12 +44,14 @@ def SampleStop(self):
 class SampleFileProfiler(object):
   """Shared functionality for sample file-based profilers."""
 
-  _FILENAME_PREFIX = None
+  _FILENAME_PREFIX = ''
```
joachimmetz: I opt to change this back to None and add a comment that this value should be overwritten by a subclass.

plaso/engine/profilers.py
```diff
@@ -62,6 +65,15 @@ def __init__(self, identifier, configuration):
     self._sample_file = None
     self._start_time = None
 
+  def _WritesString(self, content):
+    """Writes a string to the sample file.
+
+    Args:
+      content (str): content to write to the sample file.
+    """
+    sample_bytes = codecs.encode(content, 'utf-8')
```
joachimmetz: Since the docstring nor the argument uses sample, using sample_bytes is confusing to read.
I opt to use sample instead of content and rename the method to WriteSample as well.

Onager: This is also used for writing the header, so WriteSample doesn't work.

plaso/engine/profilers.py
Onager: Done, though the assigning to protected constants in the test doesn't feel good.

Issue #1950, opened 2018-06-13
Ensure multi_processing tests pass under Python 3


Onager:

Onager:
The worker_process StartAndStopProfiling test needs #1955 to pass.

PR #1956, opened 2018-06-13
Improved compatibility of multiprocessing tests with Python 3 #1950

Onager:
Improved compatibility of multiprocessing tests with Python 3 #1950

### DIFF
--- a/plaso/multi_processing/task_manager.py
+++ b/plaso/multi_processing/task_manager.py
@@ -179,6 +179,8 @@ class TaskManager(object):
       inactive_time = time.time() - self._TASK_INACTIVE_TIME
       inactive_time = int(inactive_time * definitions.MICROSECONDS_PER_SECOND)
 
+      # Abandon all tasks after they're identified so as not to modify the
+      # dict while iterating over it.
       tasks_to_abandon = []
       for task_identifier, task in iter(self._tasks_processing.items()):
         if task.last_processing_time < inactive_time:
@@ -188,7 +190,7 @@ class TaskManager(object):
           self.SampleTaskStatus(task, 'abandoned_processing')
           tasks_to_abandon.append((task_identifier, task))
 
-      for (task_identifier, task) in tasks_to_abandon:
+      for task_identifier, task in tasks_to_abandon:
         self._tasks_abandoned[task_identifier] = task
         del self._tasks_processing[task_identifier]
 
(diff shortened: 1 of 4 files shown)


plaso/multi_processing/task_manager.py
```diff
@@ -188,7 +190,7 @@ def _AbandonInactiveProcessingTasks(self):
           self.SampleTaskStatus(task, 'abandoned_processing')
           tasks_to_abandon.append((task_identifier, task))
 
-      for (task_identifier, task) in tasks_to_abandon:
+      for task_identifier, task in tasks_to_abandon:
```
joachimmetz: Could you alternatively use dict.pop() inside the previous loop?

tests/multi_processing/task_manager.py
```diff
@@ -205,6 +205,7 @@ def testUpdateLatestProcessingTime(self):
     """Tests the _UpdateLatestProcessingTime function."""
     manager = task_manager.TaskManager()
     task = manager.CreateTask(self._TEST_SESSION_IDENTIFIER)
+    task.last_processing_time = 0
```
joachimmetz: Why this change? is this a left over that could be removed?

tests/multi_processing/psort.py
```diff
@@ -513,7 +514,8 @@ def testExportEvents(self):
 
     lines = []
     output = output_writer.ReadOutput()
-    for line in output.split(b'\n'):
+    output = codecs.decode(output, 'utf-8')
```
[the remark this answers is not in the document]
joachimmetz: or implement a string output writer for testing?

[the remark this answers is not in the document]
Onager: #1963 for that

tests/multi_processing/task_manager.py
Onager: This is because the UpdateLatestProcessingTime() method compares this field against the latest_processing_time value, and Python3 errors on comparing None and int.

joachimmetz: Should we update _UpdateLatestProcessingTime instead? I assume this value being None will happen runtime as well

Onager: You requested that I not change that method ;) https://github.com/log2timeline/plaso/pull/1956/files/6b89d53e2d7c8e7089459e271c518d9e216ed853#diff-dce07e43bc91fcf6c32c3d51d2590951R254

No, this doesn't happen at runtime, because UpdateProcessingTime() is called before this method is invoked.

Issue #1951, opened 2018-06-13
Ensure output tests pass under Python 3


Onager:

Onager:
With recent updates to parsers, output tests now pass.

PR #1957, opened 2018-06-13
Improved compatibility of output tests with Python 3 #1951

Onager:
Improved compatibility of output tests with Python 3 #1951

### DIFF
--- a/tests/cli/test_lib.py
+++ b/tests/cli/test_lib.py
@@ -30,7 +30,7 @@ class TestOptions(object):
   """Class to define test options."""
 
 
-class TestOutputWriter(tools.FileObjectOutputWriter):
+class TestBinaryOutputWriter(tools.FileObjectOutputWriter):
   """Class that implements a test output writer."""
 
   def __init__(self, encoding='utf-8'):
@@ -39,23 +39,15 @@ class TestOutputWriter(tools.FileObjectOutputWriter):
     Args:
       encoding (Optional[str]): output encoding.
     """
-    file_object = io.StringIO()
-    super(TestOutputWriter, self).__init__(file_object, encoding=encoding)
+    file_object = io.BytesIO()
+    super(TestBinaryOutputWriter, self).__init__(file_object, encoding=encoding)
     self._read_offset = 0
-
-  def Write(self, string):
-    """Writes a string to the output.
-
-    Args:
-      string (str): output.
-    """
-    self._file_object.write(string)
 
   def ReadOutput(self):
     """Reads all output added since the last call to ReadOutput.
 
     Returns:
-      str: output data.
+      bytes: output data.
     """
     self._file_object.seek(self._read_offset, os.SEEK_SET)
     output_data = self._file_object.read()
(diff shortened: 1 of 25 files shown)


tests/serializer/json_serializer.py
```diff
@@ -149,6 +149,7 @@ def testReadAndWriteSerializedEventObject(self):
     expected_event_object.empty_string = ''
     expected_event_object.zero_integer = 0
     expected_event_object.integer = 34
+    expected_event_object.float = -122.082203542683
```
joachimmetz: maybe rename event_object to event as well

tests/cli/test_lib.py
```diff
@@ -55,6 +56,19 @@ def ReadOutput(self):
 
     return output_data
 
+  def ReadOutputString(self):
```
joachimmetz: Why would you compare against string when what the original intent was to compare against the resulting encoded data.

tests/output/kml.py
```diff
@@ -28,35 +28,35 @@ def setUp(self):
...
+    footer = self._output_writer.ReadOutputString()
     self.assertEqual(footer, expected_footer)
 
   def testWriteEventBody(self):
     """Tests the WriteEventBody function."""
 
     # Test event object without geo-location.
     self._output_module.WriteEventBody(self._event_object)
-    self.assertEqual(self._output_writer.ReadOutput(), b'')
+    self.assertEqual(self._output_writer.ReadOutputString(), '')
```
joachimmetz: suggestion: maybe break this up in
```
event_body = self._output_writer.ReadOutputString()
self.assertEqual(event_body, '')
```

so that it is more clear that we are testing event body

Onager: Done

tests/serializer/json_serializer.py
Onager: Done

tests/cli/test_lib.py
Onager: This class has a Write() method that takes a string, so it seems to me that this class deals with strings, not bytes. 

I don't think it's desirable to have output tests being dependent on the implementation of the writer class.

Test should need to care that the strings they expect are the ones they get, not how the strings are encoded when output. That sort of testing belongs in test for the writers themselves.

joachimmetz: in that case maybe we should implement a test string output writer instead of encoding and decoding back and forth?

Onager: Opened #1963 to track that

tests/cli/test_lib.py
```diff
@@ -55,6 +56,20 @@ def ReadOutput(self):
 
     return output_data
 
+  # TODO: refactor this method https://github.com/log2timeline/plaso/issues/1963
```
joachimmetz: On a second pass the intent of this class is to write output including encoding seeing some output format might specify a specific encoding. So that output is consistent cross-platform. Please undo these changes, seeing the no longer reflect what needs to be tested.

Onager: Per conversation, I've changed the TestOutputWriter to just deal with strings, with no conversion between outputs and test fixtures (expected output).

Issue #1952, opened 2018-06-13
Ensure parser tests pass under Python 3


Onager:

Onager:
Currently still failing:
bsm
chrome_cache
firefox_cache
mac_keychain
pe (dependency issue, needs update of pefile)
appcompatcache

PR #1966, opened 2018-06-18
Improve compatibility of dsv parsers with Python3 #1952

Onager:
Improve compatibility of dsv parsers with Python3 #1952

### DIFF
--- a/plaso/parsers/dsv_parser.py
+++ b/plaso/parsers/dsv_parser.py
@@ -7,6 +7,8 @@ import abc
 import csv
 import sys
 
+from dfvfs.helpers import text_file
+
 from plaso.lib import errors
 from plaso.lib import line_reader_file
 from plaso.lib import py2to3
@@ -87,23 +89,18 @@ class DSVParser(interface.FileObjectParser):
 
     return row
 
-  def _CreateDictReader(self, parser_mediator, line_reader):
+  def _CreateDictReader(self, line_reader):
     """Returns a reader that processes each row and yields dictionaries.
 
     csv.DictReader does this job well for single-character delimiters; parsers
     that need multi-character delimiters need to override this method.
 
     Args:
-      parser_mediator (ParserMediator): mediates interactions between parsers
-          and other components, such as storage and dfvfs.
       line_reader (iter): yields lines from a file-like object.
 
     Returns:
       iter: a reader of dictionaries, as returned by csv.DictReader().
     """
-    if not self._encoding:
-      self._encoding = parser_mediator.codepage
-
     delimiter = self.DELIMITER
     quotechar = self.QUOTE_CHAR
     magic_test_string = self._MAGIC_TEST_STRING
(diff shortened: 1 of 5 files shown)


plaso/parsers/dsv_parser.py
```diff
@@ -147,13 +144,19 @@ def ParseFileObject(self, parser_mediator, file_object, **unused_kwargs):
           '[{0:s}] Unable to parse DSV file: {1:s} size of file exceeds '
           'maximum supported size').format(self.NAME, display_name))
 
-    line_reader = line_reader_file.BinaryLineReader(file_object)
+    if not self._encoding:
+      self._encoding = parser_mediator.codepage
```
joachimmetz: I opt not to set `self._encoding` here, seeing it changes the state of the object, maybe define encoding as a local variable instead?

```
encoding = self._encoding or parser_mediator.codepage
```

plaso/parsers/dsv_parser.py
```diff
@@ -147,13 +144,19 @@ def ParseFileObject(self, parser_mediator, file_object, **unused_kwargs):
...
+
+    if py2to3.PY_3:
+      line_reader = text_file.TextFile(file_object, encoding=self._encoding)
+    else:
+      line_reader = line_reader_file.BinaryLineReader(file_object)
```
joachimmetz: why not use binary line reader for both python 2 and 3?

plaso/parsers/mactime.py
```diff
@@ -26,7 +26,7 @@ class MactimeEventData(events.EventData):
     filename (str): name of the file.
     inode (int): "inode" of the file. Note that inode is an overloaded term
         in the context of mactime and used for MFT entry index values as well.
-    md5_hash (str): MD5 hash of the file content.
+    md5 (str): MD5 hash of the file content.
```
joachimmetz: maybe change to:
MD5 hash of the file content, formatted as a hexadecimal string.

plaso/parsers/mcafeeav.py
```diff
@@ -137,8 +139,10 @@ def VerifyRow(self, parser_mediator, row):
     # This file can have a UTF-8 byte-order-marker at the beginning of
     # the first row.
     # TODO: Find out all the code pages this can have.  Asked McAfee 10/31.
-    if row['date'][0:3] == b'\xef\xbb\xbf':
-      row['date'] = row['date'][3:]
+    byte_order_mark = codecs.decode(b'\xef\xbb\xbf', parser_mediator.codepage)
```
joachimmetz: please do not change b'\xef\xbb\xbf' is a sequence specific to UTF-8 (UTF-8 BOM, which technically does not need a BOM ;), this is the binary sequence and should not be translated into a string. Seeing that the string u'\ufeff' can be used for UTF-16 and UTF-32 as well.

joachimmetz:
@Onager please address comments from previous review first

plaso/parsers/dsv_parser.py
Onager: I agree with what you're getting at here, but this is currently needed to support the way some parsers work, notably McAfee. Filed #1971 for this.

PR #1977, opened 2018-06-23
Improved compatibility of sqlite parser and plugins with Python 3 #1952

Onager:
Improved compatibility of sqlite parser and plugins with Python 3 #1952


plaso/parsers/cookie_plugins/ganalytics.py
```diff
@@ -410,29 +410,27 @@ def GetEntries(
...
-        try:
-          value_line = utf_stream.decode('utf-8')
-        except UnicodeDecodeError:
-          value_line = utf_stream.decode('utf-8', errors='replace')
-          parser_mediator.ProduceExtractionError(
-              'Cookie value did not decode to Unicode string. Non UTF-8 '
-              'characters have been replaced.')
-
-        extra_attributes[key] = value_line
+        if isinstance(value, py2to3.UNICODE_TYPE) and py2to3.PY_2:
```
joachimmetz: if cookie_data is str why do we need to have this check here?

plaso/parsers/cookie_plugins/ganalytics.py
```diff
@@ -355,7 +355,7 @@ def GetEntries(
 
     Args:
       parser_mediator (ParserMediator): parser mediator.
-      cookie_data (bytes): cookie data.
+      cookie_data (str): cookie data.
```
joachimmetz: if cookie_data is str why do we need to decode it in this function?

plaso/parsers/sqlite_plugins/skype.py
```diff
@@ -446,7 +446,7 @@ def ParseAccountInformation(
 
     display_name = self._GetRowValue(query_hash, row, 'given_displayname')
     fullname = self._GetRowValue(query_hash, row, 'fullname')
-    username = '{0:s} <{1:s}>'.format(fullname, display_name)
+    username = '{0!s} <{1!s}>'.format(fullname, display_name)
```
joachimmetz: why the change, what other type can fullname, display_name be?

tests/parsers/sqlite_plugins/interface.py
```diff
@@ -78,7 +78,7 @@ def ParseMyTableRow(self, parser_mediator, query, row, **unused_kwargs):
 
     # If Python 2 is used field3 needs to be converted to a string
     # because it is a read-write buffer.
-    if sys.version_info[0] < 3:
+    if py2to3.PY_2 and field3:
```
joachimmetz: `and field3 is not None` because `''` should be converted. Do you plan to move this into interface.py (under plaso) ? If so add a TODO and link to the issue.

tests/parsers/sqlite_plugins/skype.py
```diff
@@ -37,14 +38,15 @@ def testProcess(self):
     storage_writer = self._ParseDatabaseFileWithPlugin(
         ['skype_main.db'], plugin)
 
-    self.assertEqual(storage_writer.number_of_events, 24)
+    # self.assertEqual(storage_writer.number_of_events, 24)
```
joachimmetz: Why comment this out?

Onager: Debugging artifact

plaso/parsers/sqlite_plugins/skype.py
Onager: They can be None.

tests/parsers/sqlite_plugins/interface.py
Onager: Done,

plaso/parsers/cookie_plugins/ganalytics.py
Onager: We need to convert the unicode object to python2 str to pass to the Python 2 version of unquote().

plaso/parsers/cookie_plugins/ganalytics.py
Onager: See comment below.

plaso/parsers/cookie_plugins/ganalytics.py
joachimmetz: maybe add a comment, to remind ourselves why we do this like this.

plaso/parsers/sqlite_plugins/skype.py
joachimmetz: Seeing this is a user represented string I opt changing this logic to represent username as:
```
if fullname and display_name:
  username = '{0:s} <{1:s}>'.format(fullname, display_name)
elif fullname:
  username = fullname
else:
  username = display_name
```

Onager: This changes the format of the field though, I'm not keen on doing this. Display name is already set as a separate attribute in the event data.

plaso/parsers/cookie_plugins/ganalytics.py
Onager: Done.

tests/parsers/sqlite_plugins/interface.py
Onager: This is actually done now.

plaso/parsers/cookie_plugins/ganalytics.py
```diff
@@ -410,29 +410,30 @@ def GetEntries(
...
-        try:
-          value_line = utf_stream.decode('utf-8')
-        except UnicodeDecodeError:
-          value_line = utf_stream.decode('utf-8', errors='replace')
-          parser_mediator.ProduceExtractionError(
-              'Cookie value did not decode to Unicode string. Non UTF-8 '
-              'characters have been replaced.')
-
-        extra_attributes[key] = value_line
+        # Urllib2 in Python2 requires a str argument, not unicode. We thus need
```
joachimmetz: please don't use `str` it is ambiguous I opt to use bytes.

unicode => Unicode

plaso/parsers/cookie_plugins/ganalytics.py
```diff
@@ -410,29 +410,30 @@ def GetEntries(
...
+
+        value = urlparse.unquote(value)
+
+        if py2to3.PY_2:
+          try:
+            value = codecs.encode(value, 'utf-8')
+          except UnicodeDecodeError:
+            value = codecs.encode(value, 'utf-8', errors='replace')
+            parser_mediator.ProduceExtractionError(
+                'Cookie value did not decode to Unicode string. Non UTF-8 '
```
joachimmetz: decode => encode ?

plaso/parsers/sqlite_plugins/skype.py
joachimmetz: But results like `"Name <None>"` look silly and are not useful IMHO

tests/parsers/sqlite_plugins/skype.py
```diff
@@ -21,9 +21,10 @@ def testProcess(self):
     """Tests the Process function on a Skype History database file.
 
       The History file contains 24 events:
-      4 call events
+      3 call events
```
joachimmetz: Any indication of why this changed?

plaso/parsers/cookie_plugins/ganalytics.py
Onager: These are both the built-in type names from Python 2. I've added some quote marks to make this clear.

plaso/parsers/cookie_plugins/ganalytics.py
Onager: Changed this. decode/encode is a little confusing here.

plaso/parsers/sqlite_plugins/skype.py
Onager: I don't agree, I think it's more obvious to the user that there's something strange about that row.

tests/parsers/sqlite_plugins/skype.py
Onager: As far as I can tell, this comment was never correct.

plaso/parsers/sqlite_plugins/skype.py
joachimmetz: Ack, can we add a comment to move this to the formatter and let's chat how to represent values that are not set.

joachimmetz:
@Onager please update branch and make the CI tests pass.

PR #2009, opened 2018-07-04
367740043: Migrated winreg shutdown construct-based plugin to use dtfabric #1893

joachimmetz:
[Code review: 367740043: Migrated winreg shutdown construct-based plugin to use dtfabric #1893](https://codereview.appspot.com/367740043/)

### DIFF
--- a/plaso/parsers/winreg_plugins/shutdown.py
+++ b/plaso/parsers/winreg_plugins/shutdown.py
@@ -3,15 +3,15 @@
 
 from __future__ import unicode_literals
 
-import construct
-
 from dfdatetime import filetime as dfdatetime_filetime
 from dfdatetime import semantic_time as dfdatetime_semantic_time
 
 from plaso.containers import events
 from plaso.containers import time_events
 from plaso.lib import definitions
+from plaso.lib import errors
 from plaso.parsers import winreg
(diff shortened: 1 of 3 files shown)


plaso/parsers/winreg_plugins/shutdown.py
```diff
@@ -43,7 +44,39 @@ class ShutdownPlugin(interface.WindowsRegistryPlugin):
...
+    try:
+      filetime = self._ReadStructureFromByteStream(
+          byte_stream, 0, filetime_map)
+    except (ValueError, errors.ParseError) as exception:
+      raise errors.ParseError(
+          'Unable to parse FILETIME value with error: {0!s}'.format(
+              exception))
+
+    if filetime == 0:
+      return None
```
Onager: Why not return the not-set semantic time here instead of returning None?

joachimmetz: I considered this but was thinking that I would like to be able to tell when a filetime could not be parsed. For this specific parser is does not matter too much but e.g. in case if more than 1 filetime value would be parsed.

joachimmetz:
Replied to question did not make changes.

Issue #386, opened 2015-10-19
pcap: parser rewrite to fix high memory consumption


joachimmetz:
As indicated in: https://github.com/log2timeline/plaso/issues/360

* [ ] rewrite pcap parser to reduce memory consumption
* [ ] re-enable pcap parser after rewrite
* [ ] test running on Ubuntu 14.04 live VM 2G RAM with http://download.netresec.com/pcap/maccdc-2012/maccdc2012_00016.pcap.gz

joachimmetz:
Issue not touched in multiple years, removing PCAP parser seeing it is triggereing timestamp out of bounds

PR #2026, opened 2018-07-08
Removed PCAP parser #386

joachimmetz:
Removed PCAP parser #386


joachimmetz:
Added removal of dpkt to https://github.com/log2timeline/plaso/pull/2022

PR #2032, opened 2018-07-10
 Migrated fsevents parser to dtfabric. Resolves #1925

### LINKED ISSUE #1925: Migrate fsevents parser to dtfabric

Onager:
Migrated fsevents parser to dtfabric. Resolves #1925


joachimmetz:
@Onager missing assignee ?

plaso/parsers/fseventsd.py
```diff
@@ -89,28 +72,39 @@ def GetFormatSpecification(cls):
     format_specification.AddNewSignature(cls._DLS_V2_SIGNATURE, offset=0)
     return format_specification
 
-  def _ParseDLSPageHeader(self, file_object):
+  def _ParseDLSPageHeader(self, file_object, offset):
     """Parses a DLS page header from a file-like object.
 
     Args:
       file_object (file): file-like object to read the header from.
+      offset (int): offset of the beginning of the header.
```
joachimmetz: offset relative to what? beginning of file?

plaso/parsers/fseventsd.py
```diff
@@ -89,28 +72,39 @@ def GetFormatSpecification(cls):
...
-    except construct.ConstructError:
-      raise errors.UnableToParseFile('Unable to parse DLS header from file')
+      page_header, page_size = self._ReadStructureFromFileObject(
+          file_object, offset, page_header_map)
+    except (ValueError, errors.ParseError) as exception:
+      raise errors.UnableToParseFile(
+          'Unable to parse page header with error: {0!s}'.format(
+              exception))
+    if page_header.signature not in self._DLS_SIGNATURES:
+      raise errors.UnableToParseFile('Invalid file signature')
```
joachimmetz: File of page signature? Also maybe add the offset to the error message?

plaso/parsers/fseventsd.py
```diff
@@ -89,28 +72,39 @@ def GetFormatSpecification(cls):
...
+    page_header_map = self._GetDataTypeMap('dls_page_header')
     try:
-      return self._DLS_PAGE_HEADER.parse_stream(file_object)
-    except construct.ConstructError:
-      raise errors.UnableToParseFile('Unable to parse DLS header from file')
+      page_header, page_size = self._ReadStructureFromFileObject(
+          file_object, offset, page_header_map)
+    except (ValueError, errors.ParseError) as exception:
+      raise errors.UnableToParseFile(
+          'Unable to parse page header with error: {0!s}'.format(
```
joachimmetz: Maybe add the offset to the error message?

plaso/parsers/fseventsd.py
```diff
@@ -159,8 +153,23 @@ def ParseFileObject(self, parser_mediator, file_object, **kwargs):
     Raises:
       UnableToParseFile: when the header cannot be parsed.
     """
-    page_header = self._ParseDLSPageHeader(file_object)
+    page_header_map = self._GetDataTypeMap('dls_page_header')
+    file_offset = 0
+    file_size = file_object.get_size()
+    try:
+      page_header, header_size = self._ReadStructureFromFileObject(
```
joachimmetz: Maybe?
```
     page_header, file_offset = self._ReadStructureFromFileObject(
         file_object, 0, page_header_map)
```

plaso/parsers/fseventsd.py
```diff
@@ -159,8 +153,23 @@ def ParseFileObject(self, parser_mediator, file_object, **kwargs):
...
+          file_object, file_offset, page_header_map)
+    except (ValueError, errors.ParseError) as exception:
+      raise errors.UnableToParseFile(
+          'Unable to parse page header with error: {0!s}'.format(
+              exception))
+
+    if page_header.signature not in self._DLS_SIGNATURES:
+      raise errors.UnableToParseFile('Invalid file signature')
+
+    file_offset += header_size
```
joachimmetz: Remove this if you apply previous suggestion

plaso/parsers/fseventsd.py
```diff
@@ -172,15 +181,26 @@ def ParseFileObject(self, parser_mediator, file_object, **kwargs):
...
       else:
-        record = self._DLS_RECORD_V2.parse_stream(file_object)
+        record_map = self._GetDataTypeMap('dls_record_v2')
+
+      try:
+        record, record_length = self._ReadStructureFromFileObject(
+            file_object, file_offset, record_map)
+        file_offset += record_length
+      except (ValueError, errors.ParseError) as exception:
+        raise errors.UnableToParseFile(
```
joachimmetz: Create an extraction error here, seeing the some records were successfully parsed and the file does not need to be tried by other parsers

plaso/parsers/fseventsd.py
```diff
@@ -159,8 +153,23 @@ def ParseFileObject(self, parser_mediator, file_object, **kwargs):
     Raises:
       UnableToParseFile: when the header cannot be parsed.
     """
-    page_header = self._ParseDLSPageHeader(file_object)
+    page_header_map = self._GetDataTypeMap('dls_page_header')
```
joachimmetz: Why not collapse the first page header check into the loop?

plaso/parsers/fseventsd.yaml
```diff
@@ -0,0 +1,76 @@
...
+  type: string
+  encoding: utf8
+  element_data_type: byte
+  elements_terminator: "\x00"
+- name: event_identifier
+  data_type: ulint64
+- name: flags
+  data_type: ubint32
+- name: node_identifier
+  data_type: ulint64
```
joachimmetz: add EOL ?

plaso/parsers/fseventsd.py
Onager: There's several things that need to be set up only once the first header has been read. I think this way is easier to read.

plaso/parsers/fseventsd.py
Onager: Changed this a little bit, but I find the suggestion here a little too implicit.

plaso/parsers/fseventsd.py
Onager: Ack

joachimmetz:
A couple of points of feedback:

* the UnableToParseFile exception has a special purpose in plaso, namely that another parser should be applied
* there is no need to define the byte order on a per struct member granularity for most formats.

joachimmetz:
So the flags are very likely stored in little-endian and represent values in:
https://github.com/apple/darwin-xnu/blob/master/bsd/sys/fsevents.h

```
#define FSE_CLONE               14

1 << FSE_CLONE which is 0x00004000 litte-endian (0x00400000 big-endian)
```

PR #2033, opened 2018-07-10
Added codecov configuration

joachimmetz:
Added codecov configuration


Onager:
Most of these settings are already set in the team yaml, or are defaults, per https://docs.codecov.io/docs/codecov-yaml#section-default-yaml and https://github.com/log2timeline/l2tdevtools/commit/3bc87235d70cb7c79e1baec084af94a2ec725657

PR #2052, opened 2018-07-19
Unify "partitions" and "partition" options

Onager:


plaso/cli/storage_media_tool.py
```diff
@@ -1003,21 +992,13 @@ def AddStorageMediaImageOptions(self, argument_group):
...
-            'Choose a partition number from a disk image. This partition '
-            'number should correspond to the partition number on the disk '
-            'image, starting from partition 1. All partitions can be '
-            'defined as: "all".'))
-
-    argument_group.add_argument(
-        '--partitions', dest='partitions', action='store', type=str,
-        default=None, help=(
-            'Define partitions that need to be processed. A range of '
+        '--partitions', '--partition', dest='partitions', action='store',
```
joachimmetz: why not remove partition option?

joachimmetz:
@Onager why not remove the partition option?

Onager:
Because I don't want to have to remember whether the argument is plural or not. This is also why we have arguments with both underscores and hypens, AFAIK.

joachimmetz:
ACK, SG

Issue #2068, opened 2018-08-02
pytz raises AssertionError exception on Fedora Core


joachimmetz:
This does not look like an issue in plaso, but it is affected by it. Considering adding a work-around.

```
======================================================================
FAIL: testListTimeZones (cli.tools.CLIToolTest)
Tests the ListTimeZones function.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "plaso/tests/cli/tools.py", line 149, in testListTimeZones
    cli_tool.ListTimeZones()
  File "./plaso/cli/tools.py", line 291, in ListTimeZones
    local_timezone = pytz.timezone(timezone_name)
  File "/usr/lib/python2.7/site-packages/pytz/__init__.py", line 168, in timezone
    _tzinfo_cache[zone] = build_tzinfo(zone, fp)
  File "/usr/lib/python2.7/site-packages/pytz/tzfile.py", line 33, in build_tzinfo
    assert magic == _byte_string('TZif'), 'Got magic %s' % repr(magic)
AssertionError: Got magic '# Al'

----------------------------------------------------------------------
```

joachimmetz:
Merged as part of https://github.com/log2timeline/plaso/commit/87b3096fcb90ca5a426a7fddbab00742379c359d

PR #2069, opened 2018-08-02
Added work-around for pytz AssertionError #2068

joachimmetz:
Added work-around for pytz AssertionError #2068


joachimmetz:
Merged as part of https://github.com/log2timeline/plaso/commit/87b3096fcb90ca5a426a7fddbab00742379c359d

PR #2080, opened 2018-08-10
Updated pefile dependency version

Onager:
Updated pefile dependency version


setup.py
```diff
@@ -5,7 +5,6 @@
 from __future__ import print_function
```
joachimmetz: Please revert changes to this file

Onager: Done

PR #2094, opened 2018-08-16
Migrated MacOS keychain construct-based plugin to use dtfabric #1893

joachimmetz:
Migrated MacOS keychain construct-based plugin to use dtfabric #1893

### DIFF
--- a/plaso/parsers/mac_keychain.py
+++ b/plaso/parsers/mac_keychain.py
@@ -58,7 +58,8 @@ class KeychainApplicationRecordEventData(events.EventData):
     account_name (str): name of the account.
     comments (str): comments added by the user.
     entry_name (str): name of the entry.
-    ssgp_hash (str): password / cert hash formatted as an hexadecimal string.
+    ssgp_hash (str): password/certificate hash formatted as an hexadecimal
+        string.
     text_description (str): description.
   """
 


plaso/parsers/mac_keychain.py
```diff
@@ -183,298 +142,719 @@ class KeychainParser(interface.FileObjectParser):
       'imap': 'imap',
       'http': 'http'}
 
-  def _ReadEntryApplication(self, parser_mediator, file_object):
-    """Extracts the information from an application password entry.
+  # TODO
```
Onager: Todo what? ;)

plaso/parsers/mac_keychain.py
```diff
@@ -500,46 +879,21 @@ def ParseFileObject(self, parser_mediator, file_object, **kwargs):
...
-      # Table_offset: absolute byte in the file where the table starts.
-      # table.first_record: first record in the table, relative to the
-      #                     first byte of the table.
-      file_object.seek(table_offset + table.first_record, os.SEEK_SET)
+    tables = self._ReadTablesArray(file_object, file_header.tables_array_offset)
 
-      if table.record_type == self.RECORD_TYPE_INTERNET:
-        for _ in range(table.number_of_records):
-          self._ReadEntryInternet(parser_mediator, file_object)
+    table = tables.get(self._RECORD_TYPE_APPLICATION_PASSWORD, None)
```
Onager: You're calling get() with an int argument here, but the docstring from _ReadTablesArray indicates that the dict has string keys. Please check the docstring for the function, I can't readily tell what it's supposed to be.

plaso/parsers/mac_keychain.py
```diff
@@ -183,298 +142,719 @@ class KeychainParser(interface.FileObjectParser):
...
+      raise errors.ParseError('Unsupported format version: {0:s}.{1:s}'.format(
+          file_header.major_format_version, file_header.minor_format_version))
+
+    return file_header
+
+  def _ReadRecord(self, tables, file_object, record_offset, record_type):
+    """Reads the record.
+
+    Args:
+      tables (dict[str, KeychainDatabaseTable]): tables per name.
```
Onager: Should be 'str' be 'int' here? On line 363, you use an int as a key into this dict.

plaso/parsers/mac_keychain.py
joachimmetz: this was/is my refactor marker, I've removed it.

plaso/parsers/mac_keychain.py
joachimmetz: made some changes.

plaso/parsers/mac_keychain.py
joachimmetz: made some changes.

PR #2098, opened 2018-08-19
Changes to prevent scripts/__init__.py from being installed

ant1:
Do not install `__init__.py` as part of scripts


joachimmetz:
@ant1 thx for the suggested, I've made some small modification and will merge once the checks have run

PR #2099, opened 2018-08-20
Updated artifact filter test to support Python 3.3+

Onager:


tests/engine/artifact_filters.py
```diff
@@ -186,7 +187,13 @@ def testBuildFindSpecsFromFileArtifact(self):
     self.assertEqual(len(find_specs), 1)
 
     # Location segments should be equivalent to \Windows\test_data\*.evtx.
-    path_segments = ['Windows', 'test\\_data', '.*\\.evtx']
+    if py2to3.PY_3:
```
joachimmetz: Shouldn't the check here be `Python version >= 3.3` ?

Onager: We don't support any python version below 3.4, so that's not necessary.

joachimmetz: I understand but if change the check we technically can remove the comment.

Onager: I don't think it's worth the extra complexity. Changed the docstring a bit.

joachimmetz: ack.

Issue #1556, opened 2017-12-11
Add parser for Santa OSX whitelisting logs


Onager:
Reference: https://github.com/google/santa

Onager:
Completed it #2100

PR #2100, opened 2018-08-21
Initial support for santa logs #1556

aguilajesus:
Initial support for santa logs #1556


joachimmetz:
Assigning to @Onager due to lack of available time

plaso/parsers/santa.py
```diff
@@ -0,0 +1,397 @@
...
+    action (str): event type recorded by Santa.
+    mount (str): disk mount point.
+    volume (str): disk volume name.
+    bsd_name (str): disk BSD name.
+    fs (str): disk volume kind.
+    model (str): disk model.
+    serial (str): disk serial.
+    bus (str): device protocol.
+    dmg_path (str): DMG file path.
+    appearance (str): disk appearance date.
```
Onager: This looks like a separate (and very useful!) timestamp. What do you think about adding this as a separate event as well?

aguilajesus: I agree, but I am not 100% sure the field is consistent across all DISKAPPEAR events; if appearance field is present we can create another SantaMountEventData() event with definitions.TIME_DESCRIPTION_FIRST_CONNECTED.

aguilajesus:
@Onager PTAL

PR #2119, opened 2018-08-30
Update Chrome Cookies plugin #2118

### LINKED ISSUE #2118: Chrome Cookies schema change
**Description of problem:**

Chrome v66 changed the Cookies db schema slightly, prepending _is__ to three columns: 

- secure -> is_secure
- httponly -> is_httponly
- persistent -> is_persistent

The values remain the same. This causes the chrome_cookies sqlite parser to fail on Chrome v66+.

obsidianforensics:
## Description:

Chrome v66 changed the Cookies db schema slightly, prepending is_ to three columns:

secure -> is_secure
httponly -> is_httponly
persistent -> is_persistent
The values remain the same. This causes the chrome_cookies sqlite parser to fail on Chrome v66+.

This also fixes a problem where the expires flag was incorrectly being parsed as a timestamp, resulting in a time near the epoch. 

**Related issue (if applicable):** fixes #2118

## Notes
This fixes the plugin so it can parse newer Cookie dbs. However, it will log an error when it first tries the Chrome pre-66 query. This isn't ideal, but I didn't see a way to avoid this without making the plugin very fragile (like requiring a full schema check would). I'm open to ideas on how to make this better.

## Checklist:
  - [x] No new [new dependencies](https://github.com/log2timeline/plaso/wiki/Adding-a-new-dependency) are required or l2tdevtools has been updated
  - [x] Reviewer assigned


plaso/parsers/sqlite_plugins/chrome_cookies.py
```diff
@@ -54,11 +54,18 @@ class ChromeCookiePlugin(interface.SQLitePlugin):
 
   # Define the needed queries.
   QUERIES = [
+      # Query for Chrome versions 17 - 65.
       (('SELECT creation_utc, host_key, name, value, path, expires_utc, '
         'secure, httponly, last_access_utc, has_expires, persistent '
+        'FROM cookies'), 'ParseCookieRow'),
+      # Query for Chrome versions 66 and above. The column names changed
+      # slightly, but the values are the same.
+      (('SELECT creation_utc, host_key, name, value, path, expires_utc, '
```
Onager: This means there's going to be an error logged with "unable to run query", when there isn't really a problem. Is there any way around that? Alter the queries, or check for a different table?

aguilajesus: we could use CASE WHEN (https://sqlite.org/syntax/expr.html) but it seems sqlite check that all columns names are present in the table. otherwise, we would have to split the queries in two classes.

obsidianforensics: Yeah, this is what I meant by not ideal. The table check doesn't help here, as the table names are the same. The only way I can see to avoid this error is to split the plugin into multiple sub-plugins that run on each version of the database and enforce an exact schema match. The downside to this is that it makes it "fragile"; minor changes that don't effect the data we're parsing (like adding an additional column or renaming a column we don't even look at) will break the plugin. 

In my opinion, the "required tables" check is too coarse, and the "exact schema match" is too fine. It would be great to have an in-between option, something like "required columns", which would function like "required tables" but one step lower. If the columns specified in "required columns" are a subset of what's present, then run the plugin. I think this approach would help with a bunch of plugins, as the schemas of many databases are constantly evolving. 

I have more ideas about this but it might be a bit much for this PR. I can make an issue for this and we can talk about it more? And in the meantime just make the Cookies parser use the schema matching? Let me know what you think.

Onager: Fair enough - I think we can live with the error message here for now. I think it's clear we need some of "partial schema match" functionality. I've opened #2133 to track that. I'll merge this for now though.

PR #2124, opened 2018-09-02
Migrated systemd journal construct-based plugin to use dtfabric #1893

joachimmetz:
Migrated systemd journal construct-based plugin to use dtfabric #1893


joachimmetz:
Significantly improved performance of refactored parser, having a look if I can squeeze out more.

Onager:
Added @rgayon for a look at this too, since he's familiar with the format

PR #2125, opened 2018-09-02
Updated dependencies and added 64-bit tests

joachimmetz:
Updated dependencies and added 64-bit tests


joachimmetz:
I bumped the version as well

Issue #1628, opened 2018-01-04
Unify data type strings


joachimmetz:
* [ ] change "mac:" and "macosx:" to "macos:"

joachimmetz:
This issue has been superseded by https://github.com/log2timeline/plaso/issues/2836

PR #2140, opened 2018-09-08
unify macOS data type.

aguilajesus:
unify macOS data type.

### DIFF
--- a/data/tag_macos.txt
+++ b/data/tag_macos.txt
@@ -1,5 +1,5 @@
 application_execution
-  data_type is 'macosx:application_usage'
+  data_type is 'macos:application_usage'
   data_type is 'syslog:line' AND body contains 'COMMAND=/bin/launchctl'
 
 application_install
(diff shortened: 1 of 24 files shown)


aguilajesus:
part of #1628

Onager:
This breaks compatibility - psort after this change won't be able to output from older storage files.  This isn't a huge problem, but there's no extra functionality added here, and this is mostly cosmetic.

I'd like to unify the data types at some point, but lets do it all at once, so we only break compatibility once.

aguilajesus:
I see what you're saying, what about unifying the data types now and work a little bit the formatters to maintain compatibility with older storage files e.g.

```
class ApplicationUsageFormatter(interface.EventFormatter):
  """Formatter for a Macosx Application usage event."""
   DATA_TYPE = 'macosx:application_usage'


class MacOSApplicationUsageFormatter(ApplicationUsageFormatter):
  DATA_TYPE = 'macos:application_usage'
```

aguilajesus:
per offline discussion, closing PR until we decide to break backward compatibility.

PR #2158, opened 2018-09-26
Changed DSV file parser to check for expected line length. Remove DSV  file size check.

Onager:


plaso/parsers/dsv_parser.py
```diff
@@ -43,6 +43,9 @@ class DSVParser(interface.FileObjectParser):
   # it can be defined here.
   QUOTE_CHAR = b'"'
 
+  # The maximum size of a single field in the parser
+  FILE_SIZE_LIMIT = csv.field_size_limit()
```
rgayon: this should probably be FIELD_SIZE_LIMIT

plaso/parsers/dsv_parser.py
```diff
@@ -114,6 +118,63 @@ def _CreateDictReader(self, line_reader):
...
+    Args:
+      file_object (dfvfs.FileIO): file-like object.
+
+    Returns:
+        bool: True if the file has lines of the expected length.
+    """
+    original_file_position = file_object.tell()
+    line_reader = self._CreateLineReader(file_object)
+    for _ in range(0, 20):
+      # Attempt to read a line that twice as long as any line that should be in
```
rgayon: +'is'

plaso/parsers/dsv_parser.py
```diff
@@ -114,6 +118,63 @@ def _CreateDictReader(self, line_reader):
...
+      # the file.
+      sample_line = line_reader.readline(self._maximum_line_length * 2)
```
rgayon: shouldn't self._maximum_line_length + 1 be enough of a signal?

plaso/parsers/dsv_parser.py
```diff
@@ -114,6 +118,63 @@ def _CreateDictReader(self, line_reader):
...
+          encoding.
+    """
+    # The Python 2 csv module reads bytes and the Python 3 csv module Unicode
+    # reads strings.
+    if py2to3.PY_3:
+      line_reader = text_file.TextFile(file_object, encoding=self._encoding)
+    else:
+      line_reader = line_reader_file.BinaryLineReader(file_object)
+    # If we specifically define a number of lines we should skip, do that here.
+    for _ in range(0, self.NUMBER_OF_HEADER_LINES):
```
rgayon: Can you have multiple header lines? Or is it "header is present or not" ?
if the latter, maybe have boolean switch instead?

plaso/parsers/dsv_parser.py
Onager: Indeed

plaso/parsers/dsv_parser.py
Onager: The idea behind this seems to have been that there might be multiple header lines (I suspect like the iis format has). As far as I can tell, this constant is never actually set in any subclass, and can be removed. I'd prefer to do that in a different PR though, just in case someone does need this.

plaso/parsers/dsv_parser.py
Onager: Done

plaso/parsers/dsv_parser.py
Onager: Good point - I was thinking of trying to optimize caching here, but both line reader use their own read caches. Changed this.

plaso/parsers/dsv_parser.py
```diff
@@ -59,6 +62,7 @@ def __init__(self, encoding=None):
     """
     super(DSVParser, self).__init__()
     self._encoding = encoding
+    self._maximum_line_length = len(self.COLUMNS) * self.FIELD_SIZE_LIMIT
```
joachimmetz: Shouldn't this be:
`( len(self.COLUMNS) * ( self.FIELD_SIZE_LIMIT + length of field separator) ) +  length of end of line` ?

plaso/parsers/dsv_parser.py
```diff
@@ -114,6 +118,63 @@ def _CreateDictReader(self, line_reader):
         quotechar=quotechar, restkey=magic_test_string,
         restval=magic_test_string)
 
+  # pylint: disable=missing-return-type-doc
+  def _CreateLineReader(self, file_object):
+    """Returns an object that returns lines from a text file.
```
joachimmetz: Please change docstring.

`Returns an object that returns lines from a text file.`

=> `Creates an object that reads lines from a text file.`

plaso/parsers/dsv_parser.py
```diff
@@ -134,35 +195,28 @@ def ParseFileObject(self, parser_mediator, file_object):
...
-    else:
-      line_reader = line_reader_file.BinaryLineReader(file_object)
-
-    # If we specifically define a number of lines we should skip, do that here.
-    for _ in range(0, self.NUMBER_OF_HEADER_LINES):
-      line_reader.readline()
-
-    reader = self._CreateDictReader(line_reader)
+    try:
+      if not self._HasExpectedLineLength(file_object):
```
joachimmetz: What if the file starts with 30 good lines and then is followed by a large blob of non-textual data?

Do we want to check for this edge case? Don't think it is blocking for this CL but let's add a todo or issue?

tests/parsers/dsv_parser.py
```diff
@@ -107,6 +107,18 @@ def testParseFileObject(self):
...
+    """Tests the _HasExpectedLineLength function."""
+    parser = TestDSVParser()
+    test_file_entry = self._GetTestFileEntry(['password.csv'])
+    test_file_object = test_file_entry.GetFileObject()
+
+    self.assertTrue(parser._HasExpectedLineLength(test_file_object))
+
+    parser._maximum_line_length = 2
+    parser._HasExpectedLineLength(test_file_object)
+    self.assertFalse(parser._HasExpectedLineLength(test_file_object))
```
joachimmetz: add white line (2 white lines before main)

plaso/parsers/dsv_parser.py
Onager: I didn't think we needed to do this, and then I couldn't explain it adequately. Then I investigated a bunch more and found another bug in line_reader_file. It was doing unbounded reads, which was probably causing the memory issues in the first place. 

In the end though, we don't need to do anything for this edge case - there'll be a crash in the csv reader now, which is what we want.

Onager:
@joachimmetz @rgayon could you take a look at this when you have minute? Once it's in, we can proceed with the release.

plaso/parsers/dsv_parser.py
```diff
@@ -59,6 +62,13 @@ def __init__(self, encoding=None):
     """
     super(DSVParser, self).__init__()
     self._encoding = encoding
+    if py2to3.PY_2:
+      self._end_of_line = b'\n'
```
joachimmetz: What about files with `\r\n` end of line?

plaso/parsers/dsv_parser.py
```diff
@@ -114,6 +124,65 @@ def _CreateDictReader(self, line_reader):
...
+
+    As we know the maximum length of valid lines in the DSV file, the presence
+    of lines longer than this indicates that the file will not be parsed
+    successfully, without reading excessive data from a large file.
+
+    Args:
+      file_object (dfvfs.FileIO): file-like object.
+
+    Returns:
+        bool: True if the file has lines of the expected length.
```
joachimmetz: 2 space indentation instead of 4 spaces

plaso/parsers/dsv_parser.py
```diff
@@ -134,35 +203,28 @@ def ParseFileObject(self, parser_mediator, file_object):
...
-    for _ in range(0, self.NUMBER_OF_HEADER_LINES):
-      line_reader.readline()
-
-    reader = self._CreateDictReader(line_reader)
+    try:
+      if not self._HasExpectedLineLength(file_object):
+        display_name = parser_mediator.GetDisplayName()
+        raise errors.UnableToParseFile(
+            '[{0:s}] Unable to parse DSV file: {1:s} with error: '
+            'unexpected line length.'.format(
```
joachimmetz: line 216 and 217 looks like they can fit on a single line.

joachimmetz: Also please put a multi line string in between ( .. )

plaso/parsers/dsv_parser.py
Onager: This is to match the current behavior. In dfvfs text file (https://github.com/log2timeline/dfvfs/blob/master/dfvfs/helpers/text_file.py#L19) and line_reader_file where end of line defaults to `\n`, and it wasn't overridden here: https://github.com/log2timeline/plaso/blob/master/plaso/parsers/dsv_parser.py#L154

plaso/parsers/dsv_parser.py
Onager: Done

plaso/parsers/dsv_parser.py
Onager: Done

PR #2169, opened 2018-10-04
Compound ZIP parser

SamuelePilleri:

As discussed in the dev mailing list, this is an early PR about initial work on compound ZIP parser.

## Description:

It obviously isn't ready yet, just tell me what do you think about it and if you have any suggestions. Hope I've followed all the code styling rules.

Also, could someone give me a few hints about tests and linting? I'm not familiar with those in Python.

## Notes:

Replying with a possible problem I've found.

## Checklist:
  - [x] Automated checks (Travis, Codecov, Codefactor) pass
  - [x] No new [new dependencies](https://github.com/log2timeline/plaso/wiki/Adding-a-new-dependency) are required or l2tdevtools has been updated
  - [x] Reviewer assigned


SamuelePilleri:
The parser I've written to replace the OXML one (which is still present) gives different results on test data.

## OXML (original one)

./log2timeline.py

```
Checking availability and versions of dependencies.
[OPTIONAL]	missing: hachoir_core.
[OPTIONAL]	missing: hachoir_metadata.
[OPTIONAL]	missing: hachoir_parser.
[OPTIONAL]	missing: lz4.
[OPTIONAL]	missing: lzma.
[OK]


Source path	: /mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx
Source type	: single file
Processing time	: 00:00:00

Processing started.
plaso - log2timeline version 20180930

Source path	: /mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx
Source type	: single file
Processing time	: 00:00:00

Identifier      PID     Status          Memory          Sources         Events          File
Main            554     collecting      133.5 MiB       0 (0)           0 (0)           

plaso - log2timeline version 20180930

Source path	: /mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx
Source type	: single file
Processing time	: 00:00:00

Identifier      PID     Status          Memory          Sources         Events          File
Main            554     running         133.4 MiB       1 (1)           0 (0)           OS:/mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx

plaso - log2timeline version 20180930

Source path	: /mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx
Source type	: single file
Processing time	: 00:00:00

Identifier      PID     Status          Memory          Sources         Events          File
Main            554     completed       133.3 MiB       1 (0)           0 (0)           

Processing completed.
```

./pinfo.py

```

************************** Plaso Storage Information ***************************
            Filename : oxml.plaso
      Format version : 20180101
Serialization format : json
--------------------------------------------------------------------------------

*********************************** Sessions ***********************************
da355eb2-ffb0-4c4a-8a68-66c3c8411937 : 2018-10-04T14:23:43.281626+00:00
--------------------------------------------------------------------------------

**************** Session: da355eb2-ffb0-4c4a-8a68-66c3c8411937 *****************
                Start time : 2018-10-04T14:23:43.281626+00:00
           Completion time : 2018-10-04T14:23:43.467822+00:00
              Product name : plaso
           Product version : 20180930
    Command line arguments : ./log2timeline --parser oxml
                             ../out/oxml/oxml.plaso test_data/Document.docx
  Parser filter expression : oxml
Enabled parser and plugins : N/A
        Preferred encoding : UTF-8
                Debug mode : False
          Artifact filters : N/A
               Filter file : N/A
--------------------------------------------------------------------------------
No errors stored.

No analysis reports stored.
```

./psort.py

```
plaso - psort version 20180930

Storage file	: ../out/oxml/oxml.plaso

Identifier              PID     Status          Memory          Events          Tags            Reports
Main                    588     exporting       146.4 MiB       0 (0)           0 (0)           0 (0)

Processing completed.

******************************** Export results ********************************
       Events filtered : 0
Events from time slice : 0
      Events processed : 0
--------------------------------------------------------------------------------
```

oxml.csv

```
date,time,timezone,MACB,source,sourcetype,type,user,host,short,desc,version,filename,inode,notes,format,extra
```

## CZIP (rewritten one)

./log2timeline.py

```
Checking availability and versions of dependencies.
[OPTIONAL]	missing: hachoir_core.
[OPTIONAL]	missing: hachoir_metadata.
[OPTIONAL]	missing: hachoir_parser.
[OPTIONAL]	missing: lz4.
[OPTIONAL]	missing: lzma.
[OK]


Source path	: /mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx
Source type	: single file
Processing time	: 00:00:00

Processing started.
plaso - log2timeline version 20180930

Source path	: /mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx
Source type	: single file
Processing time	: 00:00:00

Identifier      PID     Status          Memory          Sources         Events          File
Main            596     collecting      133.5 MiB       0 (0)           0 (0)           

plaso - log2timeline version 20180930

Source path	: /mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx
Source type	: single file
Processing time	: 00:00:00

Identifier      PID     Status          Memory          Sources         Events          File
Main            596     running         133.4 MiB       1 (1)           0 (0)           OS:/mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx

plaso - log2timeline version 20180930

Source path	: /mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx
Source type	: single file
Processing time	: 00:00:00

Identifier      PID     Status          Memory          Sources         Events          File
Main            596     completed       133.0 MiB       1 (0)           2 (2)           

Processing completed.
```

./pinfo.py

```

************************** Plaso Storage Information ***************************
            Filename : czip.plaso
      Format version : 20180101
Serialization format : json
--------------------------------------------------------------------------------

*********************************** Sessions ***********************************
cddabb72-44ec-4477-9f68-57564dd0df80 : 2018-10-04T14:34:02.093675+00:00
--------------------------------------------------------------------------------

**************** Session: cddabb72-44ec-4477-9f68-57564dd0df80 *****************
                Start time : 2018-10-04T14:34:02.093675+00:00
           Completion time : 2018-10-04T14:34:02.277631+00:00
              Product name : plaso
           Product version : 20180930
    Command line arguments : ./log2timeline --parser czip
                             ../out/czip/czip.plaso test_data/Document.docx
  Parser filter expression : czip
Enabled parser and plugins : czip, czip/czip_oxml
        Preferred encoding : UTF-8
                Debug mode : False
          Artifact filters : N/A
               Filter file : N/A
--------------------------------------------------------------------------------

************************* Events generated per parser **************************
Parser (plugin) name : Number of events
--------------------------------------------------------------------------------
           czip_oxml : 2
               Total : 2
--------------------------------------------------------------------------------
No errors stored.

No analysis reports stored.
```

./psort.py

```
plaso - psort version 20180930

Storage file	: ../out/czip/czip.plaso

Identifier              PID     Status          Memory          Events          Tags            Reports
Main                    604     exporting       146.4 MiB       0 (0)           0 (0)           0 (0)

Processing completed.

******************************** Export results ********************************
   Events MACB grouped : 2
      Events processed : 2
       Events filtered : 0
Events from time slice : 0
--------------------------------------------------------------------------------
```

czip.csv

```
date,time,timezone,MACB,source,sourcetype,type,user,host,short,desc,version,filename,inode,notes,format,extra
11/07/2012,23:29:00,UTC,...B,META,Open XML Metadata,Creation Time,-,-,Author: Nides,Creating App: Microsoft Office Word App version: 14.0000 Last saved by: Nides Author: Nides Revision number: 3 Template: Normal.dotm Number of pages: 1 Number of words: 2 Number of characters: 13 Number of characters with spaces: 14 Number of lines: 1 Hyperlinks changed: false Links up to date: false Scale crop: false,2,OS:/mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx,12103423998625070,-,czip/czip_oxml,doc_security: 0; i4: 1; number_of_paragraphs: 1; sha256_hash: 82bc1daae1924f80731ee4d12c48b07e2918389fda465a0954c581ff1b0cb502; shared_doc: false; total_time: 1385
08/25/2013,22:18:00,UTC,M...,META,Open XML Metadata,Content Modification Time,-,-,Author: Nides,Creating App: Microsoft Office Word App version: 14.0000 Last saved by: Nides Author: Nides Revision number: 3 Template: Normal.dotm Number of pages: 1 Number of words: 2 Number of characters: 13 Number of characters with spaces: 14 Number of lines: 1 Hyperlinks changed: false Links up to date: false Scale crop: false,2,OS:/mnt/c/Users/samue/Desktop/plaso/test_data/Document.docx,12103423998625070,-,czip/czip_oxml,doc_security: 0; i4: 1; number_of_paragraphs: 1; sha256_hash: 82bc1daae1924f80731ee4d12c48b07e2918389fda465a0954c581ff1b0cb502; shared_doc: false; total_time: 1385
```

plaso/parsers/czip.py
```diff
@@ -0,0 +1,98 @@
...
+    """Reads a given file from the compound ZIP archive.
+
+    Args:
+      file_name (string): file to read
+
+    Returns:
+      string: the content of the file
+    """
+    with zipfile.ZipFile(self._file_object, 'r') as zip_file:
+      return zip_file.read(file_name)
```
Onager: I'm a bit worried about this being called on a large zipped file here - can you add a check that the file isn't too large before the read is called here?

plaso/parsers/czip_plugins/interface.py
```diff
@@ -0,0 +1,45 @@
...
+  # This is used to understand whether a plugin is suited for a given compound
+  # ZIP archive.
+  # This is expected to be overridden by actual plugins.
+  # Ex. frozenset(['[Content_Types].xml', '_rels/.rels', 'docProps/core.xml'])
+  REQUIRED_ITEMS = frozenset(['any'])
+
+  NAME = 'czip'
+
+  # pylint: disable=arguments-differ
+  def Process(self, parser_mediator, archive_proxy, **kwargs):
```
Onager: Archive_proxy seems to add a fair bit of complexity here, with not a lot of utility. It's also not a great name, since this is actually the czip parser. You can just move Read() into the czip plugin interface and not need this.

plaso/parsers/czip_plugins/oxml.py
```diff
@@ -0,0 +1,320 @@
...
+          and other components, such as storage and dfvfs.
+      archive_proxy (parsers.CompoundZIPParser): a slim proxy giving simple
+          access to archive files.
+
+    Raises:
+      UnableToParseFile: when the file cannot be parsed.
+    """
+
+    # First of all, call upper processing logic
+    super(OpenXMLPlugin, self).Process(parser_mediator, archive_proxy, **kwargs)
```
Onager: Rather than overloading Process here, just create a different method that's called from the Process() implementation in the interface - this is the way the other parsers-that-have-plugins work.

plaso/parsers/czip_plugins/interface.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Process() checks if the current czip file being processed is a match for
+    a plugin by comparing REQUIRED_ITEMS: if they match processing continues;
+    else raise WrongCompoundZIPPlugin.
+
+    This method is intended to be extended by actual plugins with archive
+    specific logic.
+    """
+
+    if not set(archive_proxy.NameList()).issuperset(self.REQUIRED_ITEMS):
```
Onager: Please break this up into multiple lines, to make debugging easier. Also, we probably want to cache the result of NameList() so its only called once per file, not once for every plugin.

plaso/parsers/czip_plugins/interface.py
```diff
@@ -0,0 +1,45 @@
...
+
+class CompoundZIPPlugin(plugins.BasePlugin):
+  """This is and "abstract" class from which plugins should be based."""
+
+  # REQUIRED_ITEMS is a list of items required by a plugin.
+  # This is used to understand whether a plugin is suited for a given compound
+  # ZIP archive.
+  # This is expected to be overridden by actual plugins.
+  # Ex. frozenset(['[Content_Types].xml', '_rels/.rels', 'docProps/core.xml'])
+  REQUIRED_ITEMS = frozenset(['any'])
```
Onager: Rename this to REQUIRED_PATHS? Since these are file paths inside the archive.

What does 'any' mean here? Just leave it empty if there's no content left.

plaso/parsers/czip.py
```diff
@@ -0,0 +1,98 @@
...
+
+  def NameList(self):
+    """Lists all files inside the archive with paths referring to the root
+    of the archive.
+
+    Returns:
+      list: a list of files in the archive, for example
+          ['foo.txt', 'bar/quxx.doc']
```
Onager: Is the separator here always '/' or is it platform dependent? Is it '\' on Windows?

SamuelePilleri: Just checked and it's '/' on Windows as well.

plaso/parsers/czip_plugins/interface.py
SamuelePilleri: My fear was that an empty set is a superset for everything, so forgetting to override that field with a proper value would result in a sneaky bug. Let me know which approach do you prefer.

plaso/parsers/czip.py
SamuelePilleri: I thought this was enough, it's the way the old parser used to do it https://github.com/SamuelePilleri/plaso/blob/e19fb6c5c10dd63c6c4e015708c127a0e17aba63/plaso/parsers/czip.py#L60

plaso/parsers/czip_plugins/interface.py
SamuelePilleri: So I just pass `file_object` instead? Or is there a smarter way to access parent parser from within a plugin?

plaso/parsers/czip.py
SamuelePilleri: I've found a question on StackOverflow that seems to address this: [When are files too large to be read as strings in Python?](https://stackoverflow.com/questions/40894264/when-are-files-too-large-to-be-read-as-strings-in-python) Quoting: _"The theoretical limit is the maximum size of a Python string, which is determined by it's index, which is `2 ** 63` [...] (the theoretical limit is smaller depending on the virtual memory available: that will be around 2GB by default for 32bit python.)"_ and _"if holding a string in memory demands more memory than your system has then you will get a `MemoryError`"_. Also, [Python documentation for `zipfile.read()`](https://docs.python.org/2/library/zipfile.html#zipfile.ZipFile.read) doesn't specify any exception to cover this specific case and I've already checked that the zip file itself isn't too large. All I can do here is blindly catch all exceptions that line could generate and wrap the error message in a more descriptive string.

A possible, but more restrictive, check would be to test if [uncompressed file size](https://docs.python.org/2/library/zipfile.html#zipfile.ZipInfo.file_size) is smaller than 2GB: in fact, the [docs](https://docs.python.org/2/library/zipfile.html#zipfile.ZipFile) specify that ZIP64 extensions are disabled by default: that would be safe on 32 bit systems, but if someone ever enables such extensions, then said check should be dropped.

Let me know what do you prefer.

Onager: "Too large" is a bit arbitrary, unfortunately. We limit worker memory usage to 4GiB https://github.com/log2timeline/plaso/blob/master/plaso/cli/helpers/process_resources.py#L35 and something like reading a 2GiB file and trying to parse it with a JSON or CSV parser will exceed that pretty easily, and there are plenty of other ways to end up duplicating a 1GB string a few times and exceeding a limit or just being generally slow. 

Catching exceptions wont work by the way, the Plaso foreman will kill the worker process for exceeding the limit.

Good point about Zip64 - please set it on explicitly as it's on by default in Python 3. Also, I assume Facebook dumps can be >2GiB, and that's the main motivation for this change...

plaso/parsers/czip_plugins/interface.py
Onager: Yes, you could parse file_object, or even zip_file. There's no easy way to reference the parent parser from a plugin.

plaso/parsers/czip.py
Onager: BTW, https://github.com/log2timeline/dfvfs/blob/master/dfvfs/file_io/zip_file_io.py for the dfVFS implementation of reading a zip file without exploding memory. It's pretty complicated, but might be a way to get a Read() primitive safely.

SamuelePilleri: I wonder if I can use it in place of `zipfile`...

SamuelePilleri: A perfect implementation would, given a filename which is present inside the zip, return a file-like object from which the JSON (or XML or whatever) parser can read... Do you think this solves our problem? https://docs.python.org/2/library/zipfile.html#zipfile.ZipFile.open

SamuelePilleri: Ok, I need to know, is [`plugin.UpdateChainAndProcess()`](https://github.com/SamuelePilleri/plaso/blob/e19fb6c5c10dd63c6c4e015708c127a0e17aba63/plaso/parsers/czip.py#L69) somewhat asynchronous? Because if it's not (or if context manager can handle it) I could just write:

```python
with zipfile.ZipFile(file_object, 'r') as zip_file_object:
  plugin.UpdateChainAndProcess(parser_mediator, zip_file_object)
```

And that is smooooth!

SamuelePilleri:
I think I've addressed all the issues, except for this one:

>>What does 'any' mean here? Just leave it empty if there's no content left.
>>
>My fear was that an empty set is a superset for everything, so forgetting to override that field with a proper value would result in a sneaky bug.

Let me know if you think it's ready so I can remove the legaxy OpenXML parser (if you agree) and fix [this wording](https://github.com/log2timeline/plaso/pull/2169/commits/ee5fc5d8cccc0659da8201b5f53a7e1c6cf80d8d#diff-230b7f73adf299e1f251f1f8a82949a6R37).

plaso/parsers/czip.py
SamuelePilleri: Resolving this conversation, current implementation leaves it up to the plugin to decide to either `read()` a whole file inside the archive or `open()` it and handle it more carefully. As for the XML/JSON part, I don't think there will ever be individual files larger than 2 GiB and string copy shouldn't be a problem since the XML/JSON parser can directly read from the archive as the algorithm incrementally deflates it.

SamuelePilleri:
I'm considering changing [plugin name](https://github.com/log2timeline/plaso/pull/2169/files#diff-ab1acd65ee51ea9a6a929854e9fa6a80R75) to a much simpler "oxml", this way we can use `--parsers czip/oxml`. [Tests](https://github.com/log2timeline/plaso/pull/2169/files#diff-023d263f917acb1a4945ab3908fc0460R24) should be fixed too. What do you think?

Onager:
Sorry for the delayed response on this - I've been travelling. I hope to get to back to this review in the next day or so.

SamuelePilleri:
Don't worry, take your time, I'll be off for the weekend anyway.

plaso/parsers/czip_plugins/interface.py
```diff
@@ -1,73 +1,64 @@
...
+          this method, but will be closed by the parser logic in czip.py.
+      archive_members (list[str]): file paths in the archive.
 
     Raises:
       UnableToParseFile: when the file cannot be parsed.
+      ValueError: if a subclass has not specified REQUIRED_PATHS.
     """
-    if not set(name_list).issuperset(self.REQUIRED_PATHS):
-      raise errors.WrongCompoundZIPPlugin(self.NAME)
+    if not self.REQUIRED_PATHS:
```
SamuelePilleri: This may cause a problem in the future with localized czips, such as Google Takeout ones. Since filenames are localized, there is no way to tell if a given file is present inside the archive. With the previous implementation, one could leave this set empty on purpose and implement a custom logic [*], but now the whole thing needs to be reengineered.

Moreover, I've found a strange behaviour with both Python 2 and 3: in some cases [`ZipFile.namelist()`](https://docs.python.org/2/library/zipfile.html#zipfile.ZipFile.namelist) will list folder paths while in others won't; I've tested a couple of examples and I guess it depends on the library used to create the zip file, but I can further investigate.

In other words, given a context where there are some files (say _foo.txt_ and _bar.txt_) inside a directory (eg. _folder/_), `namelist()` may return either `['folder/', 'folder/foo.txt', 'folder/bar.txt']` or `['folder/foo.txt', 'folder/bar.txt']`.

[*]: I get it's not elegant, but it works

plaso/parsers/czip_plugins/oxml.py
```diff
@@ -72,13 +69,15 @@ def __init__(self):
     self.template = None
     self.total_time = None
 
-
-class OpenXMLParser(interface.FileObjectParser):
+class OpenXMLPlugin(interface.CompoundZIPPlugin):
   """Parse metadata from OXML files."""
 
-  NAME = 'openxml'
+  NAME = 'czip_oxml'
```
SamuelePilleri: Here I repeated "czip_" because other plugins do the same, but since others don't, I'd rather change it to "oxml" or "openxml", so we can use `--parser czip/oxml` instead. Let me know which one do you prefer.

Onager: Good point, I've changed to just oxml

plaso/parsers/czip_plugins/interface.py
Onager: As discussed, we'll need to look at this for future compound zips, but we'll go ahead with this approach for now.

Issue #2170, opened 2018-10-05
ERROR: Unable to read artifact definitions from: \artifacts with error: At start Artifact definition: HadoopAppRoot found undefined labels: Hadoop.


tdiogos:
Hi,
The Windows version of log2timeline.exe fails with the error "ERROR: Unable to read artifact definitions from: \artifacts with error: At start Artifact definition: HadoopAppRoot found undefined labels: Hadoop."

Version used: [plaso-20180930.zip](https://github.com/log2timeline/plaso/releases/download/20180930/plaso-20180930-amd64.zip) 

Tested on Windows 10

jturkey:
ERROR: Unable to read artifact definitions from: D:\Tools\Plaso20180930\artifacts with error: At start Artifact definition: HadoopAppRoot found undefined labels: Hadoop.

same issue here as well.

Sim4n6:
I deleted the line of "labels" but this is a hardcoded and an untrusted solution.

mark-hallman:
I am also experiencing this same error and it is also happening with a few other parsers too.  I'm testing in the Windows version of 20180930.  I think this needs an immediate review as the only way I can get log2timeline to run is the rename the offending .ymal file.  Obviously this not ideal.  I don't think that this should be labeled a "question".

jturkey:
Thank you.

mark-hallman:
I'm wondering if there is a fix or workaround on this issue?

joachimmetz:
> I think this needs an immediate review 

@mark-hallman this is a best effort project, if you want an immediate fix/response please dedicate your time to the project.

mark-hallman:
Joachim,

Point well taken.  Please accept my apologies for my poorly chosen words.
 I am very appreciative of all the hard that goes into this project.

-Mark

Onager:
Fix to the windows distribution script is pending in https://github.com/log2timeline/plaso/pull/2191 and I've re-packaged plaso-20180930.zip with the fixes. Try the new package, which I've attached to the release: https://github.com/log2timeline/plaso/releases/tag/20180930

jturkey:
Awesome!  Thanks for you're time and commitment to the project.  I will try
running it this morning.

big-husky:
I can also confirm this is fixed by #2191   Thank you.

ctmayhew:
Fixed for me too - thank you!

PR #2191, opened 2018-11-06
Changed Windows release script to use last tagged version of dependency data files #2170

Onager:


joachimmetz:
@Onager  plase rebase with upstream master so test can run

PR #2239, opened 2018-12-03
Removed unused filter and circular dependency

Onager:


joachimmetz:
@kiddinn could you have a look to see what this functionality was intended for

kiddinn:
this was used "back in the day" for filters like 'parser inlist "winxp AND ...." so that you could filter based on both the content of the event as well as the parser it was generated by... that is you could specify a list of parsers instead of being explicit of "parser is registry", the parser was part of a [parser preset](https://github.com/log2timeline/plaso/blob/master/plaso/parsers/presets.py).

I don't think this got heavily used, and is probably "flaky", and is also quite old, so I think it is more than fine to get rid of it.

joachimmetz:
@kiddinn thx for the context

Issue #2277, opened 2018-12-21
update Parsers-and-plugins documentation


joachimmetz:
Information in docs/sources/user/Parsers-and-plugins.md can easily go stale, auto-generate it from the source?

joachimmetz:
`PYTHONPATH=. ./tools/log2timeline.py --parsers list --use-markdown > docs/sources/user/Parsers-and-plugins.md`

joachimmetz:
https://github.com/log2timeline/l2tdocs/pull/38

joachimmetz:
Updated documentation: https://github.com/log2timeline/plaso/pull/2280

PR #2280, opened 2018-12-21
Updated parsers and plugins documentation

joachimmetz:
Updated parsers and plugins documentation


joachimmetz:
@romaing yes this is partially automated see: https://github.com/log2timeline/plaso/issues/2277

Issue #1662, opened 2018-01-19
Move presets to configuration file


Onager:
Move from plaso/parsers/presets.py

joachimmetz:
* Update "Parsers-and-plugins documentation" to reflect that preset is defined in configuration file

joachimmetz:
Moving aftercare to Feb milestone

PR #2282, opened 2018-12-22
Moved parser presets into class #1662

joachimmetz:
Moved parser presets into class #1662

### DIFF
--- a/plaso/parsers/presets.py
+++ b/plaso/parsers/presets.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""The parser and parser pluging presets."""
+"""The parser and parser plugin presets."""
 
 from __future__ import unicode_literals
 
--- a/plaso/parsers/presets.py
+++ b/plaso/parsers/presets.py
@@ -1,46 +1,105 @@
 # -*- coding: utf-8 -*-
-"""The parser preset categories."""
+"""The parser and parser pluging presets."""
 
 from __future__ import unicode_literals
 
-CATEGORIES = {
-    'win_gen': [
-        'bencode', 'esedb', 'filestat', 'sqlite/google_drive', 'gdrive_synclog',
-        'java_idx', 'lnk', 'mcafee_protection', 'olecf', 'czip/oxml', 'pe',
-        'prefetch', 'sccm', 'skydrive_log', 'skydrive_log_old', 'sqlite/skype',
-        'symantec_scanlog', 'usnjrnl', 'webhist', 'winfirewall', 'winjob',
-        'winreg'],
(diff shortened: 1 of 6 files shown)


plaso/parsers/presets.py
```diff
@@ -1,46 +1,106 @@
 # -*- coding: utf-8 -*-
-"""The parser preset categories."""
+"""The parser and parser pluging presets."""
```
Onager: plugin

plaso/parsers/presets.py
```diff
@@ -1,46 +1,106 @@
...
+    """Initializes a parser and parser plugin preset definition."""
+    super(ParserPresetDefinition, self).__init__()
+    self.name = None
+    self.parsers = None
+
+
+class ParserPresets(object):
+  """Parser and parser plugin presets."""
+
+  CATEGORIES = {
```
Onager: These don't seem to be categories, these are presets by name, using the new terminology. Unless there's some other factor at work, please  at least rename this, or even better, move directly to using the manager approach as elsewhere, and directly create ParserPresetDefinitions.

plaso/parsers/presets.py
```diff
@@ -1,46 +1,106 @@
...
-        'mac_securityd', 'sqlite/mackeeper_cache', 'macwifi', 'olecf',
-        'czip/oxml', 'plist', 'sqlite/skype', 'syslog', 'utmpx', 'webhist',
-        'zsh_extended_history'],
-    'android': [
-        'android_app_usage', 'filestat', 'chrome_cache',
-        'sqlite/android_calls', 'sqlite/android_sms', 'sqlite/android_webview',
-        'sqlite/android_webviewcache', 'sqlite/chrome_cookies',
-        'sqlite/chrome_8_history', 'sqlite/chrome_27_history', 'sqlite/skype']}
+
+class ParserPresetDefinition(object):
```
Onager: -Definition from the name here, as a preset is the name and list of parsers.

plaso/parsers/presets.py
joachimmetz: Keeping this as-is, this class constant will be entirely removed in the next step.

plaso/parsers/presets.py
joachimmetz: done

plaso/parsers/presets.py
joachimmetz: There will be more attributes in this class in the next steps.

joachimmetz:
@Onager this change needs to be considered in the context of multiple steps also see https://github.com/log2timeline/plaso/pull/1662. You are asking for changes for code that will be affected by the next steps.

PR #2288, opened 2018-12-26
Changed preset configuration to define operation system mapping #1662

joachimmetz:
Changed preset configuration to define operation system mapping #1662


plaso/cli/extraction_tool.py
```diff
@@ -103,16 +107,21 @@ def _CreateProcessingConfiguration(self, knowledge_base):
...
               operating_system, operating_system_product,
               operating_system_version))
 
-      if parser_filter_expression:
-        logger.info('Parser filter expression changed to: {0:s}'.format(
-            parser_filter_expression))
+      if preset_definitions:
+        if len(preset_definitions) != 1:
+          raise errors.BadConfigOption(
+              'More than 1 parser presets found for: {0:s}'.format(
```
Onager: "more than one parser preset", not "parser presets". Also, please print the preset names here to aid debugging.

plaso/containers/artifacts.py
```diff
@@ -66,6 +67,120 @@ def __init__(self, name=None, schema='DNS'):
...
+      version (str): version, such as "10.14.1" or "5.1".
+    """
+    super(OperatingSystemArtifact, self).__init__()
+    self.name = name
+    self.product = product
+    self.version = version
+
+  @property
+  def normalized_product(self):
+    """str: normalized product name."""
```
Onager: Is what a "normalized product name" is defined somewhere? Without reading the method logic here, I have no idea what this value would be. Please add either a description of what this is here, or mention where one might find a definition.

plaso/containers/artifacts.py
```diff
@@ -66,6 +67,120 @@ def __init__(self, name=None, schema='DNS'):
...
+      product (str): product name, such as "macOS Mojave" or "Windows XP".
+      version (str): version, such as "10.14.1" or "5.1".
+    """
+    super(OperatingSystemArtifact, self).__init__()
+    self.name = name
+    self.product = product
+    self.version = version
+
+  @property
+  def normalized_product(self):
```
Onager: Shouldn't this be normalized_product_name, per the docstring?

plaso/containers/artifacts.py
```diff
@@ -66,6 +67,120 @@ def __init__(self, name=None, schema='DNS'):
...
+
+  @property
+  def version_tuple(self):
+    """tuple[int]: version tuple or None if version is not set or invalid."""
+    try:
+      return tuple([int(digit, 10) for digit in self.version.split('.')])
+    except (AttributeError, TypeError, ValueError):
+      return None
+
+  def Compare(self, other):
```
Onager: Please rename to "Equal" or "IsEquivalent" or something. "Compare" is too generic for what this method does. It's not clear what sort of comparison is being performed.

plaso/containers/artifacts.py
```diff
@@ -66,6 +67,120 @@ def __init__(self, name=None, schema='DNS'):
...
+    Args:
+      other (OperatingSystemArtifact): operating system artifact attribute
+          container to compare with.
+
+    Returns:
+      bool: True if the operating systems are considered equivalent, False if
+          the most specific criteria do no match, or no criteria are available.
+    """
+    if self.product and other.product:
+      return self.normalized_product == other.normalized_product
```
Onager: As I read this, this method seems to consider "macOS Mojave 10.14.1" and "macOS Mojave 10.4.2" to be equivalent, but this doesn't seem like "the most specific criteria". Was this the intention here?

tests/containers/artifacts.py
```diff
@@ -7,93 +7,131 @@
...
-        'code_page': 'cp1252',
-        'time_zone': 'UTC',
-        'user_accounts': []}
+    attribute_container = artifacts.OperatingSystemArtifact(version="5.a")
+    self.assertIsNone(attribute_container.version_tuple)
 
-    test_dict = attribute_container.CopyToDict()
+  def testCompare(self):
+    """Tests the Compare function."""
+    attribute_container1 = artifacts.OperatingSystemArtifact(
```
Onager: Rather than reusing "attribute_container1" and "attribute_container2" in multiple asserts, please rename this variables so these tests are easier to read - self.assertFalse(win2k12_artifact.Compare(winxp_artifact)) is a lot easier to reason about than self.assertFalse(attribute_container1.Compare(attribute_container2))

plaso/parsers/presets.py
```diff
@@ -13,6 +14,9 @@ class ParserPreset(object):
 
   Attributes:
     name (str): name of the preset.
+    operating_system (list[OperatingSystemArtifact]): operating system
+        artifact attribute containers, that specify to which operating
+        systems the preset should apply.
```
Onager: -should

plaso/cli/extraction_tool.py
joachimmetz: ack

plaso/parsers/presets.py
joachimmetz: ack

tests/containers/artifacts.py
joachimmetz: Seeing this has come up multiple times, I opt to add this to the style guide: https://github.com/log2timeline/plaso/issues/2303

plaso/containers/artifacts.py
joachimmetz: Why should it in your opinion? Please explain your need to change this?

plaso/containers/artifacts.py
joachimmetz: It's a normalized version of the product name, it is unclear to me how a formal definition is going to help?

Instead could you explain what is not clear about the logic below, maybe that could be changed instead?

plaso/containers/artifacts.py
joachimmetz: ack.

plaso/containers/artifacts.py
joachimmetz: Unclear what you mean, as indicated in the docstring

```
These criteria, in order of most to least specific, are:
    * product
   ...
```

So why would this not be the most specific criteria in your opinion? Could you be more specific what is confusing you here?

tests/containers/artifacts.py
Onager: This is already addressed in the google style guide: "Function names, variable names, and filenames should be descriptive". `attribute_container1` is less descriptive than `win2k12_artifact`. 

We could debate "sufficiently descriptive", but I don't think we could write a style guide definition without a lot of pain. We'll have to keep discussing this in code review, IMO.

joachimmetz: this remains subjective interpret-able, for me the type is more descriptive than the function, especially for tests. This needs to be clarified in more detail in the style guide.

plaso/containers/artifacts.py
joachimmetz: ack, as discussed specific here does not refer to the version string, I'll see if I can make this less ambiguous.

plaso/containers/artifacts.py
```diff
@@ -71,58 +71,62 @@ class OperatingSystemArtifact(ArtifactAttributeContainer):
   """Operating system artifact attribute container.
 
   Attributes:
-    name (str): name, such as "Linux", "MacOS" or "Windows".
-    product (str): product name, such as "macOS Mojave" or "Windows XP".
-    version (str): version, such as "10.14.1" or "5.1".
+    family (str): operating system family name, such as "Linux", "MacOS"
+        or "Windows". This value is used to programmatically link a parser
+        preset to an operating system and therefore must be one of predefined
```
Onager: I think this means to be "predefined values in definitions.OPERATING_SYSTEM_FAMILIES". If not, please clarify.

There are a few other places where this docstring is replicated, please fix those also.

joachimmetz: ack

PR #2299, opened 2019-01-20
Changes to run CI tests with pylint 2.x and removed Trusty CI tests

joachimmetz:
Changes generated with log2timeline/l2tdevtools#560


joachimmetz:
@Onager PTAL lint tests are finally passing, new pylint is a bit more picky. I opt that we clean this up further after the January release.

PR #2306, opened 2019-01-24
fix error on config tag analysis

lprat:
Fix error on config of tag analysis. 

## Description:
In my last PR, I make a mistake, because I did not have understand than tag analysis modify the plaso dump. I was thinking than only file generate from psort was impacted.
The Rule with "contains" on field "regvalue" don't work. I choose to use another syntaxe of efilter with regexp ("=~"), and it's work fine.
I tried of reuse __all__, but same crash on plaso if field not exist or empty.
I added a new rule in tag "document_open" with use 'link' file in "recent" directory.

## Checklist:
  - [ ] Reviewer assigned


lprat:
It's not code, just config file for plugin analysis...

Onager:
This seems to be triggering a bug in efilter on python 3, hence the failing test. I need some more time to look at this, so it won't make the January release, unfortunately.

lprat:
I close because efilter disabled.

PR #2346, opened 2019-02-09
Removed ObjectFilterList

joachimmetz:
Removed ObjectFilterList


joachimmetz:
@kiddinn @Onager to my knowledge the filter manager is only used for the psort event filter (which is a string) and the tagging file (which are strings as well). So it is unclear to me if we are using ObjectFilterList.

PR #2347, opened 2019-02-09
Removed DynamicFilter

joachimmetz:
Removed DynamicFilter


joachimmetz:
@kiddinn @Onager to my knowledge we have not documented DynamicFilter or use it.

Issue #2354, opened 2019-02-19
source file names incompatible NTFS file names


286DX:
**Description of problem:**
The names of the following files are not compatible with the NTFS filesystem:
```
docs/sources/user/Analysis-plugin:-nsrlsvr.md
docs/sources/user/Analysis-plugin:-tagging.md
docs/sources/user/Analysis-plugin:-viper.md
docs/sources/user/Analysis-plugin:-virustotal.md
```
Git tries to write them into ADSs of the file docs/sources/user/Analysis-plugin.

**Command line and arguments:**

```
D:\>git clone https://github.com/log2timeline/plaso
Cloning into 'plaso'...
remote: Enumerating objects: 42341, done.
remote: Total 42341 (delta 0), reused 0 (delta 0), pack-reused 42341
Receiving objects: 100% (42341/42341), 184.01 MiB | 1.49 MiB/s, done.
Resolving deltas: 100% (32295/32295), done.
Checking out files: 100% (1467/1467), done.

D:\>cd plaso

D:\plaso>git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        deleted:    docs/sources/user/Analysis-plugin:-nsrlsvr.md
        deleted:    docs/sources/user/Analysis-plugin:-tagging.md
        deleted:    docs/sources/user/Analysis-plugin:-viper.md
        deleted:    docs/sources/user/Analysis-plugin:-virustotal.md

Untracked files:
  (use "git add <file>..." to include in what will be committed)

        docs/sources/user/Analysis-plugin

no changes added to commit (use "git add" and/or "git commit -a")
```
**Plaso version:**
Current master

joachimmetz:
@Onager this issue looks completed, can it be closed?

PR #2355, opened 2019-02-19
Removed colon from analysis plugin doc filenames #2354

Onager:


joachimmetz:
efilter test is broken due to changes in l2tdevtools

Issue #2365, opened 2019-02-21
ValueError: Invalid size value exceeds maximum value 16777216


soundwave01:
**Description of problem:**

Log2timeline appears to be unable to process large files, like the output from analyzeMFT.   The following error is thrown  "ValueError: Invalid size value exceeds maximum value 16777216."

The error is being thrown on the MAXIMUM_READ_BUFFER_SIZE checks in the readline function of line_reader_file.py  which is called by dsv-parser.pv.

Having checked the files that log2timeline is processing, the first lines are not bigger than 16777216 in size.

I'm struggling to understand where the value for size is getting set (see debug output), it does not appear to correlate to the size of any of the lines or the size of the file.

While in this example I have just shown the error from processing the output from analyzeMFT, I have the same error on around 80 different files.

**Command line and arguments:**

```
analyzeMFT.py -f $MFT --bodyfull -p -b -c mft.csv
log2timeline.py --no_dependencies_check --logfile out.txt --single_process -d test-out MFTbody
```

**Plaso version:**

plaso - log2timeline version 20190131

**Operating system Plaso is running on:**

Ubuntu 16.04.5 LTS

**Installation method:**

apt-get

**Debug output/tracebacks:**

stack trace

```
2019-02-21 14:17:56,340 [ERROR] (MainProcess) PID:7156 <single_process> Invalid size value exceeds maximum value 16777216.
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/plaso/engine/single_process.py", line 48, in _ProcessPathSpec
    extraction_worker.ProcessPathSpec(parser_mediator, path_spec)
  File "/usr/local/lib/python2.7/dist-packages/plaso/engine/worker.py", line 818, in ProcessPathSpec
    self._ProcessFileEntry(mediator, file_entry)
  File "/usr/local/lib/python2.7/dist-packages/plaso/engine/worker.py", line 641, in _ProcessFileEntry
    self._ProcessFileEntryDataStream(mediator, file_entry, data_stream)
  File "/usr/local/lib/python2.7/dist-packages/plaso/engine/worker.py", line 733, in _ProcessFileEntryDataStream
    mediator, file_entry, data_stream.name)
  File "/usr/local/lib/python2.7/dist-packages/plaso/engine/worker.py", line 329, in _ExtractContentFromDataStream
    mediator, file_entry, data_stream_name)
[...]

PR #2366, opened 2019-02-21
Added check for DSV line length #2365

Onager:


plaso/parsers/dsv_parser.py
```diff
@@ -50,9 +50,6 @@ class DSVParser(interface.FileObjectParser):
   # file to see if it confirms to standards.
   _MAGIC_TEST_STRING = b'RegnThvotturMeistarans'
 
-  # Maximum supported file size of 16 MiB.
```
Onager: This is a refactoring artifact, and should have been removed in https://github.com/log2timeline/plaso/pull/2158

PR #2370, opened 2019-02-23
Removed elasticsearch5-py as a dependency

joachimmetz:
Removed elasticsearch5-py as a dependency


Onager:
@joachimmetz Please resolve conflicts before merge

PR #2375, opened 2019-03-04
Added support to set Windows 10 console to ANSI mode

joachimmetz:
Added support to set Windows 10 console to ANSI mode

### DIFF
--- a/plaso/cli/status_view.py
+++ b/plaso/cli/status_view.py
@@ -3,6 +3,7 @@
 
 from __future__ import unicode_literals
 
+import ctypes
 import sys
 import time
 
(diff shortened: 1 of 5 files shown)


plaso/cli/status_view.py
```diff
@@ -56,6 +58,12 @@ def __init__(self, output_writer, tool_name):
     self._storage_file_path = None
     self._tool_name = tool_name
 
+    if win32console:
+      kernel32 = ctypes.windll.kernel32
+      stdout_handle = kernel32.GetStdHandle(-11)
```
Onager: Please extract these values (-11, 7) to class constants.

joachimmetz: done

PR #2376, opened 2019-03-05
Set log file by default

joachimmetz:
Set log file by default

### DIFF
--- a/plaso/cli/tools.py
+++ b/plaso/cli/tools.py
@@ -146,6 +146,13 @@ class CLITool(object):
       options (argparse.Namespace): command line arguments.
     """
     self._log_file = self.ParseStringOption(options, 'log_file')
+    if not self._log_file:
+      local_date_time = datetime.datetime.now()
+      self._log_file = (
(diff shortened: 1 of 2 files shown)


Onager:
Ref issue #2377

plaso/cli/tools.py
```diff
@@ -223,8 +230,9 @@ def AddLogFileOptions(self, argument_group):
     argument_group.add_argument(
         '--logfile', '--log_file', '--log-file', action='store',
         metavar='FILENAME', dest='log_file', type=str, default='', help=(
-            'If defined all log messages will be redirected to this file '
-            'instead the default STDERR.'))
+            'Path of the file in which to store log messages, by default '
+            'this file will be named: {0:s}-YYYYMMDDThhmmss.log').format(
```
Onager: Per line 152, it's actually .log.gz. Please harmonize the values here.

I think it's also worth mentioning that if the file name ends in .gz, the file will be compressed, but otherwise it'll be uncompressed.

joachimmetz: ack, changes made

PR #2382, opened 2019-03-09
Ensured session identifier passed as string in pinfo

Onager:
Session_identifier is passed as a string to several methods, but was actually a UUID.

### DIFF
--- a/plaso/cli/pinfo_tool.py
+++ b/plaso/cli/pinfo_tool.py
@@ -337,7 +337,7 @@ class PinfoTool(
     """
     for session_number, session in enumerate(storage.GetSessions()):
       session_identifier = uuid.UUID(hex=session.identifier)
-      session_identifier = '{1!s}'.format(session_identifier)
+      session_identifier = '{0!s}'.format(session_identifier)
 
       start_time = 'N/A'
       if session.start_time is not None:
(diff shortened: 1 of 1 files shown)


plaso/cli/pinfo_tool.py
```diff
@@ -337,6 +337,7 @@ def _PrintSessionsDetails(self, storage):
     """
     for session_number, session in enumerate(storage.GetSessions()):
       session_identifier = uuid.UUID(hex=session.identifier)
+      session_identifier = str(session_identifier)
```
joachimmetz: `str()` is not portable since it is bytes on Python 2 and unicode on Python 3 use `'{0!s}'.format()` instead

Onager: It gives the native string type on each version, but if you want a larger refactor, I'll change the other usages.

Onager: I've removed the other usages of str() in the file.

plaso/cli/pinfo_tool.py
```diff
@@ -337,6 +337,7 @@ def _PrintSessionsDetails(self, storage):
     """
     for session_number, session in enumerate(storage.GetSessions()):
       session_identifier = uuid.UUID(hex=session.identifier)
+      session_identifier = '{1!s}'.format(session_identifier)
```
joachimmetz: change 1 to 0 since it is the first argument

joachimmetz: an alternative is to use the py2to3.UNICODE type macro

Onager: Done

joachimmetz:
Rerunning CI tests.

PR #2406, opened 2019-03-21
Removed Travis CI zeromq work-around

joachimmetz:
Removed Travis CI zeromq work-around


joachimmetz:
> though maybe update the commit message with the default log file option?

sry about that, these are changes from other PRs leaking into this one

PR #2407, opened 2019-03-21
Code clean up and added storage file format version on write check

Onager:


plaso/storage/interface.py
```diff
@@ -698,6 +759,27 @@ def __init__(self, path):
     self._path = path
     self._storage_file = None
 
+  @property
+  def format_version(self):
+    """int: format version"""
```
joachimmetz: please add something in line of `or None if not set`

plaso/storage/sqlite/sqlite_file.py
```diff
@@ -194,6 +197,10 @@ def _CheckStorageMetadata(cls, metadata_values):
     except (TypeError, ValueError):
       raise IOError('Invalid format version: {0!s}.'.format(format_version))
 
+    if not check_readable_only and format_version != cls._FORMAT_VERSION:
```
joachimmetz: why not use something like `check_writable` and reverse the logic?

Onager: This was to match the "read_only" behavior in the storage interface.

plaso/storage/interface.py
Onager: Done

PR #2414, opened 2019-03-27
Update version for 20190331 release

Onager:


joachimmetz:
@Onager please bump version to 20190330 or 20190331

Issue #2371, opened 2019-02-23
image_export.py doesn't expand variables in artifact definitions


ant1:
**Description of problem:**

It seems image_export.py doesn't expand variables in artifact definitions

**Command line and arguments:**

This is with plaso installed on ubuntu from gift ppa.

```
$ log2timeline.py --version
plaso - log2timeline version 20190131

$ image_export.py --artifact_filters WindowsSystemRegistryFiles xp-tdungan-c-drive.E01 
2019-02-23 20:34:00,149 [INFO] (MainProcess) PID:7468 <data_location> Determined data location: /usr/share/plaso
2019-02-23 20:34:00,150 [INFO] (MainProcess) PID:7468 <artifact_definitions> Determined artifact definitions path: /usr/share/artifacts
Export started.
Extracting file entries from: TSK:/
2019-02-23 20:34:12,235 [WARNING] (MainProcess) PID:7468 <artifact_filters> The path filter must be defined as an absolute path: "%%environ_systemroot%%\System32\config\SAM"
2019-02-23 20:34:12,240 [WARNING] (MainProcess) PID:7468 <artifact_filters> The path filter must be defined as an absolute path: "%%environ_systemroot%%\System32\config\SECURITY"
2019-02-23 20:34:12,241 [WARNING] (MainProcess) PID:7468 <artifact_filters> The path filter must be defined as an absolute path: "%%environ_systemroot%%\System32\config\SOFTWARE"
2019-02-23 20:34:12,244 [WARNING] (MainProcess) PID:7468 <artifact_filters> The path filter must be defined as an absolute path: "%%environ_systemroot%%\System32\config\SYSTEM"
Export completed.

$ find export/
export/
```

log2timeline.py expands variables correctly for the WindowsSystemRegistryFiles artifact

joachimmetz:
It appears the preprocessor is not run: https://github.com/log2timeline/plaso/blob/master/plaso/cli/image_export_tool.py#L321

Since knowledgebase is never none

joachimmetz:
@ant1 changes to address this issue have been merged. Let us know if this now works for you

ant1:
This looks fine

```
# find export/ -type f
export/WINDOWS/system32/config/software
export/WINDOWS/system32/config/system
export/WINDOWS/system32/config/SAM
export/WINDOWS/system32/config/SECURITY
```

joachimmetz:
@ant1 thx for the quick confirmation, closing issue

PR #2421, opened 2019-03-29
Changes to fix preprocessing in image_export #2371

joachimmetz:
Changes to fix preprocessing in image_export #2371


Onager:
Merge after tests complete

PR #2426, opened 2019-03-29
Show total number of events in psort

joachimmetz:
Show total number of events in psort

### DIFF
--- a/plaso/cli/status_view.py
+++ b/plaso/cli/status_view.py
@@ -199,6 +199,10 @@ class StatusView(object):
         'Storage file\t\t: {0:s}\n'.format(self._storage_file_path))
 
     self._PrintProcessingTime(processing_status)
+
+    if processing_status and processing_status.events_status:
+      self._PrintEventsStatus(processing_status.events_status)
+
     self._output_writer.Write('\n')
 
   def _PrintAnalysisStatusUpdateLinear(self, processing_status):
@@ -306,6 +310,25 @@ class StatusView(object):
     if self._stdout_output_writer:
       # We need to explicitly flush stdout to prevent partial status updates.
       sys.stdout.flush()
+
+  def _PrintEventsStatus(self, events_status):
+    """Prints the status of the events.
+
+    Args:
+      events_status (EventsStatus): events status.
+    """
+    # TODO: print additional event status as "Events MACB grouped",
+    # "Duplicate events removed", "Events filtered"
+    if events_status:
+      table_view = views.CLITabularTableView(
+          column_names=['Events:', 'Total'],
+          column_sizes=[15, 0])
+
+      table_view.AddRow([
+          '', events_status.total_number_of_events])
+
+      self._output_writer.Write('\n')
+      table_view.Write(self._output_writer)
 
   def _PrintProcessingTime(self, processing_status):
     """Prints the processing time.
--- a/plaso/engine/processing_status.py
+++ b/plaso/engine/processing_status.py
@@ -313,6 +313,7 @@ class ProcessingStatus(object):
     aborted (bool): True if processing was aborted.
     error_path_specs (list[dfvfs.PathSpec]): path specifications that
         caused critical errors during processing.
+    events_status (EventsStatus): status information about events.
     foreman_status (ProcessingStatus): foreman processing status.
     start_time (float): time that the processing was started. Contains the
         number of micro seconds since January 1, 1970, 00:00:00 UTC.
@@ -326,6 +327,7 @@ class ProcessingStatus(object):
 
     self.aborted = False
     self.error_path_specs = []
+    self.events_status = None
     self.foreman_status = None
     self.start_time = time.time()
     self.tasks_status = None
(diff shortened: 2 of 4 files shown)


plaso/multi_processing/psort.py
```diff
@@ -742,7 +746,7 @@ def _UpdateProcessingStatus(self, pid, process_status, used_memory):
 
     process = self._processes_per_pid[pid]
 
-    processing_status = process_status.get('processing_status', None)
+    status_indicator = process_status.get('processing_status', None)
```
Onager: I think the renaming here is good, given the large number of "status" things (ProcessStatus vs ProcessingStatus), but maybe the definitions need to be renamed to, to STATUS_INDICATOR_IDLE etc. 
These definitions are sometimes called "status" and sometimes called "status_indicator" which is a bit confusing.

plaso/cli/status_view.py
```diff
@@ -307,6 +308,25 @@ def _PrintExtractionStatusUpdateWindow(self, processing_status):
       # We need to explicitly flush stdout to prevent partial status updates.
       sys.stdout.flush()
 
+  def _PrintEventsStatus(self, processing_status):
```
Onager: Just pass the events_status, rather than processing_status as it's not used.

plaso/cli/psort_tool.py
```diff
@@ -508,6 +508,10 @@ def ProcessStorage(self):
           self._storage_file_path))
       return
 
+    total_number_of_events = 0
+    for session in storage_reader.GetSessions():
```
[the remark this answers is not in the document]
Onager: Having read more, I think this functionality actually should move to ExportEvents in the psort engine. 
We already pass in the storage reader, and we don't use the total number of events in the tool other than to pass to there. It also means we don't need to change the ExportEvents signature.

plaso/engine/processing_status.py
```diff
@@ -508,6 +518,19 @@ def UpdateWorkerStatus(
         number_of_consumed_reports, number_of_produced_reports)
 
 
+class EventsStatus(object):
+  """The status of the events.
```
Onager: Please confirm you have plans to add more attributes to this class. Otherwise, we can just use an int and not add another layer of indirection.

plaso/engine/processing_status.py
```diff
@@ -508,6 +518,19 @@ def UpdateWorkerStatus(
...
+
+  Attributes:
+    total_number_of_events (int): total number of events.
```
Onager: Is this intended to be the total number of events in a store, or just a number of events in general? If the latter, drop the "total_", if something else, please update the docstring to clarify what this is a total of.

plaso/engine/processing_status.py
```diff
@@ -508,6 +518,19 @@ def UpdateWorkerStatus(
...
+  """
+
+  def __init__(self):
+    """Initializes a tasks status."""
```
Onager: -tasks +events

plaso/cli/psort_tool.py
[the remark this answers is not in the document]
joachimmetz: ack

plaso/cli/psteal_tool.py
```diff
@@ -191,6 +191,10 @@ def AnalyzeEvents(self):
           self._storage_file_path))
       return
 
+    total_number_of_events = 0
+    for session in storage_reader.GetSessions():
```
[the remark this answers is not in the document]
joachimmetz: ack

plaso/cli/status_view.py
joachimmetz: ack

plaso/engine/processing_status.py
joachimmetz: Yes I think we could add other information such as:

* Events MACB grouped
* Duplicate events removed
* Events filtered

plaso/cli/status_view.py
```diff
@@ -307,6 +308,25 @@ def _PrintExtractionStatusUpdateWindow(self, processing_status):
...
+    """Prints the status of the events.
```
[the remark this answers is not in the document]
joachimmetz: I opt to keep as-is to extend EventStatus as indicated below

plaso/engine/processing_status.py
joachimmetz: ack

plaso/multi_processing/psort.py
joachimmetz: Moved AI to separate issue https://github.com/log2timeline/plaso/issues/2437

plaso/engine/processing_status.py
Onager: Ping for @joachimmetz - I think you missed this comment

plaso/multi_processing/psort.py
```diff
@@ -975,6 +979,7 @@ def AnalyzeEvents(
     if keyboard_interrupt:
       raise KeyboardInterrupt
 
+  # pylint: disable=too-many-arguments
```
Onager: Why is this necessary? There's no change to this method, was this not being caught by the linter before for some reason? If not, please remove.

plaso/engine/processing_status.py
Onager: Extra comment to add to code review

joachimmetz: Total in storage file.

plaso/multi_processing/psort.py
joachimmetz: no longer needed.

Issue #2429, opened 2019-03-29
Registry only artifact filters errors with no valid specifications built


joachimmetz:
`log2timeline.py --artifact_filters WindowsProductName image.raw`

```
2019-03-29 15:49:22,831 [DEBUG] (MainProcess) PID:13666 <artifact_filters> building find spec from key path: HKEY_LOCAL_MACHINE\Software\Microsoft\Windows NT\CurrentVersion
2019-03-29 15:49:22,831 [WARNING] (MainProcess) PID:13666 <log2timeline> Unable to build filter specification: Error processing filters, no valid specifications built.
```

joachimmetz:
Expected behavior is that only winreg files are processed with the Registry find specs.

joachimmetz:
Side behavior is that the Registry files are parsed by filestat

PR #2431, opened 2019-03-29
Changes to handle Windows Registry artifact definitions #2429

joachimmetz:
Changes to handle Windows Registry artifact definitions #2429

### DIFF
--- a/plaso/engine/artifact_filters.py
+++ b/plaso/engine/artifact_filters.py
@@ -2,7 +2,6 @@
 """Helper to create filters based on forensic artifact definitions."""
 
 from __future__ import unicode_literals
-from collections import defaultdict
 
 from artifacts import definitions as artifact_types
 
@@ -19,9 +18,17 @@ class ArtifactDefinitionsFilterHelper(object):
 
   For more information about Forensic Artifacts see:
   https://github.com/ForensicArtifacts/artifacts/blob/master/docs/Artifacts%20definition%20format%20and%20style%20guide.asciidoc
-  """
 
-  KNOWLEDGE_BASE_VALUE = 'ARTIFACT_FILTERS'
+  Attributes:
+    file_system_artifact_names (set[str]): names of artifacts definitions that
+        generated file system find specifications.
+    file_system_find_specs (list[dfvfs.FindSpec]): file system find
+        specifications.
+    registry_artifact_names (set[str]): names of artifacts definitions that
+        generated Windows Registry find specifications.
+    registry_find_specs (list[dfwinreg.FindSpec]): Windows Registry find
+        specifications.
+  """
 
   _COMPATIBLE_REGISTRY_KEY_PATH_PREFIXES = frozenset([
       'HKEY_LOCAL_MACHINE\\SYSTEM',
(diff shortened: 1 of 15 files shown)


plaso/engine/artifact_filters.py
```diff
@@ -65,47 +74,38 @@ def CheckKeyCompatibility(key_path):
         'Prefix of key "{0:s}" is currently not supported'.format(key_path))
     return False
 
-  def BuildFindSpecs(self, environment_variables=None):
+  def BuildFindSpecs(self, artifact_filter_names, environment_variables=None):
```
Onager: We don't have to change it now, but this method isn't needed any more, and the logic can just move to the initializer.

plaso/engine/engine.py
```diff
@@ -317,17 +319,25 @@ def BuildFilterFindSpecs(
...
-      artifact_filters_object.BuildFindSpecs(
-          environment_variables=environment_variables)
-      find_specs = knowledge_base_object.GetValue(
-          artifact_filters_object.KNOWLEDGE_BASE_VALUE)[
-              artifact_types.TYPE_INDICATOR_FILE]
+              artifacts_registry_object, knowledge_base_object))
+      self._artifacts_filter_helper.BuildFindSpecs(
+          artifact_filter_names, environment_variables=environment_variables)
+
+      if self._artifacts_filter_helper.registry_find_specs:
```
Onager: Please add a comment here about why this is being done. The idea here is if the user wants a registry artifact, they probably also want to include the registry hive files, right?

Also, doesn't this logic belong in the artifact filter helper, given this is about constructing artifact filters? Please move it there.

plaso/parsers/mediator.py
```diff
@@ -22,26 +22,32 @@ class ParserMediator(object):
...
+    artifacts_filter_helper (ArtifactDefinitionsFilterHelper): artifacts
+        definitions filter helper.
     last_activity_timestamp (int): timestamp received that indicates the last
         time activity was observed. The last activity timestamp is updated
         when the mediator produces an attribute container, such as an event
         source. This timestamp is used by the multi processing worker process
         to indicate the last time the worker was known to be active. This
         information is then used by the foreman to detect workers that are
         not responding (stalled).
+    registry_find_specs (list[dfwinreg.FindSpec]): Windows Registry find
```
Onager: No such attribute, please remove.

plaso/engine/artifact_filters.py
joachimmetz: Let's chat about this, it not entirely clear to me what you mean.

plaso/engine/engine.py
joachimmetz: Let's keep it here for now, and have a longer conversation about this. I'll add the comment.

plaso/parsers/mediator.py
joachimmetz: ack, removed

plaso/engine/engine.py
Onager: OK, https://github.com/log2timeline/plaso/issues/2438

plaso/engine/artifact_filters.py
Onager: We can just pass artifact_filter_names and environment in __init__ and we don't need this method. In the engine, which is the only place ArtifactDefinitionsFilterHelper is initialized outside of tests, we call the methods directly after each other. 
Filed https://github.com/log2timeline/plaso/issues/2438 for this (and other changes).

PR #2434, opened 2019-03-30
Clean up and added presets file path to list parsers output #1662

joachimmetz:
Clean up and added presets file path to list parsers output #1662

### DIFF
--- a/plaso/cli/extraction_tool.py
+++ b/plaso/cli/extraction_tool.py
@@ -3,6 +3,8 @@
 
 from __future__ import unicode_literals
 
+import os
+
 from dfvfs.resolver import context as dfvfs_context
 
 # The following import makes sure the analyzers are registered.
@@ -14,6 +16,7 @@ from plaso import parsers  # pylint: disable=unused-import
 from plaso.cli import logger
 from plaso.cli import storage_media_tool
 from plaso.cli import tool_options
+from plaso.cli import views
 from plaso.cli.helpers import manager as helpers_manager
 from plaso.engine import configurations
 from plaso.engine import engine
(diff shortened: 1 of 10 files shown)


plaso/cli/extraction_tool.py
```diff
@@ -123,6 +127,24 @@ def _CreateProcessingConfiguration(self, knowledge_base):
 
     return configuration
 
+  def _GetParserPresetsInformation(self):
```
Onager: Please move this functionality to the ParserManager. There's already similar methods (GetParsersInformation) there for producing lists of tuples like this for parsers and plugins.

plaso/cli/extraction_tool.py
```diff
@@ -186,6 +208,19 @@ def _PreprocessSources(self, extraction_engine):
 
     logger.debug('Preprocessing done.')
 
+  def _ReadParserPresetsFromFile(self):
```
Onager: ReadPresetsFromFile() can also result in a uncaught MalformedPresetError from _ReadPresetsFromFileObject() that would be nice to catch here and raise as BadConfigOption as well - please add this functionality.

plaso/cli/extraction_tool.py
```diff
@@ -186,6 +208,19 @@ def _PreprocessSources(self, extraction_engine):
...
+    """Reads the parser presets from the presets.yaml file.
+
+    Raises:
+      BadConfigOption: if the parser presets file cannot be read.
+    """
+    self._presets_file = os.path.join(self._data_location, 'presets.yaml')
```
Onager: Please move 'presets.yaml' into a constant - _PRESET_FILE_NAME or similar.

plaso/cli/extraction_tool.py
joachimmetz: ack

plaso/cli/extraction_tool.py
joachimmetz: ack

plaso/cli/extraction_tool.py
joachimmetz: ack

PR #2436, opened 2019-03-30
Add beautifulsoup

Onager:


dependencies.ini
```diff
@@ -4,6 +4,13 @@ minimum_version: 20190305
 rpm_name: python2-artifacts
 version_property: __version__
 
+[beautifulsoup4]
```
joachimmetz: Have a loot at https://github.com/google/timesketch/pull/790/files#diff-32f6ea337b09b9bc597215dddf395d89R46

the name of the module is bs4 not beautifulsoup4

Onager: OK, the name doesn't match in l2tdevtools then.

joachimmetz: add `l2tbinaries_name: beautifulsoup4`

joachimmetz:
Per conversation, finding an alternative solution instead of depending on beautifulsoup

PR #2439, opened 2019-03-31
Added end-to-end test for artifact support

Onager:


config/jenkins/greendale/student-pc1-browserhistory.ini
```diff
@@ -0,0 +1,8 @@
+[studentpc1-browserhistory]
+case=extract_and_output
+source=student-pc1.dd
+extract_options=--partitions all --vss_stores all --artifact_filters BrowserHistory
```
joachimmetz: Maybe add some of the artifact definitions that has been causing some of the edge cases 'WindowsExplorerNamespaceCommonPlaces,WindowsAppInitDLLs,WindowsProductName,OperaHistory,WindowsBITSQueueManagerDatabases,MicrosoftAVQuarantine'

Onager: Maybe those would be better as unit tests? I'd like to keep this config to just be for BrowserHistory, since we care especially about that artifact.

PR #2442, opened 2019-04-01
Changes to psort to show event counters as status

joachimmetz:
```
plaso - psort version 20190329

Storage file		: test.plaso
Processing time		: 00:00:40

Events:         Filtered        In time slice   Duplicates      MACB grouped    Total
                249334          0               0               3               249338

Identifier              PID     Status          Memory          Events          Tags            Reports
Main                    174692  exporting       107.4 MiB       4 (3)           0 (0)           0 (0)

Processing completed.
```

### DIFF
--- a/plaso/engine/processing_status.py
+++ b/plaso/engine/processing_status.py
@@ -522,10 +522,12 @@ class EventsStatus(object):
   """The status of the events.
 
   Attributes:
-    number_of_duplicate_events (int): number of duplicate events.
+    number_of_duplicate_events (int): number of duplicate events, not including
+        the original.
     number_of_events_from_time_slice (int): number of events from time slice.
-    number_of_filtered_events (int): number of filtered events.
-    number_of_macb_grouped_events (int): number of MACB grouped events.
+    number_of_filtered_events (int): number of events excluded by the event
+        filter.
+    number_of_macb_grouped_events (int): number of events grouped based on MACB.
     total_number_of_events (int): total number of events in the storage file.
   """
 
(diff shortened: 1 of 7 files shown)


plaso/engine/processing_status.py
```diff
@@ -522,12 +522,20 @@ class EventsStatus(object):
   """The status of the events.
 
   Attributes:
+    number_of_duplicate_events (int): number of duplicate events.
```
Onager: Please add more verbose descriptions of what the counters are. Specifically:
- Do the duplicate events just count events beyond the first one? ie, if events A and B are duplicates, is this number 1 or 2? If A, B and C are duplicates, is this number 1, 2 or 3?
- Is the number of filtered events the number filtered out of the output, or the number filtered into the output?
- Is the number of MACB group events the number of events grouped together, or number of coalesced events? ie, if an Access and Modify events are grouped together, is this number 1 or 2?

joachimmetz: > Do the duplicate events just count events beyond the first one?

Yes

> ie, if events A and B are duplicates, is this number 1 or 2?

it contains 1, since A is seen as the original

> If A, B and C are duplicates, is this number 1, 2 or 3?

it contains 2, since A is seen as the original

> Is the number of filtered events the number filtered out of the output, or the number filtered into the output?

This is number of events excluded by the event filter.

> Is the number of MACB group events the number of events grouped together, or number of coalesced events? ie, if an Access and Modify events are grouped together, is this number 1 or 2?

I assume with "coalesced events" you are referring to the resulting MACB group?
This value contains the number of events grouped, not the number of resulting groups.

Issue #2381, opened 2019-03-09
Firmware file causes utmp extraction warnings


Onager:
Noticed in 'acserver' Greendale image:
```
********************************* Warning: 17 **********************************
           Message : unable to decode hostname string
      Parser chain : utmp
Path specification : type: OS, location: /media/greendale_images/acserver.dd
                   : type: RAW
                   : type: TSK_PARTITION, location: /p1, part index: 2, start
                     offset: 0x00100000
                   : type: TSK, inode: 139450, location:
                     /lib/firmware/matrox/g400_warp.fw
```

g400_warp.fw is not a utmp file, but is being recognized as such.

Onager:
This looks to be due to utmp files not having a good strong signature, so arbitrary binary files can readily pass the verification the utmp parser attempts.

joachimmetz:
https://github.com/log2timeline/plaso/pull/2430

joachimmetz:
Changes pushed, but likely more improvements possible

joachimmetz:
@Onager close this issue? is there more you want to change in the context of this issue?

Onager:
The extra checks have stopped the misleading warnings.

PR #2459, opened 2019-04-15
Added additional checks for first utmp record validity #2381

Onager:


joachimmetz:
I need to double check my test files. If I recall correctly the additional fields you test for presence are not necessarily filled

joachimmetz:
Hostname can be legitimate empty

```
Type of login                                                           : 0x00000006 (LOGIN_PROCESS)
PID                                                                     : 1115
Terminal                                                                : tty4
Terminal ID                                                             : 52
Username                                                                : LOGIN
Hostname                                                                : 
Termination status                                                      : 0x0000
Exit status                                                             : 0x0000
Session                                                                 : 1115
Timestamp                                                               : 2013-12-13 14:45:09 UTC
Microseconds                                                            : 0
IP address                                                              : 0.0.0.0
```

joachimmetz:
username can be legitimate empty

```
Type of login                                                           : 0x00000008 (DEAD_PROCESS)
PID                                                                     : 1823
Terminal                                                                : tty2
Terminal ID                                                             : 0
Username                                                                : 
Hostname                                                                : /dev/tty2
Termination status                                                      : 0x0000
Exit status                                                             : 0x0000
Session                                                                 : 0
Timestamp                                                               : 2018-05-17 10:23:22 UTC
Microseconds                                                            : 883446
IP address                                                              : 0.0.0.0
```

plaso/parsers/utmp.py
```diff
@@ -162,13 +162,22 @@ def ParseFileObject(self, parser_mediator, file_object):
...
+
+    if not event_data.hostname:
+      raise errors.UnableToParseFile(
+          'Unable to parse first utmp entry with error: missing hostname')
+
+    if not event_data.terminal:
+      raise errors.UnableToParseFile(
+          'Unable to parse first utmp entry with error: missing terminal')
+
     if not event_data.username:
```
joachimmetz: Please add check to see if `type of login != DEAD_PROCESS`

plaso/parsers/utmp.py
```diff
@@ -162,13 +162,22 @@ def ParseFileObject(self, parser_mediator, file_object):
           'Unable to parse first utmp entry with error: {0!s}'.format(
               exception))
 
+    if not timestamp:
```
joachimmetz: since both username and hostname can be legitimately empty do the following:

if the `type of login` of the first record is 0x00000001 (RUN_LVL) or 0x00000002 (0x00000002)
* username is not empty and in "shutdown", "reboot", "runlevel"
* terminal should be "~"

if the `type of login` is 0 (EMPTY)
* all should be empty?

plaso/parsers/utmp.py
Onager: Done.

plaso/parsers/utmp.py
Onager: I thought about this a bit more, and realized the real issue here was the decode errors being caught and emitted as warnings in _ReadEntry. I've changed this around, but I don't think we need to add more validation here just yet.

Issue #2478, opened 2019-04-30
Extraction warnings when parsing Apache access logs


cshanahan:
**Description of problem:**
Plaso doesn't consistently parse Apache access log files in common or combined format. In two recent investigations, Plaso didn't parse Apache access log files at all. Both access log files were in common format ("%h %l %u %t \"%r\" %>s %b"). In those two cases, the error message was the same, wrong file structure. Though the fields were correct.
`<extractors> apache_access unable to parse file: OS:/locker/REDACTED/access.log with error: Wrong file structure.`

I performed three separate tests using publicly available data. All three tests were performed using Plaso versions 20190131 and 20190331. Tests 1 and 2 failed with both Plaso versions, but with different error messages. Test 3 succeeded with both Plaso versions, though with a number of errors (300 errors associated with 150 log lines). For example, the following log line failed to parse:
```
24.225.105.158 - - [29/Apr/2019:17:04:40 +0000] "GET /category/garden HTTP/1.1" 200 40 "http://www.google.com/search?ie=UTF-8&q=google&sclient=psy-ab&q=Garden+Toys&oq=Garden+Toys&aq=f&aqi=g-vL1&aql=&pbx=1&bav=on.2,or.r_gc.r_pw.r_qf.,cf.osb&biw=2281&bih=371" "Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 5.1; Trident/4.0; YTB720; GTB7.2; .NET CLR 1.1.4322; .NET CLR 2.0.50727; .NET CLR 3.0.4506.2152; .NET CLR 3.5.30729)"
```

The two errors associated with this log line were:

```
********************************** Error: 200 **********************************
           Message : unable to parse log line: u'24.225.105.158 - -
                     [29/Apr/2019:17:04:40 +0000] "GET /category/garden
                     HTTP/1....' at offset: 2690567
      Parser chain : apache_access
Path specification : type: OS, location:
                     /locker/Apache_Log_Test/Test03/apache-daily-access.log
--------------------------------------------------------------------------------

********************************** Error: 201 **********************************
           Message : unable to parse log line: u'; .NET CLR 3.5.30729)"' at
                     offset: 2690967
      Parser chain : apache_access
Path specification : type: OS, location:
                     /locker/Apache_Log_Test/Test03/apache-daily-access.log
--------------------------------------------------------------------------------
```

**Command line and arguments:**
**Test 01**
[...]

PR #2486, opened 2019-05-04
Improved Apache accesss.log parser #2478

joachimmetz:
Improved Apache accesss.log parser


plaso/parsers/apache_access.py
```diff
@@ -54,24 +54,17 @@ class ApacheAccessParser(text_parser.PyparsingSingleLineTextParser):
   NAME = 'apache_access'
   DESCRIPTION = 'Apache access Parser'
 
+  MAX_LINE_LENGTH = 2048
+
   _PYPARSING_COMPONENTS = {
```
Onager: Please keep the pyparsing components here, to be consistent with other pyparsing parsers.

joachimmetz: Why? this is addition indirection, which appears to not be used for any other purpose

joachimmetz:
> Moving the parsing components around makes it more difficult to track what the actual improvements here are.
> 
> It looks like the only functional change is increasing the maximum line length, and the rest is refactoring, is that correct?

Per #2478 there are 2 changes:

* increased max line length
* and changed grammar to support "response bytes" value of `-`

@Onager PTAL

plaso/parsers/apache_access.py
Onager: To be consistent with the other pyparsing parsers. If you don't like the pyparsing_components constant, why are remote_name and user_name still defined there?

tests/parsers/apache_access.py
```diff
@@ -22,15 +22,15 @@ def testParse(self):
     parser = apache_access.ApacheAccessParser()
     storage_writer = self._ParseFile(['access.log'], parser)
 
-    self.assertEqual(storage_writer.number_of_warnings, 0)
-    self.assertEqual(storage_writer.number_of_events, 5)
+    self.assertEqual(storage_writer.number_of_warnings, 1)
```
Onager: Please add a test to validate the warning.

plaso/parsers/apache_access.py
joachimmetz: Because I did not had to touch them. I can change them. Agree that consistency is important but ` _PYPARSING_COMPONENTS` was never was introduces across all pyparsing based parsers e.g. https://github.com/log2timeline/plaso/blob/master/plaso/parsers/xchatlog.py

If you say this indirection is not used I'll create an issue to remove it everywhere.

tests/parsers/apache_access.py
joachimmetz: ack

PR #2489, opened 2019-05-05
Added event data support to analysis plugin ExamineEvent method

joachimmetz:
Added event data support to analyzers ExamineEvent method

### DIFF
--- a/plaso/analysis/browser_search.py
+++ b/plaso/analysis/browser_search.py
@@ -355,7 +355,6 @@ class BrowserSearchPlugin(interface.AnalysisPlugin):
       return
 
     parser_or_plugin_name = getattr(event_data, 'parser', 'N/A')
-    parser_or_plugin_name = getattr(event_data, 'plugin', parser_or_plugin_name)
 
     for engine, url_expression, method_name in self._URL_FILTERS:
       callback_method = getattr(self, method_name, None)
--- a/plaso/analysis/browser_search.py
+++ b/plaso/analysis/browser_search.py
@@ -17,7 +17,6 @@ from plaso.analysis import interface
 from plaso.analysis import logger
 from plaso.analysis import manager
 from plaso.containers import reports
-from plaso.formatters import manager as formatters_manager
 from plaso.lib import py2to3
 
 
@@ -38,6 +37,37 @@ class BrowserSearchPlugin(interface.AnalysisPlugin):
 
   _EVENT_TAG_COMMENT = 'Browser Search'
   _EVENT_TAG_LABELS = ['browser_search']
+
+  _SUPPORTED_EVENT_DATA_TYPES = frozenset([
+      'chrome:autofill:entry',
+      'chrome:cache:entry',
+      'chrome:cookie:entry',
+      'chrome:extension_activity:activity_log',
+      'chrome:history:file_downloaded',
+      'chrome:history:page_visited',
+      'firefox:cache:record',
+      'firefox:cookie:entry',
+      'firefox:places:bookmark_annotation',
+      'firefox:places:bookmark_folder',
+      'firefox:places:bookmark',
+      'firefox:places:page_visited',
+      'firefox:downloads:download',
+      'cookie:google:analytics:utma',
+      'cookie:google:analytics:utmb',
+      'cookie:google:analytics:utmt',
+      'cookie:google:analytics:utmz',
+      'msiecf:leak',
+      'msiecf:redirected',
+      'msiecf:url',
+      'msie:webcache:container',
+      'msie:webcache:containers',
+      'msie:webcache:leak_file',
+      'msie:webcache:partitions',
+      'opera:history:entry',
+      'opera:history:typed_entry',
+      'safari:cookie:entry',
+      'safari:history:visit',
+      'safari:history:visit_sqlite'])
 
   # TODO: use groups to build a single RE.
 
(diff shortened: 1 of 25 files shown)


plaso/analysis/browser_search.py
```diff
@@ -39,6 +38,37 @@ class BrowserSearchPlugin(interface.AnalysisPlugin):
   _EVENT_TAG_COMMENT = 'Browser Search'
   _EVENT_TAG_LABELS = ['browser_search']
 
+  _SUPPORTED_EVENT_DATA_TYPES = frozenset([
```
Onager: I don't think this is necessary any more, we can just check for a url attribute. I know this was check for WEBHIST before, but this doesn't add much.

plaso/analysis/browser_search.py
```diff
@@ -308,26 +338,25 @@ def CompileReport(self, mediator):
...
-    # TODO: refactor this the source should be used in formatting only.
-    # Check if we are dealing with a web history event.
-    source, _ = formatters_manager.FormattersManager.GetSourceStrings(event)
-
-    if source != 'WEBHIST':
+    url = getattr(event_data, 'url', None)
+    if not url:
       return
 
+    parser_or_plugin_name = getattr(event_data, 'parser', 'N/A')
```
Onager: parser is set on event, not event_data: https://github.com/log2timeline/plaso/blob/master/plaso/parsers/mediator.py#L412

plaso/analysis/browser_search.py
```diff
@@ -308,26 +338,25 @@ def CompileReport(self, mediator):
...
+    parser_or_plugin_name = getattr(event_data, 'plugin', parser_or_plugin_name)
```
Onager: This can be removed, plugin isn't set any more, it's in the parser attribute.

plaso/analysis/interface.py
```diff
@@ -198,37 +199,38 @@ def _EnsureRequesterStarted(self):
...
     """
+    if (event_data.data_type not in self.DATA_TYPES or
+        not self._analyzer.lookup_hash):
+      return
+
     self._EnsureRequesterStarted()
 
-    path_spec = event.pathspec
-    event_identifiers = self._event_identifiers_by_pathspec[path_spec]
+    event_identifiers = self._event_identifiers_by_pathspec[event_data.pathspec]
```
Onager: Pathspec is set on the event, not the data: https://github.com/log2timeline/plaso/blob/master/plaso/parsers/mediator.py#L423

plaso/analysis/windows_services.py
```diff
@@ -111,27 +111,28 @@ def __eq__(self, other_service):
...
-
-    if service_event.pathspec:
-      source = (service_event.pathspec.location, service_event.key_path)
+    service_type = event_data.regvalue.get('Type', '')
+    image_path = event_data.regvalue.get('ImagePath', '')
+    start_type = event_data.regvalue.get('Start', '')
+    service_dll = event_data.regvalue.get('ServiceDll', '')
+    object_name = event_data.regvalue.get('ObjectName', '')
+
+    if event_data.pathspec:
```
Onager: Pathspec is set on event, not event data

tests/analysis/test_lib.py
```diff
@@ -36,32 +35,51 @@ def _AnalyzeEvents(
...
+      plugin.ExamineEvent(mediator, event, event_data)
 
     analysis_report = plugin.CompileReport(mediator)
     storage_writer.AddAnalysisReport(analysis_report)
 
     return storage_writer
 
-  def _CreateTestEventObject(self, event_dictionary):
-    """Create a test event with a set of attributes.
+  def _CreateTestEvent(self, event_values):
```
Onager: This needs some refactoring, as there's a bunch of attributes that are set on the event_data that aren't set in normal operation - see the notes about filename and parser above.

plaso/analysis/browser_search.py
Onager: After looking at this a bit more, I think this is actually set on the event_data after all: https://github.com/log2timeline/plaso/blob/master/plaso/parsers/mediator.py#L503 but it's quite confusing from the method naming etc.

Also, I'm not sure whether "parser" should be set on the event_data or the event or both. It doesn't seem to be defined on either though: https://github.com/log2timeline/plaso/blob/master/plaso/containers/events.py and it should be defined _somewhere_.

plaso/analysis/interface.py
Onager: Per above, it looks like the pathspec is set on EventData, but it's not defined there: https://github.com/log2timeline/plaso/blob/master/plaso/containers/events.py

Can you please clarify what's intended here? It would also be helpful to have a clear statement about what sort of data belongs in each other. If I have the same file, copy it to a different machine and rename it, should the eventdatas produced from the files be the same? Can both the events and eventdatas be different?

plaso/analysis/browser_search.py
joachimmetz: Let's make these changes separate from this PR. Maybe first do the event data attribute standarization.

https://github.com/log2timeline/plaso/issues/2530

plaso/analysis/browser_search.py
joachimmetz: > but it's quite confusing from the method naming etc.

Agree that's why I'm cleaning it up

joachimmetz: > I'm not sure whether "parser" should be set on the event_data 

It is https://github.com/log2timeline/plaso/blob/master/plaso/parsers/mediator.py#L412

This might be a bit hard to follow (and needs cleanup, hence this PR) 

But it is passed event data from https://github.com/log2timeline/plaso/blob/master/plaso/parsers/mediator.py#L503

plaso/analysis/browser_search.py
joachimmetz: ack, removing

plaso/analysis/interface.py
joachimmetz: > Pathspec is set on the event, not the data: 

It is set on event data at https://github.com/log2timeline/plaso/blob/master/plaso/parsers/mediator.py#L423

This might be a bit hard to follow (and needs cleanup, hence this PR)

But it is passed event data from https://github.com/log2timeline/plaso/blob/master/plaso/parsers/mediator.py#L503

> Can you please clarify what's intended here?

it should be set on event data, event should only have a restrictive set of attributes event should be something like:
(timestamp, timestamp_desc, event_data)

plaso/analysis/windows_services.py
joachimmetz: see previous comment, it is set on event data but that might not be directly obvious from the code, and agree needs clean up

tests/analysis/test_lib.py
joachimmetz: ack, I opt to this as part of https://github.com/log2timeline/plaso/issues/2524

Issue #324, opened 2015-09-01
Clean up: filters


joachimmetz:
The code relevant to the filters in plaso needs a considerable overhaul:
- [x] ~~[Clean up: remove lib registry](https://github.com/log2timeline/plaso/issues/154)~~
- [x] ~~Fix https://github.com/log2timeline/plaso/issues/56~~
- [x] ~~move LoggingFilter from frontend/log2timeline.py to filters/ - not needed different type of filter~~
- [x] ~~remove efilter~~
  * ~~https://github.com/log2timeline/plaso/pull/2338~~
- [x] ~~change WrongPlugin exception in filter/*.py to something more appropriate~~
  * ~~https://github.com/log2timeline/plaso/pull/2338, https://github.com/log2timeline/plaso/pull/2346, https://github.com/log2timeline/plaso/pull/2347, https://github.com/log2timeline/plaso/pull/2348~~
- [x] ~~clean up docstrings filters/*.py~~
  * ~~https://github.com/log2timeline/plaso/pull/2338, https://github.com/log2timeline/plaso/pull/2346, https://github.com/log2timeline/plaso/pull/2347, https://github.com/log2timeline/plaso/pull/2348~~
- [x] ~~clean up of lexer https://github.com/log2timeline/plaso/pull/2493, https://github.com/log2timeline/plaso/pull/2496, https://github.com/log2timeline/plaso/pull/2498~~
- [x] ~~Clean up plaso/lib/objectfilter.py and plaso/lib/pfilter.py~~
  * ~~move lexer code into object filter https://github.com/log2timeline/plaso/pull/2499~~
  * ~~move objectfilter code into pfilter https://github.com/log2timeline/plaso/pull/2507~~
  * ~~rename objectfilter into event filter https://github.com/log2timeline/plaso/pull/2508~~
  * ~~split ast and parser classes https://github.com/log2timeline/plaso/pull/2509~~
- [x] ~~Clean up plaso/filters/~~
  * ~~merge value expanders into single class https://github.com/log2timeline/plaso/pull/2510~~
  * ~~remove "filter implementation" indirection https://github.com/log2timeline/plaso/pull/2511~~
  * ~~clean up https://github.com/log2timeline/plaso/pull/2512~~
  * ~~remove "context operator"; it is not used by [event filter](https://plaso.readthedocs.io/en/latest/sources/user/Event-filters.html) https://github.com/log2timeline/plaso/pull/2513~~
[...]

PR #2493, opened 2019-05-06
Clean up of event filter expression parser (#324)

joachimmetz:
Clean up of event filter expression parser

* removed SelfFeederMixIn
* merged Lexer and SearchParser classes
* changes to docstrings

### DIFF
--- a/plaso/lib/lexer.py
+++ b/plaso/lib/lexer.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
-"""An LL(1) lexer. This lexer is very tolerant of errors and can resync.
+"""A parser for event filter expressions.
 
-This lexer is originally copied from the GRR project:
-https://code.google.com/p/grr
+The parser is based on the lexer from the GRR project:
+https://github.com/google/grr
 """
 
 from __future__ import unicode_literals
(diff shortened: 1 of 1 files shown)


joachimmetz:
@Onager PTAL, if you have comments on the docstrings, please add a suggestion, some of them are educated guesses. The idea is to clean up this file more and improve the docstrings in the process.

plaso/lib/lexer.py
```diff
@@ -12,26 +12,22 @@
...
 
     Args:
-
-      state_regex: If this regular expression matches the current state this
-                   rule is considered.
-      regex: A regular expression to try and match from the current point.
-      actions: A command separated list of method names in the Lexer to call.
-      next_state: The next state we transition to if this Token matches.
-      flags: re flags.
+      state_regex (str): regular expression that is considered when the current
```
Onager: This is an re, not a string. The type seems to in be re._pattern_type, but the exact name looks different on Python 2 and 3.

plaso/lib/lexer.py
```diff
@@ -12,26 +12,22 @@
...
+          state matches this rule.
+      regex (str): regular expression to try and match from the current point.
```
Onager: This is an re too, not a string.

plaso/lib/lexer.py
joachimmetz: this will be removed in a later change

plaso/lib/lexer.py
joachimmetz: this will be made protected in a later change

PR #2497, opened 2019-05-07
Added utility script to export event data schema

joachimmetz:
Added utility script to export event data schema


joachimmetz:
@aguilajesus per our conversation this is the PR with the script, if you find prior art that is related either point me at it or feel free to send a PR with improvements to the script

joachimmetz:
> Not sure what was the intent, and this could be follow up PRs, but would you consider:

The intent is to provide an overview of the attribute names of event data objects for event filters.

> parse the docstring and display the attribute types too

I have thought about this, this might be easier once we're at Python 3 with type information

PR #2498, opened 2019-05-08
Worked on tests for event filter expression parser and clean up (#324)

joachimmetz:
Worked on tests for event filter expression parser and clean up

### DIFF
--- a/plaso/lib/lexer.py
+++ b/plaso/lib/lexer.py
@@ -7,6 +7,7 @@ https://github.com/google/grr
 
 from __future__ import unicode_literals
 
+import abc
 import logging
 import re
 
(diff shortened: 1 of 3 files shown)


plaso/lib/lexer.py
```diff
@@ -139,18 +151,20 @@ def AddOperands(self, lhs, rhs):
...
           is not an instance of Expression.
     """
-    if isinstance(lhs, Expression) and isinstance(rhs, Expression):
-      self.args = [lhs, rhs]
-    else:
-      raise errors.ParseError(
-          'Expected expression, got {0:s} {1:s} {2:s}'.format(
-              lhs, self.operator, rhs))
+    if not isinstance(lhs, Expression):
+      raise errors.ParseError('Left hand side is not an expression')
```
Onager: Please add information about what lhs and rhs are to these error messages

joachimmetz: Not sure what you mean, but let's chat about this in the clean up. Adding a note for now.

Issue #2491, opened 2019-05-05
Change pyparsing parsers to return none instead of empty bytes values when no match for token


joachimmetz:
Some pyparsing parsers seem to return empty bytes values when pyparsing has no match for the value. Seen in syslog and syslog cron.

* [x] ~~syslog https://github.com/log2timeline/plaso/pull/2500~~
* [x] ~~single line parsers https://github.com/log2timeline/plaso/pull/2538~~
* [x] ~~mutli line parsers https://github.com/log2timeline/plaso/pull/2542~~

joachimmetz:
For context see: https://travis-ci.org/log2timeline/plaso/jobs/528483547

ptmcg:
Let me know if I can be of any help with this

joachimmetz:
@ptmcg thx what needs to be done is:

* determine which parsers and which fields generate `b''` empty values when there is no value
* changes these parsers to generate `None` instead

ptmcg:
Look over these changes in my forked repo: https://github.com/ptmcg/plaso/commit/0cc9145631fbaccfaf805de4d3e90ede2c635749 If this will work for you I'll work up a PR.

joachimmetz:
@ptmcg thx for the suggestion, we're happy to accept PRs

Though I'm hesitant to use the solution you propose. `PyParseInsertDefaultNone` seems to to involve manipulating Python internal methods. With the plaso codebase we're very cautious about debug-ability and maintainability, and shun for using Python "magic" as much as we can. For context also see: https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md

There seems to be a more straight forward approach e.g. https://github.com/log2timeline/plaso/blob/master/plaso/parsers/syslog.py#L314

`event_data.severity = structure.severity` if we change this into `event_data.severity = structure.get('severity', None)` it appears to have the desired result.

joachimmetz:
Issue seems to be cause by accessing a non-exiting token in a Pyparsing structure when using `structure.name` instead of `structure.get(name, None)`

ptmcg:
That works fine too, if you prefer to go that route. I was just trying to address your request, "change these parsers to generate None instead." For the scope of this issue, it looks like it affects any data field that is parsed using `Optional`. You can use my changeset as a guide for where these have related results names, and then use `.get('name', None)` to get None values if the particular attribute is not present in the parsed data.

[...]

PR #2500, opened 2019-05-08
Changes to syslog parser for handling empty bytes values #2491

joachimmetz:
Changes to syslog parser for handling empty bytes values #2491

### DIFF
--- a/plaso/parsers/text_parser.py
+++ b/plaso/parsers/text_parser.py
@@ -251,6 +251,24 @@ class PyparsingSingleLineTextParser(interface.FileObjectParser):
     # a structural fix.
     self._line_structures = list(self.LINE_STRUCTURES)
 
+  def _GetValueFromStructure(self, structure, name, default_value=None):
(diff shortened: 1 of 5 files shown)


plaso/parsers/text_parser.py
```diff
@@ -251,6 +251,24 @@ def __init__(self):
     # a structural fix.
     self._line_structures = list(self.LINE_STRUCTURES)
 
+  def _GetValueFromStructure(self, structure, name, default_value=None):
```
rgayon: Is this additional level of indirection really needed?
I feel like having tructure.get('hostname', None) would be explicit enough (as this is also how we access dict values)

joachimmetz: the reason for this is the docstring is used to explain why this is needed, instead of putting comments in the different parsers
IMHO this is one of those things that is easy to forget

rgayon: SGTM

PR #2516, opened 2019-05-11
Cleaned up docstrings of path filter

joachimmetz:
Cleaned up docstrings of path filter

### DIFF
--- a/plaso/filters/path_filter.py
+++ b/plaso/filters/path_filter.py
@@ -617,7 +617,7 @@ class PathFilterScanTreeNode(object):
 
     Returns:
       str|PathFilterScanTreeNode: a scan object, which is either
-          a scan tree sub node, a path of the default value.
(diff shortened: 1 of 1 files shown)


plaso/filters/path_filter.py
```diff
@@ -593,24 +609,24 @@ def AddPathSegment(self, path_segment, scan_object):
...
 
     Args:
-      path_segment: a string containing the path segment.
+      path_segment (str): path segment.
 
     Returns:
-      A scan object, which can be a scan tree sub node (instance of
-      PathFilterScanTreeNode), a path or the default value.
+      str|PathFilterScanTreeNode: a scan object, which is either
+          a scan tree sub node, a path of the default value.
```
rgayon: I don't really undersand. "a path to the default value" maybe?

joachimmetz: sry this should be "or" not "of"

Issue #1537, opened 2017-12-01
Allow filter file to be used as either include or exclude list


joachimmetz:
The filter file is currently used to indicate which paths should be included in extraction, however a recurring feature request is to be able to exclude paths from extraction.

joachimmetz:
Bumped to next release

joachimmetz:
Unlikely to be implemented before Jan 25, bumping release

Onager:
Not going to make release, removing milestone.

joachimmetz:
Additional work moved to next milestone

rodgermoore:
Just to let you know this is great! 🥇

PR #2519, opened 2019-05-11
Added YAML-based filter file support (#1537)

joachimmetz:
Added YAML-based filter file support


joachimmetz:
For reviewer this is first PR in a batch: https://github.com/log2timeline/plaso/issues/1537

joachimmetz:
After conversation with @rgayon and @Onager I've remove suffixes and prefixes

technically paths are prefixes already and suffixes are ambiguous, so a more specific find spec definition might be preferable.

PR #2521, opened 2019-05-12
Created collection filters helper and added exclusion path filters (#1537)

joachimmetz:
Created collection filters helper and added exclusion path filters


plaso/engine/path_filters.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""Path filter.
+"""Path filters.
 
 Path filter are specified in filter files and are used during collection
```
rgayon: *filters

PR #2522, opened 2019-05-12
Changes to not catch exception related to coding errors

joachimmetz:
Changes to not catch exception related to coding errors

### DIFF
--- a/plaso/engine/single_process.py
+++ b/plaso/engine/single_process.py
@@ -64,6 +64,7 @@ class SingleProcessEngine(engine.BaseEngine):
         self._file_system_cache.remove(file_system)
         self._file_system_cache.append(file_system)
 
+  # pylint: disable=missing-raises-doc
   def _ProcessPathSpec(self, extraction_worker, parser_mediator, path_spec):
     """Processes a path specification.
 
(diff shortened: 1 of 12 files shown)


Onager (changes requested):
Can you explain the rationale for this? Why would we want to crash at runtime if there's a bug?

joachimmetz:
> Can you explain the rationale for this? Why would we want to crash at runtime if there's a bug?

Pylint and "compile time" can catch these errors when not absorbed by the try except

plaso/engine/single_process.py
```diff
@@ -49,10 +50,16 @@ def _ProcessPathSpec(self, extraction_worker, parser_mediator, path_spec):
       excluded_find_specs = (
           self.collection_filters_helper.excluded_file_system_find_specs)
 
+    # pylint: disable=try-except-raise
     try:
       extraction_worker.ProcessPathSpec(
           parser_mediator, path_spec, excluded_find_specs=excluded_find_specs)
 
+    # Raise on coding errors.
+    except (AttributeError, ImportError, NameError, TypeError,
```
Onager: Please factor this tuple out a constant somewhere (lib/errors? lib/defintions?)

Onager (changes requested):

Onager:
> I've noticed a couple of time (not specific to plaso) that the broad try except absorbs some exception that could have been caught by the tests. Which then later on lead to other subtle issues.

I can see this happening in tests, but I don't understand could catch any more problems with a specific try-catch block. Can you give an example?

plaso/engine/single_process.py
joachimmetz: done

joachimmetz:
@Onager PTAL

joachimmetz:
Obsolete

Issue #2503, opened 2019-05-09
plaso/dependencies.py fix backports.lzma fallback


joachimmetz:
plaso/dependencies.py only detects lzma add support to fallback on backports.lzma

this can lead to a confusing warning about lzma missing, but it is not.

Python 3 provides lzma, while Python 2 backports.lzma (or lzma)

joachimmetz:
https://github.com/log2timeline/plaso/pull/2523

PR #2523, opened 2019-05-12
Changes to dependency check for lzma #2503

joachimmetz:
Changes to dependency check for lzma #2503

To do determine why MacOS travis tests are failing https://travis-ci.org/log2timeline/plaso/jobs/532113776

### DIFF
--- a/utils/dependencies.py
+++ b/utils/dependencies.py
@@ -147,7 +147,7 @@ class DependencyHelper(object):
       dependency (DependencyDefinition): dependency definition.
 
     Returns:
-      tuple: consists:
+      tuple: containing:
 
         bool: True if the Python module is available and conforms to
             the minimum required version, False otherwise.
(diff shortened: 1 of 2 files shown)


utils/dependencies.py
```diff
@@ -252,6 +252,38 @@ def _CheckPythonModuleVersion(
     status_message = '{0:s} version: {1!s}'.format(module_name, module_version)
     return True, status_message
 
+  def _CheckLZMA(self):
+    """Checks the availability of lzma.
+
+    Returns:
+      tuple: consists:
```
rgayon: I haven't seen this way of describing tuples in the doc before. Is this in the styleguide?

joachimmetz: I meant "containing", thx for flagging. Regarding the style guide have a look at: https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md#multiple-return-types

joachimmetz:
Determine why MacOS travis tests are failing https://travis-ci.org/log2timeline/plaso/jobs/532113776

PR #2542, opened 2019-05-24
Changes to multi line pyparsing parsers to handle empty bytes values #2491

joachimmetz:
Changes to multi line pyparsing parsers to handle empty bytes values #2491


plaso/parsers/gdrive_synclog.py
```diff
@@ -140,21 +148,27 @@ def _ParseRecordLogline(self, parser_mediator, structure):
     """
     date_time = dfdatetime_time_elements.TimeElementsInMilliseconds()
 
+    date_time_string = self._GetValueFromStructure(structure, 'date_time')
```
rgayon: Nit: "date_time_tokens" maybe, instead of "date_time_string"  ?
or "date_time_elements"

Issue #444, opened 2015-12-15
Create configuration file driven formatters


joachimmetz:
Take the formatters to the next level and make them configuration file driven

- [x] ~~create a yaml file that defines the formatters, should support:~~
  - ~~formatter definition with type, which maps to formatter class~~
  - ~~formatters file reader https://github.com/log2timeline/plaso/pull/2562~~
  - ~~moved AMCache formatter to configuration https://github.com/log2timeline/plaso/pull/2563~~
- [x] ~~considering storing formatter definitions in separate files~~
  - ~~added directory support https://github.com/log2timeline/plaso/pull/2853~~
  - ~~switch to directory as default https://github.com/log2timeline/plaso/pull/2854~~
  - ~~split generic.yaml https://github.com/log2timeline/plaso/pull/2855~~
- [x] ~~create formatter helpers~~
  - ~~enumeration helper https://github.com/log2timeline/plaso/pull/2869~~
  - ~~flags helper https://github.com/log2timeline/plaso/pull/3207~~
  - ~~boolean helper~~
- [x] ~~extend configuration to support helpers~~
  - ~~added support for enumeration helper https://github.com/log2timeline/plaso/pull/3126~~
  - ~~added support for flags helper https://github.com/log2timeline/plaso/pull/3208~~
- [x] ~~document formatters configuration file format~~
  - ~~added basic information - https://github.com/log2timeline/plaso/pull/2868~~
  - ~~update documentation for enumeration helper https://github.com/log2timeline/plaso/pull/3204~~
  - ~~update documentation for flags helper~~
- [x] ~~move formatters to configuration~~
  - ~~remove need for FORMAT_STRING_ALTERNATIVE https://github.com/log2timeline/plaso/issues/2837~~
  - ~~make sure parser tests, test formatting~~
  - ~~add support to set message pieces separator in configuration~~
  - ~~move remaining formatters that use enumeration helper to configuration~~
  - ~~move remaining formatters that use flags helper to configuration~~
- [x] ~~add formatters configuration file validator - https://github.com/log2timeline/plaso/issues/3388~~
- [x] create a way to define custom classes/callback e.g. winevt-rc.db - https://github.com/log2timeline/plaso/pull/3377
- [x] ~~remove `from plaso import formatters` - https://github.com/log2timeline/plaso/pull/3391~~
- [x] clean up pylint invalid name override in `plaso/formatters/yaml_formatters_file.py` - https://github.com/log2timeline/plaso/pull/3434

PR #2563, opened 2019-05-28
Moved AMCache formatter to configuration #444

joachimmetz:
Moved AMCache formatter to configuration #444


joachimmetz:
@Onager PTAL

Issue #2566, opened 2019-05-28
deprecate regvalue


joachimmetz:
Originally from: https://github.com/log2timeline/plaso/issues/1626

remove regvalue, also address:
* remove URLS from registry events (if they are needed re-add them to the formatters)
* replace source append by specific event data
* some plugins look for a single value and then store the additional values in regvalue

To do:
* [x] ~~bagmru https://github.com/log2timeline/plaso/pull/2569~~
* [x] ~~CCleaner https://github.com/log2timeline/plaso/pull/2565~~
* [x] ~~default https://github.com/log2timeline/plaso/pull/2573~~
* [x] ~~lfu https://github.com/log2timeline/plaso/pull/2570~~
* [x] ~~mountpoints https://github.com/log2timeline/plaso/pull/2572~~
* [x] ~~mrulistex https://github.com/log2timeline/plaso/pull/2576~~
* [x] ~~mrulist https://github.com/log2timeline/plaso/pull/2575~~
* [x] ~~msie_zones https://github.com/log2timeline/plaso/pull/2579~~
* [x] ~~network_drives https://github.com/log2timeline/plaso/pull/2581~~
* [x] ~~officemru https://github.com/log2timeline/plaso/pull/2627~~
* [x] ~~outlook search mru https://github.com/log2timeline/plaso/pull/2582~~
* [x] ~~programscache https://github.com/log2timeline/plaso/pull/2615~~
* [x] ~~run/runonce https://github.com/log2timeline/plaso/pull/2583~~
* [x] ~~sam_users https://github.com/log2timeline/plaso/pull/2584~~
* [x] ~~services https://github.com/log2timeline/plaso/pull/2626~~
* [x] ~~task_scheduler https://github.com/log2timeline/plaso/pull/2585~~
* [x] ~~terminal_server https://github.com/log2timeline/plaso/pull/2614~~
* [x] ~~timezone https://github.com/log2timeline/plaso/pull/2587~~
* [x] ~~typedurls https://github.com/log2timeline/plaso/pull/2591~~
* [x] ~~usb https://github.com/log2timeline/plaso/pull/2593~~
* [x] ~~usbstor https://github.com/log2timeline/plaso/pull/2595~~
* [x] ~~userassist https://github.com/log2timeline/plaso/pull/2600~~
* [x] ~~windows_version https://github.com/log2timeline/plaso/pull/2601~~
* [x] ~~winlogon https://github.com/log2timeline/plaso/pull/2612~~
[...]

PR #2575, opened 2019-05-29
Removed regvalue from MRUList Windows Registry plugin #2566

joachimmetz:
Removed regvalue from MRUList Windows Registry plugin #2566


joachimmetz:
Fixed merge conflict, waiting for completion of CI tests before merge

Issue #2473, opened 2019-04-25
Remove Python 2 tests


Onager:

joachimmetz:
Need to have a solution for MacOS first https://github.com/log2timeline/plaso/issues/2484

Onager:
Completed in https://github.com/log2timeline/plaso/pull/2797

PR #2578, opened 2019-05-29
Add REQUIRED_COLUMNS checking to SQLite parser plugins

### LINKED ISSUE #2133: Add "partial schema match" functionality to sqlite parser.
Per https://github.com/log2timeline/plaso/pull/2119/files#r214299793

obsidianforensics:
## Description:

From #2119:

> In my opinion, the "required tables" check is too coarse, and the "exact schema match" is too fine. It would be great to have an in-between option, something like "required columns", which would function like "required tables" but one step lower. If the columns specified in "required columns" are a subset of what's present, then run the plugin. I think this approach would help with a bunch of plugins, as the schemas of many databases are constantly evolving.

This PR adds REQUIRED_COLUMNS, which is a dict with table names as keys and sets of column names as values. Similar to how REQUIRED_TABLES works, the sqlite parser will check if the specified tables have the required columns as a subset before selecting the plugin as appropriate to run. 

This design allows plugin authors to check for only tables/columns in a database that are actually used in the plugin. This avoids past problems, where a unused column that isn't parsed is added/removed/changed, but still breaks the plugin since the database schema changed. 

**Related issue (if applicable):** fixes #2133

## Notes:
All contributions to Plaso undergo [code review](https://github.com/log2timeline/l2tdocs/blob/master/process/Code%20review%20process.asciidoc). This makes sure that the code has appropriate test coverage and conforms to the Plaso [style guide](https://plaso.readthedocs.io/en/latest/sources/developer/Style-guide.html).

## Checklist:
  - [ ] Reviewer assigned


obsidianforensics:
The AppVeyor tests using Python2.7 are failing because SQLite wasn't compiled with FTS3/4 support (https://www.sqlite.org/fts3.html#compiling_and_enabling_fts3_and_fts4). I'm not sure if this is something we want to fix, or just ignore as Python2.7 is going away.

joachimmetz:
> I'm not sure if this is something we want to fix, or just ignore as Python2.7 is going away.

Marking as blocked since currently python 2 is still supported https://github.com/log2timeline/plaso/issues/2473

plaso/parsers/sqlite.py
```diff
@@ -273,6 +274,13 @@ def Open(self, file_object, wal_file_object=None):
           table_name: ' '.join(query.split())
           for table_name, query in sql_results}
 
+      for table in self.schema.keys():
+        self.table_columns.setdefault(table, [])
+        pragma_results = cursor.execute('PRAGMA table_info({})'.format(table))
```
joachimmetz: '{}' => add index and type e.g. '{0:s}'

plaso/parsers/sqlite.py
```diff
@@ -273,6 +274,13 @@ def Open(self, file_object, wal_file_object=None):
...
+
+        for column in pragma_results:
```
joachimmetz: maybe: 

```
for _, column_name in pragma_results:
    self.table_columns[table].append(column_name)
```

plaso/parsers/sqlite.py
```diff
@@ -143,6 +143,7 @@ def __init__(self, filename, temporary_directory=None):
     self._temp_wal_file_path = ''
 
     self.schema = {}
+    self.table_columns = {}
```
joachimmetz: table_columns => columns_per_table

plaso/parsers/sqlite.py
```diff
@@ -273,6 +274,13 @@ def Open(self, file_object, wal_file_object=None):
           table_name: ' '.join(query.split())
           for table_name, query in sql_results}
 
+      for table in self.schema.keys():
```
joachimmetz: Preference nit:
table => table_name

since it is not a table object or equiv.

plaso/parsers/sqlite.py
```diff
@@ -418,6 +426,20 @@ def ParseFileEntry(self, parser_mediator, file_entry):
         if not plugin.REQUIRED_TABLES.issubset(table_names):
           continue
 
+        columns_match = True
+        for required_table, required_columns in plugin.REQUIRED_COLUMNS.items():
+          if required_table not in table_names:
+            columns_match = False
+            continue
```
joachimmetz: why not do a break instead of continue?

plaso/parsers/sqlite.py
```diff
@@ -418,6 +426,20 @@ def ParseFileEntry(self, parser_mediator, file_entry):
...
+
+          if not frozenset(required_columns).issubset(
+              database.table_columns.get(required_table)):
+            columns_match = False
+            continue
```
joachimmetz: same here, why loop over remaining tables and columns if you know the outcome is a fail.

plaso/parsers/sqlite.py
```diff
@@ -418,6 +426,20 @@ def ParseFileEntry(self, parser_mediator, file_entry):
         if not plugin.REQUIRED_TABLES.issubset(table_names):
           continue
 
+        columns_match = True
```
joachimmetz: Maybe move this into a `_CheckRequiredTablesAndColumns` method?

obsidianforensics: I've moved this to a method. I had kept the REQUIRED_TABLES part untouched, as it was, but now think that might not be best. What do you think about combining REQUIRED_TABLES and REQUIRED_COLUMNS into just one check? I think it makes sense and would be less duplicative for plugin authors and slightly more efficient when running. I'll need to update the REQUIRED_x definitions in each plugin though, so I'd like to get sign-off first.

joachimmetz: > What do you think about combining REQUIRED_TABLES and REQUIRED_COLUMNS into just one check?

SG. Though I'm thinking would there be plugins that only would have REQUIRED_TABLES?

obsidianforensics: Yes, there are some cases where checking for the presence of a table is needed to help identify an otherwise generically-named database, even if we don't use any data from that table. In those cases, the table name is given as required, and the required columns within it is an empty set (basically, we don't care about the columns).

Here is an example where specifying a table name is helpful, even though it isn't used in a query (`android_metadata`), because otherwise just checking for a table named `cookies` is too weak and error-prone:
```
  REQUIRED_STRUCTURE = {
      'android_metadata': frozenset([]),
      'cookies': frozenset([
          '_id', 'name', 'value', 'domain', 'expires', 'path', 'secure'])}
```

I've went and updated all the SQLite plugins to use REQUIRED_STRUCTURE. In most cases, I've added a bit stronger checks to the requirements, which should help avoid running the wrong plugin against a file, but still let the plugin run if minor things about a database change.

joachimmetz:
@obsidianforensics let me know if you need help with rebasing this PR

PR #2580, opened 2019-05-30
Clean up of NetworkList Windows Registry plugin #2566

joachimmetz:
Clean up of NetworkList Windows Registry plugin #2566

### DIFF
--- a/plaso/parsers/winreg_plugins/networks.py
+++ b/plaso/parsers/winreg_plugins/networks.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""This file contains the NetworkList registry plugin."""
+"""This file contains the NetworkList Registry plugin."""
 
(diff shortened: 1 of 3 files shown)


plaso/parsers/winreg_plugins/networks.py
```diff
@@ -25,11 +25,11 @@ class WindowsRegistryNetworkEventData(events.EventData):
     ssid (str): SSID of the connection.
   """
 
-  DATA_TYPE = 'windows:registry:network'
+  DATA_TYPE = 'windows:registry:networklist'
```
Onager: Let's not change the data type until we have a consistent definition of what these needs to look like. This can break downstream analysis processes, and doesn't provide anything for Plaso.

joachimmetz: ack

PR #2593, opened 2019-05-31
Removed regvalue from USB device Windows Registry plugin #2566

joachimmetz:
Removed regvalue from USB device Windows Registry plugin #2566


Onager:
@joachimmetz Please fix conflicts before merging

PR #2629, opened 2019-06-09
Changed output modules to use GetUnformattedAttributes

joachimmetz:
Changed output modules to use GetUnformattedAttributes

### DIFF
--- a/plaso/output/rawpy.py
+++ b/plaso/output/rawpy.py
@@ -62,15 +62,16 @@ class NativePythonFormatterHelper(object):
             'value').format(attribute_name, event_data.data_type))
         attribute_value = ''
 
-      if attribute_name not in definitions.RESERVED_VARIABLE_NAMES:
-        attribute_string = '  {{{0!s}}} {1!s}'.format(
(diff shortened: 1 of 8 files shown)


plaso/output/rawpy.py
```diff
@@ -62,15 +62,16 @@ def GetFormattedEvent(cls, event, event_data, event_tag):
             'value').format(attribute_name, event_data.data_type))
         attribute_value = ''
 
-      if attribute_name not in definitions.RESERVED_VARIABLE_NAMES:
-        attribute_string = '  {{{0!s}}} {1!s}'.format(
-            attribute_name, attribute_value)
-        additional_attributes.append(attribute_string)
+      if attribute_name == 'pathspec':
```
rgayon: It looks to me that the logic changes if you do this. Is this WAI?

joachimmetz: Not sure what you mean. It was `elif attribute_name != 'pathspec':` to exclude pathspec from the reserved_attributes results, but I've changed this to check for pathspec and move to the next attribute.

Issue #2652, opened 2019-06-17
replace xml.etree by defusedxml


joachimmetz:
replace xml.etree by [defusedxml](https://pypi.org/project/defusedxml/)

* [x] update l2tdevtools
  * https://github.com/log2timeline/l2tdevtools/pull/734
* [x] update l2tbinaries, GIFT PPA and GIFT COPR
* [x] change plaso parsers to use defusedxml
* [x] change plaso kml output module to use defusedxml
  * no need, this code is not parsing XML but generating it

PR #2653, opened 2019-06-17
Replaced xml.etree by defusedxml #2652

joachimmetz:
Replaced xml.etree by defusedxml #2652


rgayon:
Though I don't why tests break on mac & windows =)

joachimmetz:
> Though I don't why tests break on mac & windows =)

tests are failing because defusedxml was not yet uploaded to l2tbinaries

PR #2693, opened 2019-07-05
Changed descriptions relating to parser filter expressions

Onager:


joachimmetz:
Changed description of PR seeing not only docstrings are changed.

plaso/cli/helpers/parsers.py
```diff
@@ -26,21 +26,20 @@ def AddArguments(cls, argument_group):
...
-            'or a parser list. Each entry can be prepended with an '
-            'exclamation mark to negate the selection (exclude it). The list '
-            'match is an exact match while an individual parser matching is '
-            'a case insensitive substring match, with support for glob '
-            'patterns. Examples would be: "reg" that matches the substring '
-            '"reg" in all parser names or the glob pattern "sky[pd]" that '
-            'would match all parsers that have the string "skyp" or "skyd" '
-            'in its name. All matching is case insensitive. Use "--parsers '
-            'list" or "--info" to list the available parsers.'))
+        default='', metavar='PARSER_FILTER_EXPRESSION', help=(
```
joachimmetz: Mabye: ', or list available parsers' => ', or show the available parsers and plugins.'

joachimmetz: > Preset matching is exact, while parser and plugin matching is case insensitive

Don't you want to change the preset filtering to be case insensitive? or parser / plugin name filtering to be case sensitive instead?

"list" here does not refer to the Python type but list used in general, since this is help description. Hence I opt:
```
Define which parser or plugin to use or show the available parsers and plugins. The parser filter filter expression is a comma separated list of parser and plugin names, where each element is the name of a parser, plugin or preset. An element can be excluded by pre-pending it with an exclamation mark for example "!bash_history". ...
```

plaso/engine/extractors.py
```diff
@@ -39,13 +39,9 @@ def __init__(self, parser_filter_expression=None):
...
 
-          The parser filter expression is a comma separated value string that
-          denotes a list of parser names to include and/or exclude. Each entry
-          can have the value of:
-
-          * A name of a single parser (case insensitive), such as msiecf.
-          * A glob name for a single parser, such as '*msie*' (case
-            insensitive).
+          A parser filter expression is a comma separated value string that
+          denotes which parsers should be used. See filters/parser_filter.py
```
joachimmetz: parsers and plugins

plaso/engine/worker.py
```diff
@@ -81,15 +81,11 @@ def __init__(self, parser_filter_expression=None):
...
-          * An exact match of a list of parsers, or a preset (see
-            data/presets.yaml for the list of predefined presets).
-          * A name of a single parser (case insensitive), such as msiecf.
-          * A glob name for a single parser, such as '*msie*' (case
-            insensitive).
+          A parser filter expression is a comma separated value string that
+          denotes which parsers should be used. See filters/parser_filter.py
+          for details of the expression syntax.
+
+          Note that preset names in this expression will not be expanded.
```
joachimmetz: `Note that preset names in this expression will not be expanded.`

Though correct this note does not cover the core of the issue. Change to something in line of:

This function requires a parser filter expression in which preset names have been expanded.

plaso/filters/parser_filter.py
```diff
@@ -8,14 +8,16 @@ class ParserFilterExpressionHelper(object):
...
-  * A glob name for a single parser, such as '*msie*' (case insensitive).
+  which parsers should be used. Each value can contain either:
+
+  * The name of a preset (case sensitive), which is a predefined list of parsers
+    and/or plugins (see data/presets.yaml for the default presets).
+  * The name of a parser (case insensitive), for example 'msiecf'.
+  * The name of a plugin, prefixed with the parser name and a '/', for example
+    'sqlite/chrome_history'.
+
+  If the value begins with a '!' character the item will be excluded from the
```
joachimmetz: value => element

so that wording is consistent, with that you preferred in your review of the previous changes to this code

plaso/filters/parser_filter.py
```diff
@@ -8,14 +8,16 @@ class ParserFilterExpressionHelper(object):
...
   A parser filter expression is a comma separated value string that denotes
-  a list of parser names to include and/or exclude. Each entry can have
-  the value of:
-
-  * An exact match of a preset, which is a predefined list of parsers
-    and/or plugins (see data/presets.yaml for the list of predefined
-    presets).
-  * A name of a single parser (case insensitive), such as msiecf.
-  * A glob name for a single parser, such as '*msie*' (case insensitive).
+  which parsers should be used. Each value can contain either:
```
joachimmetz: parsers => parsers and plugins

since you were quite amendment in https://github.com/log2timeline/plaso/pull/2666 about this

plaso/filters/parser_filter.py
```diff
@@ -47,15 +49,16 @@ def _JoinExpression(self, excludes, includes):
...
+        * The name of a parser (case insensitive), for example 'msiecf'.
+        * The name of a plugin, prefixed with the parser name and a '/', for
+          example 'sqlite/chrome_history'.
 
-          * An exact match of a list of parsers, or a preset (see
-            data/presets.yaml for the list of predefined presets).
-          * A name of a single parser (case insensitive), such as msiecf.
-          * A glob name for a single parser, such as '*msie*' (case
-            insensitive).
+        If the value begins with a '!' character the item will be excluded from
```
joachimmetz: a '!' character => and exclamation mark ("!")

plaso/filters/parser_filter.py
```diff
@@ -181,33 +188,41 @@ def ExpandPresets(self, presets_manager, expression):
     return self._JoinExpression(excludes, includes)
 
   def SplitExpression(self, expression):
-    """Determines the excluded and included parsers from an expression string.
+    """Determines the excluded and included values from an expression string.
+
+    Note that presets will not be expanded.
```
joachimmetz: > Note that presets will not be expanded.

Note that function will not expand presets and preset names are treated like parser names.

plaso/filters/parser_filter.py
```diff
@@ -181,33 +188,41 @@ def ExpandPresets(self, presets_manager, expression):
...
-            the dictionary contains the names of plugins per parser name,
-            where '*' represents the parser itself and all its plugins.
-            A parser name can also be the name of a preset.
-        includes (dict[str, set[str]]): included parsers and plugins,
-            the dictionary contains the names of plugins per parser name,
-            where '*' represents the parser itself and all its plugins.
-            A parser name can also be the name of a preset.
+        excludes (dict[str, set[str]]): excluded presets, plugins and presets.
+            Dictionary keys are preset and/or parser names, and values are
+            either a set of plugin names to enable for a parser, or a set
```
joachimmetz: to enable ? did you mean to disable?

joachimmetz: `either` so it is not technically correct, since this function allows a `preset_name: set(['*', 'bla'])` as value so maybe:

and values are a set of plugin names to enable for a parser. Note that for a preset the set is expected to be "set([*])" but this restriction is not enforced by this function.

joachimmetz:
Also please update `docs/sources/user/Log2Timeline-Perl-(Legacy).md`

joachimmetz:
Doing a bit of digging on the glob in the parser filter:
* initial functionality was introduced in https://github.com/log2timeline/plaso/commit/ab6425895ead1891533cfdca1d238d4f587456b0
* which was changed a bit in https://github.com/log2timeline/plaso/commit/a1905ac762cc31f63ac5576ab63eadd5fb5ef05e
* and then removed in https://github.com/log2timeline/plaso/commit/22b84875e374a33dcaa1c4a7285616caae134490

plaso/filters/parser_filter.py
Onager: Changed this around some more, but there's a lot of complexity in what and why this method returns.

plaso/cli/helpers/parsers.py
Onager: Also, it turns out all matching in case insensitive.

joachimmetz:
@Onager I assume you want me to take another look?

plaso/cli/helpers/parsers.py
```diff
@@ -26,21 +26,20 @@ def AddArguments(cls, argument_group):
...
+            'Define which presets, parsers and/or plugins to use, or show '
+            'possible values. The expression is a comma separated string '
+            'where each element is a preset, parser or plugin name. Each '
+            'element can be prepended with an exclamation mark to exclude the '
+            'item. Matching is case insensitive. Examples: "linux,'
+            '!bash_history" enables the linux preset, without the '
+            'bash_history parser. "sqlite, !sqlite/chrome_history" enables '
```
joachimmetz: > "sqlite, !sqlite/chrome_history" 

remove space "sqlite,!sqlite/chrome_history"

plaso/cli/helpers/parsers.py
```diff
@@ -26,21 +26,20 @@ def AddArguments(cls, argument_group):
...
+            'all sqlite plugins except for chrome_history". win7,syslog" '
```
joachimmetz: > chrome_history". win7,syslog" '

fix location of double quote

chrome_history. "win7,syslog" '

plaso/cli/helpers/parsers.py
joachimmetz: ack, more clear now, thx for fixing

joachimmetz:
@Onager some minor help string formatting nits remaining

plaso/cli/helpers/parsers.py
Onager: Done

plaso/cli/helpers/parsers.py
Onager: Done

Issue #2691, opened 2019-07-05
Validate parser names are valid 


Onager:
Fix TODO in extraction_tool.py

joachimmetz:
Issue has been addressed, closing.

PR #2694, opened 2019-07-06
Added validation for parser and plugin names #2691

Onager:


plaso/cli/extraction_tool.py
```diff
@@ -121,7 +121,16 @@ def _CreateProcessingConfiguration(self, knowledge_base):
...
 
-    # TODO: validate parser names.
+    _, invalid_parser_elements = (
+        parsers_manager.ParsersManager.CheckFilterExpression(
+            parser_filter_expression))
+
+    if invalid_parser_elements:
+      invalid_parser_names_string = ','.join(invalid_parser_elements)
+      raise errors.BadConfigOption(
+          'Unknown element(s): "{0:s}" in parser filter '
```
joachimmetz: Maybe "Unsupported names of parsers or plugins found in element(s): "{0:s}" of parser filter."

Onager: Done

joachimmetz:
@Onager please address merge conflicts

plaso/parsers/manager.py
```diff
@@ -45,6 +45,58 @@ def CreateSignatureScanner(cls, specification_store):
...
+      return set(cls._parser_classes.keys()), set()
+
+    known_parser_elements = set()
+    unknown_parser_elements = set()
+    for element in parser_filter_expression.split(','):
+      parser_expression = element
+      if element.startswith('!'):
+        parser_expression = element[1:]
+
+      parser_name, _, plugin_name = parser_expression.partition('/')
```
joachimmetz: Why not use SplitExpression as in GetParserObjects, so that the parser filter logic is contained in ParserFilterExpressionHelper?

Onager: I did consider that, but it makes it more difficult to display the element that cause the problem. This method would have to reconstruct the expression element, so the logic would still be here.

Issue #160, opened 2015-04-02
Code clean up


joachimmetz:
- [x] ~~move dependencies to dfvfs/dependencies~~
- [x] ~~[move tests to separate directory](https://codereview.appspot.com/236520044/)~~
- [x] ~~move tools tests~~ - task abandoned causes import issues
- [x] ~~fix names and email addresses in commit history~~
- [x] ~~[Removed the default is from docstrings where not needed](https://codereview.appspot.com/284890043/)~~
- [x] ~~add tests for bencode plugins~~
- [x] ~~[Moved test cases to single base test case](https://codereview.appspot.com/300010043/)~~
- [x] ~~[remove utils.GetUnicodeString](https://codereview.appspot.com/303220043/)~~
- [x] ~~remove errors.NotAText only used by opera parser~~
- [x] ~~move version checks into dependencies so that the version info is in one place~~
  - ~~libyal, https://codereview.appspot.com/240360043~~
  - ~~sqlite~~
- [x] ~~Clean up parsers~~
  - ~~firefox_cache~~
- [x] ~~replace Utf16StreamCopyToString by in parser utf16 conversion to localize parser errors~~
- [x] ~~Implement a ToString functions that replace `__str__` and `__unicode__`~~
  - ~~`__str__` is only used in filters moving this to https://github.com/log2timeline/plaso/issues/324~~
- [x] ~~Make docstrings consistent~~
- [x] ~~make client protected https://codereview.appspot.com/315530043/diff/20001/plaso/output/elastic.py~~
- [x] ~~add pass to empty classes~~ - pylint now enforces to not do this
  * ~~https://codereview.appspot.com/326740043/diff/60001/tests/cli/tool_options.py?context=10&column_width=80~~
- [x] ~~move `def _GetTestFilePath(self, path_segments):` to main test_lib~~
  - ~~largely in main test lib now~~
- [x] ~~Replace e.g. by "for example" or "such as"~~
  - ~~update [style guide](https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md#docstrings)~~
  - ~~update code: https://github.com/log2timeline/plaso/pull/2700~~
- [x] ~~remove LinearOutputModule - https://github.com/log2timeline/plaso/pull/3369~~
- [x] ~~Fix import order, add comment why pytz is not following the style guide - https://github.com/log2timeline/plaso/pull/3373~~
[...]

PR #2700, opened 2019-07-08
Replaced e.g. by 'for example' or 'such as' #160

joachimmetz:
Replaced e.g. by 'for example' or 'such as' #160


plaso/parsers/selinux.py
```diff
@@ -19,8 +19,8 @@
 
 * http://selinuxproject.org/page/NB_AL
 * http://blog.commandlinekungfu.com/2010/08/episode-106-epoch-fail.html
-* http://www.redhat.com/promo/summit/2010/presentations/
-taste_of_training/Summit_2010_SELinux.pdf
+* http://www.redhat.com/promo/summit/2010/presentations/taste_of_training/
```
rgayon: http://google.github.io/styleguide/pyguide.html says URL shouldn't be wrapped. Do we have a different styleguide here?

joachimmetz: pylint was complaining about this, let me have a look

Issue #2706, opened 2019-07-11
Elasticsearch errors when storage file contains bytes


Onager:
**Description of problem:**
Timesketch/elasticsearch export fails when there's a byte string in the storage file

**Command line and arguments:**

Using the elasticsearch output modules

**Source data:**

N/A

**Plaso version:**

20190708

**Operating system Plaso is running on:**

N/A

**Installation method:**

Timesketch/docker

**Debug output/tracebacks:**

```
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/bin/psort.py", line 83, in <module>
    if not Main():
  File "/usr/bin/psort.py", line 69, in Main
    tool.ProcessStorage()
  File "/usr/lib/python3/dist-packages/plaso/cli/psort_tool.py", line 563, in ProcessStorage
    time_slice=self._time_slice, use_time_slicer=self._use_time_slicer)
  File "/usr/lib/python3/dist-packages/plaso/multi_processing/psort.py", line 1032, in ExportEvents
    use_time_slicer=use_time_slicer)
  File "/usr/lib/python3/dist-packages/plaso/multi_processing/psort.py", line 549, in _ExportEvents
    deduplicate_events=deduplicate_events)
  File "/usr/lib/python3/dist-packages/plaso/multi_processing/psort.py", line 451, in _ExportEvent
    deduplicate_events=deduplicate_events)
  File "/usr/lib/python3/dist-packages/plaso/multi_processing/psort.py", line 598, in _FlushExportBuffer
    output_module.WriteEventMACBGroup(macb_group)
  File "/usr/lib/python3/dist-packages/plaso/output/interface.py", line 125, in WriteEventMACBGroup
    self.WriteEvent(event, event_data, event_tag)
  File "/usr/lib/python3/dist-packages/plaso/output/interface.py", line 78, in WriteEvent
    self.WriteEventBody(event, event_data, event_tag)
  File "/usr/lib/python3/dist-packages/plaso/output/shared_elastic.py", line 342, in WriteEventBody
    self._InsertEvent(event, event_data, event_tag)
  File "/usr/lib/python3/dist-packages/plaso/output/shared_elastic.py", line 227, in _InsertEvent
    self._FlushEvents()
  File "/usr/lib/python3/dist-packages/plaso/output/shared_elastic.py", line 114, in _FlushEvents
    index=self._index_name, request_timeout=self._DEFAULT_REQUEST_TIMEOUT)
[...]

PR #2707, opened 2019-07-11
Changes to output modules for events with bytes attributes values #2706

joachimmetz:
Changes to output modules for events with bytes attributes values #2706


joachimmetz:
Equivalent changes appear to have been merged, closing PR.

PR #2708, opened 2019-07-11
Changed output modules to convert bytes fields to strings #2706

Onager:


plaso/output/l2t_csv.py
```diff
@@ -136,14 +136,14 @@ def _GetOutputValues(self, event, event_data, event_tag):
...
-              'attribute: {0:s} of data type: {1:s} contains an empty bytes '
-              'value').format(attribute_name, event_data.data_type))
-          attribute_value = ''
+        # Some parsers have written bytes values to storage.
+        if isinstance(attribute_value, py2to3.BYTES_TYPE):
+          logger.warning(
+              'Found bytes value "{0!s}" for attribute "{1:s}" for data type: '
+              '{2!s}. Value will be converted to UTF-8'.format(
+                  attribute_value, attribute_name, event_data.data_type))
+          attribute_value = codecs.decode(
```
joachimmetz: why use codecs and not attribute_value.decode('utf-8', 'replace')

plaso/output/l2t_csv.py
```diff
@@ -136,14 +136,14 @@ def _GetOutputValues(self, event, event_data, event_tag):
...
-        if (isinstance(attribute_value, py2to3.BYTES_TYPE) and
-            attribute_value == b''):
-          logging.debug((
-              'attribute: {0:s} of data type: {1:s} contains an empty bytes '
-              'value').format(attribute_name, event_data.data_type))
-          attribute_value = ''
+        # Some parsers have written bytes values to storage.
+        if isinstance(attribute_value, py2to3.BYTES_TYPE):
+          logger.warning(
+              'Found bytes value "{0!s}" for attribute "{1:s}" for data type: '
```
joachimmetz: I opt **not** to print the binary data, since that can lead to encoding errors.

tests/output/shared_elastic.py
```diff
@@ -39,7 +39,7 @@ class SharedElasticsearchOutputModuleTest(test_lib.OutputModuleTestCase):
       {'data_type': 'syslog:line',
        'display_name': 'log/syslog.1',
        'filename': 'log/syslog.1',
-       'hostname': 'ubuntu',
+       'hostname': b'ubuntu',
```
joachimmetz: please add an separate attribute "binary_data" and add a comment what it is used for

Issue #2711, opened 2019-07-12
Uncaught exception when trying to process old storage file


Onager:
**Description of problem:**

Psort crashes when analysis plugins are specified, and a storage file with an older storage format is provided.

**Command line and arguments:**

`PYTHONPATH=. ./tools/psort.py --analysis sessionize -w /tmp/out.csv ~/Downloads/codelab.plaso `

**Source data:**

Test storage file with an older format version than is supported for writing.

**Plaso version:**

20190708

**Operating system Plaso is running on:**

N/A

**Installation method:**
Developer installer

**Debug output/tracebacks:**

```
2019-07-12 11:38:47,872 [INFO] (MainProcess) PID:36549 <data_location> Determined data location: /home/onager/code/plaso/data
Traceback (most recent call last):
  File "./tools/psort.py", line 85, in <module>
    if not Main():
  File "./tools/psort.py", line 67, in Main
    tool.ProcessStorage()
  File "/home/onager/code/plaso/plaso/cli/psort_tool.py", line 544, in ProcessStorage
    worker_memory_limit=self._worker_memory_limit)
  File "/home/onager/code/plaso/plaso/multi_processing/psort.py", line 893, in AnalyzeEvents
    storage_writer.StartTaskStorage()
AttributeError: 'NoneType' object has no attribute 'StartTaskStorage'
```

PR #2712, opened 2019-07-12
Changes to psort to prevent storage exception #2711

Onager:


joachimmetz:
qq. in your description you say "Changes to psort to catch storage exception" but you're checking the storage write is None?

plaso/cli/psort_tool.py
```diff
@@ -531,6 +531,11 @@ def ProcessStorage(self):
       storage_writer = (
           storage_factory.StorageFactory.CreateStorageWriterForFile(
               session, self._storage_file_path))
+      if not storage_writer:
+        logger.error(
```
joachimmetz: Why not raise BadConfigOption instead?

Onager: Because this isn't what's done for other similar cases, see line 515.

We can other instances of this to raise in a follow up PR if you'd prefer that.

joachimmetz:
@Onager you did not answer:

qq. in your description you say "Changes to psort to catch storage exception" but you're checking the storage write is None?

Also per #2711 isn't AttributeError raised?

Onager:
Updated PR description.

The exception is triggered when the `None` return from `CreateStorageWriterForFile` isn't checked.

plaso/cli/psort_tool.py
joachimmetz: ack, this is fine for now

joachimmetz:
Ack, new description is more clear.

Issue #2714, opened 2019-07-13
Change pe parser not to load full file data into memory


joachimmetz:
Currently the pe parser loads the full file data into memory, due to the PE class interface. Discussed this with the author of pefile and he indicates that wrapping the file-like object into a data slice could do the trick without having to change pefile.

* [x] ~~add data slice helper to dfvfs https://github.com/log2timeline/dfvfs/pull/403~~
* [x] ~~add len function to data slice helper https://github.com/log2timeline/dfvfs/pull/405~~
* [x] ~~create a new dfvfs release~~
* [x] change pe parser to use data slice https://github.com/log2timeline/plaso/pull/2715

PR #2715, opened 2019-07-13
Changed PE parser to use dfVFS data slice #2714

joachimmetz:
Changed PE parser to use dfVFS data slice #2714


plaso/parsers/pe.py
```diff
@@ -44,6 +45,12 @@ class PEParser(interface.FileObjectParser):
   NAME = 'pe'
   DESCRIPTION = 'Parser for Portable Executable (PE) files.'
 
+  _PE_DIRECTORIES = [
```
Onager: Why add the extra indirection here? This value is never reused.

joachimmetz: 1. Since these definitions are static (do not change), so that they only need be created once
2. this makes the call to parse_data_directories, more compact

PR #2737, opened 2019-08-08
Changes to add Travis stages

Onager:
Just testing for now, no review required


joachimmetz:
Since we're no longer using Travis CI I'll close this

PR #2740, opened 2019-08-20
handle error ElementTree.ParseError in winevtx parser

jacksonfylle:
This commit will catch the following exception in the winevtx parser : 

> ParseError(ExpatError('not well-formed (invalid token): line 27, column 32'))


plaso/parsers/winevtx.py
```diff
@@ -277,6 +277,10 @@ def _ParseRecords(self, parser_mediator, evtx_file):
         parser_mediator.ProduceExtractionWarning(
             'unable to parse event record: {0:d} with error: {1!s}'.format(
                 record_index, exception))
+      except ElementTree.ParseError as exception:
```
joachimmetz: Instead of adding another except, please change the except at line 276 to:
except (IOError, ElementTree.ParseError) as exception:

PR #2762, opened 2019-10-30
tag_linux.txt

pyllyukko:
## Description:

Wanted to open this up for discussion. So I made (initial) `tag_linux.txt` as it was completely missing from Plaso. I threw together some tags that I would see myself needing, but this is not yet complete and I want to improve this before possible merging.

Notes:

* `data_type is 'syslog:line'` shouldn't be used as it could also be `systemd:journal`. In these cases I think `reporter` is better.
* Same information can be available from different components and I think we should tag all of them and not rely on single component logging some event (e.g. login event that can be logged via PAM, audit or systemd-login)
* Should we separate login vs. authentication events?
    * Or logins & sessions?
* I tried this with Plaso's `test_data/` and also some data of my own
* What else needs to be tagged? Some ideas:
    * Kernel module loads/unloads
    * User account modifications
    * Application installs in different distros
    * Mounts/unmounts
    * Ptrace events
    * Syslog resets or HUPs
    * OOM kills, kernel panics & core dumps
    * `at`
    * AppArmor stuff

## Notes:
review](https://github.com/log2timeline/l2tdocs/blob/master/process/Code%20review%20process.asciidoc). This makes sure

## Checklist:
  - [ ] Reviewer assigned

### DIFF
--- a/data/tag_linux.txt
+++ b/data/tag_linux.txt
@@ -14,6 +14,7 @@ login_failed
   reporter is 'sshd' AND body contains 'uthentication failure'
   data_type is 'selinux:line' AND audit_type is 'USER_LOGIN' AND body contains 'res=failed'
   data_type is 'selinux:line' AND audit_type is 'ANOM_LOGIN_FAILURES'
+  reporter is 'xscreensaver' AND body contains 'FAILED LOGIN'
 
 logout
   reporter is 'login' AND body contains 'session closed'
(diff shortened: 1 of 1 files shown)


pyllyukko:
How do I write event filter where some field is null (or `""`)? I tried to do the following:

```
data_type is 'linux:utmp:event' AND type == 8 AND terminal != '' AND pid != 0 AND username is ''
```

But that didn't work (no tags produced). How should this be done?

Onager:
Thanks for this! I think the current rules you have here are a good start, and I agree with the approach of tagging events from multiple components that indicate the same event. We can add some more rules as we test these a bit more. 
I don't think we have a great way to represent empty fields at present, I think you'd need to use a regex for it.

Thanks for the contribution!

PR #2768, opened 2019-11-06
Removed remaining references to guppy

joachimmetz:
Removed remaining references to guppy


joachimmetz:
@Onager PTAL, I accidentally created a feature branch on upstream that is why appveyor test is cancelled

joachimmetz:
@Onager PTAL

Issue #2773, opened 2019-11-07
Unable to expand presets in parser filter expression with error: Parser filters: filestat defined to be both included and excluded


joachimmetz:
`log2timeline.py --parsers 'winxp,!filestat'`

raises

`Unable to expand presets in parser filter expression with error: Parser filters: filestat defined to be both included and excluded`

joachimmetz:
Changes merged, closing issue.

PR #2774, opened 2019-11-07
Support concurrent included and excluded parser filters #2773

joachimmetz:
Support concurrent included and excluded parser filters #2773


plaso/filters/parser_filter.py
```diff
@@ -43,6 +43,8 @@ def _GetParserAndPluginsList(self, parsers_and_plugins):
   def _JoinExpression(self, excludes, includes):
     """Creates an expression string of the excluded and included parsers.
 
+    Note that exclusion takes precedent over inclusion.
```
Onager: Precedence, not precedent

joachimmetz: ack thx

PR #2800, opened 2020-01-08
Documentation clean up

joachimmetz:
Documentation clean up


joachimmetz:
@tomchop per conversation today, let me know if you have other suggestions. This is a first iteration.

joachimmetz:
Note to self, spoke with @Onager let change the deleted docs to place holder docs that link to the current documentation. In case the deleted docs are referenced elsewhere

PR #2802, opened 2020-01-10
Added instruction to create a PyInstaller build

joachimmetz:
Added instruction to create a PyInstaller build


joachimmetz:
@Onager any place in the dev documentation where you think we should reference this?

Onager:
I think this actually makes more sense in in the maintainers guide for l2tdevtools, or in l2tdocs. The developer docs are about developing for plaso in general, and IIRC, that's where all the other packaging information is.

joachimmetz:
we had it in l2tdocs before. I'll add it to plaso for now, we can always move it at a later date

PR #2806, opened 2020-01-18
Removed Python 2 from dpkg files

joachimmetz:
Removed Python 2 from dpkg files


joachimmetz:
@rgayon PTAL the tests are failing due to change to cryptography.io (pending fixes https://github.com/log2timeline/plaso/pull/2807)

PR #2808, opened 2020-01-18
Added a plaso version out of date check

joachimmetz:
Added a plaso version out of date check

### DIFF
--- a/plaso/cli/tools.py
+++ b/plaso/cli/tools.py
@@ -257,6 +257,21 @@ class CLITool(object):
             'default to UTC. Use "-z list" to see a list of available '
             'timezones.'))
 
+  def CheckOutDated(self):
+    """Checks if the version of plaso is outdated and warns the user."""
+    version_date_time = datetime.datetime(
+        int(plaso.__version__[0:4], 10),
+        int(plaso.__version__[4:6], 10),
+        int(plaso.__version__[6:8], 10))
+    date_time_delta = datetime.datetime.utcnow() - version_date_time
+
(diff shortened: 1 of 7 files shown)


joachimmetz:
@rgayon I've made some changes PTAL

plaso/cli/tools.py
```diff
@@ -257,6 +257,22 @@ def AddTimeZoneOption(self, argument_group):
...
+        int(plaso.__version__[6:8], 10))
+    date_time_delta = datetime.datetime.utcnow() - version_date_time
+
+    if date_time_delta.days > 180:
+      print('WARNING the version of plaso you are using is more than 6 months '
+            'old.')
+      print('We strongly recommend to update it. Hit enter to continue or '
+            'Ctrl+C to break.')
+      print('')
+      self._input_reader.Read()
```
rgayon: I don't have strong opinions for or against the .Read() here, but this might be surprising to people trying to run plaso programatically, in environment where it's difficult for people to update regularly.
So while this LGTM,  maybe this should be very clear, in a the README, that we do this check now. Maybe even add a --yes-run-old-code flag?

joachimmetz: good point we might be "punishing" the users for the actions of maintainers

Onager: As discussed, I think we should just print the message on the console, as well as an Error log message, vs blocking.

Asking for input will cause some pain, especially in docker etc.

joachimmetz:
@Onager PR updated PTAL

PR #2817, opened 2020-01-23
Fix executable path in Winprefetch.py

Resolk:
Fix executable path in Winprefetch.py

First occurence of executable path in the files reference list is more likely to be correct.

## Description:
First occurence of executable path in the files reference list is more likely to be correct.

E.g., 64-bit version of Internet Explorer 11 can start x86 version of itself which results in the following situation:

The .pf file of 64-bit version of IE has two records ending with 'iexplore.exe': 
- path to 64-bit version (with smaller index in the files reference list);
- path to x86 version (with higher index, the one which is now wrongly shown by log2timeline as executable path).

## Checklist:
  - [x] Reviewer assigned

### DIFF
--- a/plaso/parsers/winprefetch.py
+++ b/plaso/parsers/winprefetch.py
@@ -122,6 +122,7 @@ class WinPrefetchParser(interface.FileObjectParser):
         if (filename.startswith(volume_device_path) and
             filename.endswith(executable_filename)):
           _, _, path = filename.partition(volume_device_path)
+          break
 
     mapped_files = []
     for entry_index, file_metrics in enumerate(scca_file.file_metrics_entries):


joachimmetz:
This change will make the filename look up a bit faster. However based on the title "Fix executable path in Winprefetch.py" I'm wondering if you ran into an issue with the file path? 

@Resolk could you provide a bit of context. Otherwise the PR looks good to go.

Resolk:
> However based on the title "Fix executable path in Winprefetch.py" I'm wondering if you ran into an issue with the file path?

Yes, I do have some issue with that. 

I have the .pf file of 64-bit Internet Explorer executable which has 2 filename records ending with 'iexplore.exe' (since the 64-bit version of IE starts the x86 version):
- path to 64-bit version (with smaller index in the files reference list);
- path to x86 version (with higher index, the one which **is now wrongly shown by log2timeline as executable path**).

joachimmetz:
Thx for that context. Reflecting on this, if I'm not mistaken when the x86 version would be earlier in the list (with the change) it would should it wrong? so we might need to consider a longest (most specific) match approach instead.

Resolk:
Yes, when the x86 version would be earlier in the list (with the change) it would show wrong. 
But with my experience I have not had sutiations like this.
 
As well https://github.com/libyal/libscca/blob/master/documentation/Windows%20Prefetch%20File%20(PF)%20format.asciidoc has no information about any specific approach searching for executable path. Since the OS allready has its own way to search for the .pf file when starting an executable (by different hash in different versions of OS) - there is no need for OS to save the exact executable path inside the .pf.

All I trying to say is that the filename with the smaller index has more chances to be correct as executable path.

Resolk:
Btw this is how WinPrefetchView and other forensic tools work around this sutiation. They take filename from the list with the smallest index.

joachimmetz:
> But with my experience I have not had sutiations like this.

Understood, I'm trying to determine a structural fix here.

> Btw this is how WinPrefetchView and other forensic tools work around this sutiation. They take filename from the list with smaller index.

I do not wish to blindly copy the behavior of other "forensic" tools, especially if we know that there are short comings in the methodology they apply.

> All I trying to say is that the filename with the smaller index has more chances to be correct as executable path.

Could you explain/elaborate this? If you have research / observations that back up these claims please add them.

> has no information about any specific approach searching for executable path.

Correct the scope of that document is mainly to describe the data format. I assume under the hood the Prefetcher uses the NTFS file reference which is an unique identifier for an MFT entry and even accounts for deletion.

joachimmetz:
So looking into this a bit more, the proper way is presumably to hash the path and compare it with the prefetch hash.

Resolk:
I was thinking of this as a solution, but found it failing in the following situations:
1. For hosting applications, e.g. rundll32.exe, you need the actual command line arguments which are also case-sensitive for prefetch-hash (see: http://www.hexacorn.com/blog/2012/06/13/prefetch-hash-calculator-a-hash-lookup-table-xpvistaw7w2k3w2k8/).
2. For executables having 2 (or more) hardlinks of MFT-record, you may have one hardlink path in the filenames list and prefetch-hash calculated by another (so you will have to look up the MFT-records of possible filenames and calculate hashes of all the hardlinks).

Btw, one should not forget that prefetch-hash even depends on the bitness of the OS, which I assume you can't get in 100% cases just from one .pf file.

joachimmetz:
what do you mean with "bitness" in this context? byte-order or endianess or architecture (32-bit or 64-bit), or something else? The hashing algorithm is also different for different versions of Windows.

> For hosting applications, e.g. rundll32.exe, you need the actual command line arguments which are also case-sensitive for prefetch-has

Agree there are limitations, the main thought was to use it when there is more than a single match. I was playing around with this in https://github.com/log2timeline/plaso/pull/2823

Maybe renaming this field to "full path hint" is a better reflection on its purpose.

Resolk:
Yes, I mean that hashing algorithm is different not only for different versions of Windows, but also for different architecture (e.g. Win7 32-bit and Win7 64-bit). This concerns cases of 'hosting applications' I noted above.

Which is more important - that the solution you propose has 2 limitations. The second is:
> For executables having 2 (or more) hardlinks of MFT-record, you may have one hardlink path in the filenames list and prefetch-hash calculated by another (so you will have to look up the MFT-records of possible filenames and calculate hashes of all the hardlinks).

Renaming LGFM if full path hint contains the first occurence of executable path in the files reference list.

joachimmetz:
> For executables having 2 (or more) hardlinks of MFT-record, you may have one hardlink path in the filenames list and prefetch-hash calculated by another (so you will have to look up the MFT-records of possible filenames and calculate hashes of all the hardlinks).

these will only match with the existing implementation if the hard link has the exact same name as the file.

an alternative approach is to provide both paths as hints, which IMHO is more accurate than defaulting to the first one

Resolk:
I see you're trying to find the most accurate way, which I think is impossible without the definitive description of Prefetch subsystem.

Yes, the existing implementation will only match if the hard link has exact same name as the file. But the implementation you propose has this problem as well. Moreover hints will fail as well.

Resolk:
Btw, example with hardlink is real. 

I had this with IE 11 on Win7 machine:

64-bit version .exe had 2 hardlinks (in Program Files and in WinSxS) and started 32-bit version of .exe (in Program Files (x86)). Prefetch hash was calculated with 64-bit version of .exe in Program Files, but in the file referense list I had 64-bit .exe in WinSxS and 32-bit .exe in Program Files (x86).

joachimmetz:
> Yes, the existing implementation will only match if the hard link has exact same name as the file. But the implementation you propose has this problem as well. 

> Btw, example with hardlink is real.

as I indicated, with the existing logic the path (or path hint) will not be filled if the file name does not match the name in the prefetch file. By returning multiple matches the implementation at least does not assume one match is more appropriate than the other, based on order of occurrence. 

I agree that this is not an exact match, but as we both have indicated we do not have sufficient data to determine this in the parser. Hence IMHO another reason to provide multiple possible results if appropriate. So this can be determined later, with other sources, if relevant.

I've updated https://github.com/log2timeline/plaso/pull/2823 see if it makes sense.

Resolk:
So if we base on the method you propose and forget about 'hosting applications' - the most accurate way to find the executable path is to calculate hashes of each path in the file reference list including its hardlinks from MFT (since we do have the MFT record numbers) or at least do this for each path containing the filename from .pf name.

joachimmetz:
> So if we base on the method you propose and forget about 'hosting applications' - the most accurate way to find the executable path is to calculate hashes of each path in the file reference list including its hardlinks from MFT (since we do have the MFT record numbers) or at least do this for each path containing the filename from .pf name.

No, as you can see in the the **updated** https://github.com/log2timeline/plaso/pull/2823

instead of returning the last match on volume as prefix and file name as suffix, it will now return all matches. Not only the first match or only the last match.

joachimmetz:
Also "path" was renamed to "path hint" to indicate this is an approximation

Resolk:
Thank you, I've checked the #2823.

Anyway if you search for the filename to find the 'path hints' you could check these pathes hashes first (as you've done before). If none of them match you could show them as 'path hints' then, otherwise you've found the exact path (which could be shown as 'path').

joachimmetz:
> Anyway if you search for the filename to find the 'path hints' you could check these pathes hashes first (as you've done before). If none of them match you could show them as 'path hints' then, otherwise you've found the exact path (which could be shown as 'path').

This would be still somewhat of an approximation, since I'm not sure what the collision characteristics of the hashing methods is. And I'm not familiar with any research into this aspect of the prefetch hash.

The approach is also brittle/fragile for Windows 10, where "\DEVICE\HARDDISKVOLUME#" has replaced by "\VOLUME{%IDENTIFIER%}" in the paths, but the prefetch hash appears to be calculated using "\DEVICE\HARDDISKVOLUME#".

Resolk:
> I'm not sure what the collision characteristics of the hashing methods is
That's true, I'm not sure as well.

However, this won't produce any errors in Win10, since the hash just won't match and you will show all pathes (with filename match) as 'path hints'.

joachimmetz:
Since https://github.com/log2timeline/plaso/pull/2823 seems to work for you, I'll close and not merge this PR in favor of https://github.com/log2timeline/plaso/pull/2823. Thx for helping address the underlying issue and proposing a fix.

> However, I assume this won't produce any errors in Win10, since the hash just won't match and you will show all pathes (with filename match) as 'path hints'.

We can add this at a later point if needed. I checked and this path approximation dates back to the initial version of the parser in plaso, so I assume the approach was adopted from log2timeline.pl.

The "file references" information in the prefetch might be another way to determine the full path but more research is needed there.

PR #2822, opened 2020-02-09
Changes to make AMCache parser more error tolerant.

joachimmetz:
Changes to make AMCache parser more error tolerant.


joachimmetz:
rebased @Onager PTAL

plaso/parsers/amcache.py
```diff
@@ -1,331 +1,324 @@
...
+
+    try:
+      if '0000' in file_reference_key.name:
+        # A NTFS file is a combination of MFT entry and sequence number.
+        sequence_number, mft_entry = file_reference_key.name.split('0000')
+        mft_entry = int(mft_entry, 16)
+        sequence_number = int(sequence_number, 16)
+        event_data.file_reference = '{0:d}-{1:d}'.format(
+            mft_entry, sequence_number)
+      else:
```
Onager: What if it's not FAT? What if it's exFAT, or some other file system? I don't see that there's any checking that file_reference_key.name has to be a number, and the event data is a string, so please catch ValueError here and add it directly to the file reference.

plaso/parsers/amcache.py
```diff
@@ -1,331 +1,324 @@
...
+    Args:
+      parser_mediator (ParserMediator): mediates interactions between parsers
+          and other components, such as storage and dfvfs.
+      file_object (dfvfs.FileIO): file-like object.
+    """
+    regf_file = pyregf.file()
+    try:
+      regf_file.open_file_object(file_object)
+    except IOError:
+      # The error is currently ignored -> see TODO above related to the
```
Onager: Please log the failure here at debug at least.

plaso/formatters/interface.py
```diff
@@ -220,16 +220,20 @@ class ConditionalEventFormatter(EventFormatter):
...
+    """Initializes the conditional formatter."""
+    super(ConditionalEventFormatter, self).__init__()
+    self._format_string_pieces_map = []
+    self._format_string_short_pieces_map = []
 
-    A map is build of the string pieces and their corresponding attribute
+  def _CreateFormatStringMaps(self):
+    """Creates the format string maps.
+
+    Maps are build of the string pieces and their corresponding attribute
```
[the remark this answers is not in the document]
joachimmetz: done

plaso/parsers/amcache.py
```diff
@@ -1,331 +1,324 @@
...
-    languagecode (int): language code of file
-    filesize (int): size of file in bytes
-    filedescription (str): description of file
-    linkerts (int): unix timestamp when file was linked
-    lastmodifiedts (int): filetime timestamp of last modified datetime of file
-    createdtd (int): filetime timestamp of created datetime of file
-    programid (str): GUID of entry under Root/Program key file belongs to
+    company_name (str): company name that created product file belongs to.
+    file_description (str): description of file.
+    file_reference (int): file system file reference, for example 9-1 (MFT
```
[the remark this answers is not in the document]
joachimmetz: done

plaso/parsers/amcache.py
joachimmetz: exfat and fat are the same family, other file systems like ReFS are currently unknown and will default to a single number.

ValueError is already caught at line 207 but I'll add TypeError

plaso/parsers/amcache.py
joachimmetz: the reason for not adding a debug print is that amcache.ghve are regf files that are also handled by the Registry parsing in plaso and this error will show up there as well.

joachimmetz:
@Onager PTAL

PR #2825, opened 2020-02-11
Added pyluskde dependency and changes to use pybuild debhelper

joachimmetz:
Added pyluskde dependency and changes to use pybuild debhelper


joachimmetz:
For context log2timeline/l2tdevtools#666

PR #2826, opened 2020-02-20
Added functionality to not re-serialize containers from task stores

Onager:
## Description:
This change preserves the serialized form of containers read from task stores when writing to the session store. The intent is to improve performance by not re-running the expensive serialization when no changes are made to the container by the main storage process.

## Checklist:
  - [x] Reviewer assigned


plaso/storage/fake/writer.py
```diff
@@ -74,11 +74,12 @@ def _RaiseIfNotWritable(self):
     if not self._is_open:
       raise IOError('Unable to write to closed storage writer.')
 
-  def AddAnalysisReport(self, analysis_report):
+  def AddAnalysisReport(self, analysis_report, serialized_data=None):
```
joachimmetz: design question why not add a separate API/methods for retrieving serialized data instead of overloading existing ones?

Onager: This PR is about writing serialized data, not retrieving it so I'm not sure I completely understand what you mean, but some reasons for overloading like this:
* Different storage implementations need to handle different containers in different ways, so having a generic method won't work. For example, the SQLite storage needs to know which table to write the container to.
* Some storage implementations will be able to store JSON-serialized containers, but others might not. This way there's a generic interface for all of these, and each implementation can control whether it wants to use the the JSON seralized form or do its own thing.
* Adding new methods would greatly increase the size of the storage API for what is effectively a 1-line difference in a single method. Line 122 in sqlite_file is the only significant functional difference here.

joachimmetz: > This PR is about writing serialized data, not retrieving it so I'm not sure I completely understand what you mean

Ack, I meant writing serialized data

> Adding new methods would greatly increase the size of the storage API for what is effectively a 1-line difference in a single method. 

Why not expose the underlying method that is needed, instead of passing serialized data? 

Do you have an example of where is this actually used? I assume this is used in the merge reader?

joachimmetz:
Per conversation moving this to next milestone to ensure more testing of storage changes

plaso/storage/fake/writer.py
Onager: It's used on line 292 of the SQLite merge reader (https://github.com/log2timeline/plaso/pull/2826/files#diff-b5b63b70dc1c537e4cd34e91085a5d27R293) , which then flows through to line 122 in sqlite_file https://github.com/log2timeline/plaso/pull/2826/files#diff-2d28634867c3645a616470d2f9bad420R122.  Each storage implementation "AddContainer" method determines whether to pass through the serialized data to the writer, or drop it if the storage implementation needs to modify the container

joachimmetz: ack thx

PR #2827, opened 2020-02-20
Made test paths independent of working directory.

Onager:
## Description:
Currently, tests rely on the working directory being the project root. This PR makes the relevant functions locate test files (test_data directory etc.) without relying on the current value of the working directory.

## Checklist:
  - [x] Reviewer assigned

### DIFF
--- a/tests/test_lib.py
+++ b/tests/test_lib.py
@@ -14,7 +14,7 @@ from dfvfs.lib import definitions as dfvfs_definitions
 from dfvfs.path import factory as path_spec_factory
 from dfvfs.resolver import resolver as path_spec_resolver
 
-PROJECT_PATH = os.path.normpath(
+PROJECT_PATH = os.path.abspath(
     os.path.join(os.path.dirname(__file__), '..'))
 
 def GetTestFilePath(path_segments):
@@ -35,8 +35,8 @@ class BaseTestCase(unittest.TestCase):
   """The base test case."""
 
   _PROJECT_PATH = PROJECT_PATH
-  _DATA_PATH = os.path.join(_PROJECT_PATH, 'data')
-  _TEST_DATA_PATH = os.path.join(_PROJECT_PATH, 'test_data')
+  _DATA_PATH = os.path.abspath(os.path.join(_PROJECT_PATH, 'data'))
+  _TEST_DATA_PATH = os.path.abspath(os.path.join(_PROJECT_PATH, 'test_data'))
 
   # Show full diff results, part of TestCase so does not follow our naming
   # conventions.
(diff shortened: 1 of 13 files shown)


tests/test_lib.py
```diff
@@ -14,6 +14,8 @@
 from dfvfs.path import factory as path_spec_factory
 from dfvfs.resolver import resolver as path_spec_resolver
 
+PROJECT_PATH = os.path.abspath(
```
joachimmetz: The problem is that some build environments require the paths to the test data to be passed at run-time and defining it as such is not going to work. Let's sync separately about this.

joachimmetz:
@Onager let's sync separately about these changes.

tests/test_lib.py
Onager: I think I understand your concern here. I've refactored so all the paths are defined here for easier overriding. Files need to be updated to read the test_data path at one time anyway, and now all this overriding can take place in one location.

joachimmetz: Unfortunately the refactored solution is not going to work. The test data path need to be determined at run-time. As I indicated, I can better explain why this is outside this PR.

Onager:
Re-requesting review per discussion

tests/test_lib.py
```diff
@@ -14,6 +14,20 @@
...
+# compatibility with different build systems.
+ANALYSIS_PATH = os.path.join(PROJECT_PATH, 'plaso', 'analysis')
+ANALYZERS_PATH = os.path.join(PROJECT_PATH, 'plaso', 'analyzers')
+CLI_HELPERS_PATH = os.path.join(PROJECT_PATH, 'plaso', 'cli', 'helpers')
+CONTAINERS_PATH = os.path.join(PROJECT_PATH, 'plaso', 'containers')
+DATA_PATH = os.path.join(PROJECT_PATH, 'data')
+OUTPUT_PATH = os.path.join(PROJECT_PATH, 'plaso', 'output')
+PARSERS_PATH = os.path.join(PROJECT_PATH, 'plaso', 'parsers')
+PREPROCESSORS_PATH = os.path.join(PROJECT_PATH, 'plaso', 'preprocessors')
+TEST_DATA_PATH = os.path.join(PROJECT_PATH, 'test_data')
```
joachimmetz, line 31: `+1` white line

tests/test_lib.py
```diff
@@ -14,6 +14,20 @@
 from dfvfs.path import factory as path_spec_factory
 from dfvfs.resolver import resolver as path_spec_resolver
```
joachimmetz, line 16: `+1` white line

joachimmetz:
A couple of remaining nits. I'll make the changes

tests/test_lib.py
joachimmetz: done

tests/test_lib.py
joachimmetz: done

Issue #2129, opened 2018-09-06
Create style guide for EventData contents


Onager:
Points to cover:
- What level of meaning to aim for in eventdata vs formatter (bool/string)
- Naming conventions

joachimmetz:
Some more points to address:

* storing binary data
* storing data as-is or e.g. int representing a boolean or converting to Python types first

joachimmetz:
More points to address:
- event attributes should not be dictionaries (e.g. regvalue)
  - not supported by event filter

berggren:
+1 on attributes should not be dictionaries, this is also problematic for Elasticsearch for example.

joachimmetz:
@Onager I've taken a first stab at writing something down https://github.com/log2timeline/plaso/pull/2917. You think we can get this completed as part of the May release?

joachimmetz:
Don't see this happening before next release, bumping milestone

joachimmetz:
Marking as blocked, make this part of the migration to dfKinds

joachimmetz:
Changes in https://github.com/log2timeline/l2tdocs/blob/main/process/Style-guide.md#event-data-attributes. Feel free to expand where necessary.

Issue #158, opened 2015-03-30
Extract system-level event from log sources


joachimmetz:
Show case CL: https://codereview.appspot.com/223890043/

joachimmetz:
Also see: https://codereview.appspot.com/311730043/diff/1/plaso/parsers/winevtx.py making a generic approach.

joachimmetz:
https://github.com/log2timeline/plaso/pull/2860

PR #2851, opened 2020-03-19
Parses xml_string of Windows EventLogs Into a New Field

### LINKED ISSUE #442: add a Windows Event Log XML parser
Add a parser for Window Event Log XML output

einfallstoll:

Adds an additional field for Windows EventLogs which contains a dictionary of the parsed XML.

## Description:

The XML in EventLogs contains a lot of additional information. For easier parsing and filtering in Kibana we needed to parse the XML and make it available in a dictionary.

`xml_string`:
```
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
    <Provider Name="Microsoft-Windows-Security-Auditing" Guid="{54849625-5478-4994-A5BA-3E3B0328C30D}"/>
    <EventID>4624</EventID>
    <Version>0</Version>
    <Level>0</Level>
    <Task>12544</Task>
    <Opcode>0</Opcode>
    <Keywords>0x8020000000000000</Keywords>
    <TimeCreated SystemTime="2015-08-08T02:06:17.187500000Z"/>
    <EventRecordID>2</EventRecordID>
    <Correlation/>
    <Execution ProcessID="448" ThreadID="452"/>
    <Channel>Security</Channel>
    <Computer>37L4247E29-32</Computer>
    <Security/>
  </System>
  <EventData>
    <Data Name="SubjectUserSid">S-1-0-0</Data>
    <Data Name="SubjectUserName">-</Data>
    <Data Name="SubjectDomainName">-</Data>
    <Data Name="SubjectLogonId">0x0000000000000000</Data>
    <Data Name="TargetUserSid">S-1-5-18</Data>
    <Data Name="TargetUserName">SYSTEM</Data>
    <Data Name="TargetDomainName">NT AUTHORITY</Data>
    <Data Name="TargetLogonId">0x00000000000003e7</Data>
    <Data Name="LogonType">0</Data>
    <Data Name="LogonProcessName">-</Data>
    <Data Name="AuthenticationPackageName">-</Data>
    <Data Name="WorkstationName">-</Data>
    <Data Name="LogonGuid">{00000000-0000-0000-0000-000000000000}</Data>
    <Data Name="TransmittedServices">-</Data>
    <Data Name="LmPackageName">-</Data>
    <Data Name="KeyLength">0</Data>
    <Data Name="ProcessId">0x0000000000000004</Data>
    <Data Name="ProcessName"/>
    <Data Name="IpAddress">-</Data>
    <Data Name="IpPort">-</Data>
  </EventData>
</Event>
```

Parsed `xml`:
```
{
  "Event": {
    "EventData": {
      "AuthenticationPackageName": "-",
      "IpAddress": "-",
      "IpPort": "-",
      "KeyLength": "0",
      "LmPackageName": "-",
      "LogonGuid": "{00000000-0000-0000-0000-000000000000}",
      "LogonProcessName": "-",
      "LogonType": "0",
      "ProcessId": "0x0000000000000004",
      "ProcessName": "",
      "SubjectDomainName": "-",
      "SubjectLogonId": "0x0000000000000000",
      "SubjectUserName": "-",
      "SubjectUserSid": "S-1-0-0",
      "TargetDomainName": "NT AUTHORITY",
      "TargetLogonId": "0x00000000000003e7",
      "TargetUserName": "SYSTEM",
      "TargetUserSid": "S-1-5-18",
      "TransmittedServices": "-",
      "WorkstationName": "-"
    },
    "System": {
      "Channel": "Security",
      "Computer": "37L4247E29-32",
      "Correlation": {},
      "EventID": "4624",
      "EventRecordID": "2",
      "Execution": {
        "ProcessID": "448",
        "ThreadID": "452"
      },
      "Keywords": "0x8020000000000000",
      "Level": "0",
      "Opcode": "0",
      "Provider": {
        "Guid": "{54849625-5478-4994-A5BA-3E3B0328C30D}",
        "Name": "Microsoft-Windows-Security-Auditing"
      },
      "Security": {},
      "Task": "12544",
      "TimeCreated": {
        "SystemTime": "2015-08-08T02:06:17.187500000Z"
      },
      "Version": "0"
    }
  }
}
```

As you can see, the dictionary is designed to have all values of the xml in an easy to access manner:
 - Tagnames are converted to keys, content is assigned as value
 - Attributes are converted to keys, values are assigned as value
 - Tags that have exactly one attribute called "Name" are specially treated and the "Name" is converted to a key, content is assigned as value
 - If a key has multiple values, it will be converted to an array
 - Text is usually assigned as value (see above). However, if there is already a subkey assigned, there will be a new "#text" key added.

I think this will cover most of the scenarios and will parse the `xml_string` in a very easy way.

Some additional thoughts on this:
 - Currently the field is called `xml`, it contains a dictionary and will be later converted to JSON. According to the naming scheme of `strings` and `strings_parsed` it would make sense to call it `xml_string_parsed`, however, I think the name is too long. I would suggest `xml_parsed` instead. Does something speak against this?
 - Performance-wise and for code quality it would make sense to only parse the XML once and then directly access the data there in order to fill the other fields. This requires additional effort and could be tracked in an additonal issue.  

**Related issue:** fixes #442 

## Checklist:
  - [ ] Automated checks (Travis, Codecov, Codefactor) pass
  - [ ] Reviewer assigned


joachimmetz:
@einfallstoll thanks for the PR but can you explain your reasoning behind making these changes? Especially since the approach significantly deviates from the attribute container approach used across plaso.

Also please have a look at https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md. Please follow the style convention (for maintenance sake) add the docstrings.

einfallstoll:
The XML contains additional data we want to parse during forensic analysis. However, as long as it's XML it requires lots of computing power for Kibana and Elasticsearch. If it's contained within a JSON it's possible to easily access the dynamic data without bloating up the current set of fields.

I don't think this is a significant deviation, since it just contains the `xml_string` in a different format. What do you think?

Thanks for the reference to the Style Guide. I will check this and update my PR if you think there's a chance for merge.

joachimmetz:
> I don't think this is a significant deviation, since it just contains the xml_string in a different format. What do you think?

The idea of event data (attribute containers) is to be typed information. Using a dict as a typeless container within a container defeats that principle.

einfallstoll:
But then we would have to get rid of `xml_string`? Or for what reason do we keep this?

joachimmetz:
> But then we would have to get rid of xml_string? 

Why?

> Or for what reason do we keep this?

To provide a somewhat human readable version of the event log entry in case the message string cannot be reconstructed or compare the message string against.

einfallstoll:
> Why?

Because in my opinion the new field (`xml`) is just a _different representation_ of what's already there (`xml_string`) so "Using a dict as a typeless container within a container defeats that principle." would apply to this as well.

If I was in your situation, I would probably argue the same, because it would make more sense to add the fields in typed container (as you called them above), so it would be nice to analyze the different EventLogs, pick out the fields for every single one of them, parse them and add them to the fieldset. That would probably be the best solution and I would agree with you on that.

However, this is how I interpreted #442 and this would add a low-cost benefit by parsing the XML and provide it in a Kibana- and Elasticsearch-friendlier format. Additional fields could be still parsed out of the XML and added at a later point.

joachimmetz:
> That would probably be the best solution and I would agree with you on that.

I would recommend let's focus on trying to accomplish this.

Can you elaborate what additional data you need from the XML that you currently don't have?

> However, this is how I interpreted #442 and this would add a low-cost benefit by parsing the XML and provide it in a Kibana- and Elasticsearch-friendlier format.

If this is only for kibana, then this be something more appropriate for an output module.

joachimmetz:
@einfallstoll friendly ping

* Can you elaborate what additional data you need from the XML that you currently don't have?
* does adding a kibana / changing the ES output module work for ?

plaso/parsers/winevtx.py
```diff
@@ -292,6 +297,50 @@ def _ParseRecords(self, parser_mediator, evtx_file):
             'unable to parse recovered event record: {0:d} with error: '
             '{1!s}').format(record_index, exception))
 
+  def _ElementTreeToDict(
```
joachimmetz: Please follow the plaso style guide https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md

joachimmetz: Please add unit tests for `_ElementTreeToDict`

plaso/parsers/winevtx.py
```diff
@@ -292,6 +297,50 @@ def _ParseRecords(self, parser_mediator, evtx_file):
...
+
+    if e.text and e.text.strip():
+      if d[t]:
+        d[t]["#text"] = e.text
+      else:
+        d[t] = e.text
+
+    return d
+
+  def _ElementTagName(
```
joachimmetz: Please add unit tests and follow style guide

plaso/parsers/winevtx.py
```diff
@@ -36,6 +36,7 @@ class WinEvtxRecordEventData(events.EventData):
     strings_parsed ([dict]): parsed information from event strings.
     user_sid (str): user security identifier (SID) stored in the event record.
     xml_string (str): XML representation of the event.
+    xml ([dict]): parsed information from XML string.
```
joachimmetz: Per conversation, since this is kibana specific move this into an output module instead.

joachimmetz:
Some more relevant context: https://github.com/log2timeline/plaso/issues/2129

einfallstoll:
So I had a discussion with my colleague and we dug a bit deeper, but came up with a hopefully better solution.

But first things first. We took [Security.evtx from JPCERTCC](https://github.com/JPCERTCC/LogonTracer/blob/master/sample/Security.evtx) as baseline for the analysis and found two problems (just examples):
- 4688 EventLogs are missing username and domain
- 4624 EventLogs are missing domain (however, they include the username in the field `strings_parsed`)

After a bit brainstorming we came up with this: [winevtx.py#L189](https://github.com/log2timeline/plaso/blob/master/plaso/parsers/winevtx.py#L189) currently uses a mapping dictionary to fill the field `strings_parsed` but only for a very limited set of fields. What if we get rid of the mapping and dynamically add the "EventData" from the `xml_string` to this field? This would keep the existing functionality (existing fields in `strings_parsed` are automatically covered), not add new fields (no `xml` field) and still gives more data (such as `target_domain_name`)?

joachimmetz:
> 4688 and 4624

Note that these IDs are not unique, you also need the qualifiers and provider identifier.
These can differ per Windows version/build. The parameters for the message string templates can be found in strings (represented by EventData in XML). Also see https://github.com/libyal/libevtx/blob/master/documentation/Windows%20XML%20Event%20Log%20(EVTX).asciidoc#541-message-strings

Note that some of these strings can be not set for a specific ID when not needed for the event message.

Note that the XML is just a different view on the event log entry data than message string and strings parameters.

> What if we get rid of the mapping and dynamically add the "EventData" from the xml_string to this field? 

I would need to check the role that strings_parsed plays.

einfallstoll:
> > 4688 and 4624
> 
> 
> 
> Note that these IDs are not unique, you also need the corresponding qualifiers and provider identifier.

Are we talking about the same? I mean a `4688` event not a specific one in the provided dataset. `winevtx.py` maps the `strings_parsed` based on this ID.

joachimmetz:
> Are we talking about the same? I mean a 4688 event not a specific one in the provided dataset. winevtx.py maps the strings_parsed based on this ID.

I mean this as general context. The event identifier is not global unique given an event log. It is unique local to an event provider with the qualifiers. Hence a potential explanation for "missing username and domain"

joachimmetz:
If the event data is set, it should be in the string e.g. SubjectDomainName = "EXAMPLE"

```
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
    <Provider Name="Microsoft-Windows-Security-Auditing" Guid="{54849625-5478-4994-a5ba-3e3b0328c30d}"/>
    <EventID>4688</EventID>
    <Version>0</Version>
    <Level>0</Level>
    <Task>13312</Task>
    <Opcode>0</Opcode>
    <Keywords>0x8020000000000000</Keywords>
    <TimeCreated SystemTime="2016-10-06T01:47:07.166302400Z"/>
    <EventRecordID>227127</EventRecordID>
    <Correlation/>
    <Execution ProcessID="4" ThreadID="80"/>
    <Channel>Security</Channel>
    <Computer>WIN-WFBHIBE5GXZ.example.co.jp</Computer>
    <Security/>
  </System>
  <EventData>
    <Data Name="SubjectUserSid">S-1-5-21-1524084746-3249201829-3114449661-500</Data>
    <Data Name="SubjectUserName">Administrator</Data>
    <Data Name="SubjectDomainName">EXAMPLE</Data>
    <Data Name="SubjectLogonId">0x0000000000032cfb</Data>
    <Data Name="NewProcessId">0x000000000000098c</Data>
    <Data Name="NewProcessName">C:\Windows\System32\cmd.exe</Data>
    <Data Name="TokenElevationType">%%1936</Data>
    <Data Name="ProcessId">0x0000000000000b74</Data>
  </EventData>
</Event>
```

Is also present here:
```
2016-10-06T01:47:07.166302+00:00,Content Modification Time,EVT,WinEVTX,[4688 / 0x1250] Source Name: Microsoft-Windows-Security-Auditing Strings: ['S-1-5-21-1524084746-3249201829-3114449661-500'  'Administrator'  'EXAMPLE'  '0x0000000000032cfb'  '0x000000000000098c'  'C:\Windows\System32\cmd.exe'  '%%1936'  '0x0000000000000b74'] Computer Name: WIN-WFBHIBE5GXZ.example.co.jp Record Number: 227127 Event Level: 0,winevtx,OS:/tmp/Security.evtx,-
```

Unfortunately you loose the context `<Data Name="SubjectDomainName">EXAMPLE</Data>` however that can be misleading if this string is not used by the message template.

joachimmetz:
And here you can see https://www.ultimatewindowssecurity.com/securitylog/encyclopedia/event.aspx?eventID=4688 that there are at least 2 different (known) message string template for this event identifier for the Microsoft-Windows-Security-Auditing provider.

einfallstoll:
> I mean this as general context. The event identifier is not global unique given an event log. It is unique local to an event provider with the qualifiers. Hence a potential explanation for "missing username and domain"

Here is an example of a `4624`. Some strings are parsed, some are not, hence the "domain is missing, username is not":
image

The domain is not only in the `xml_string` field, but in the `strings` field as well:
image

However, this works as intended, because the domain field is just not mapped in https://github.com/log2timeline/plaso/blob/master/plaso/parsers/winevtx.py#L73

A `4688` Event contains even less, because there is no mapping at all, hence no `strings_parsed` hence the `username and domain missing`:
image

There is not a directly accessible field containing the username or the domain, so one would have to parse the message or xml first and then access it.

And this is where I'm confused: Why is there even a static mapping? Why are there only two Event Identifiers mapped? Why is it not dynamically mapped based on the XML EventData as I suggested above?

joachimmetz:
> Here is an example

Thx for the context, this specifically applies to the strings_parsed attribute

> And this is where I'm confused: Why is there even a static mapping? Why are there only two Event Identifiers mapped? Why is it not dynamically mapped based on the XML EventData as I suggested above?

I need to double check, this might be left over from earlier days and strings_parsed attribute might just need to be removed.

joachimmetz:
Seeing the change strings_parsed was introduced in https://github.com/log2timeline/plaso/commit/2abe9e903dda2698e98238366a6dcf8602a95a4b. I assume this would have been relevant for https://github.com/log2timeline/eccemotus.

I'll make a note to remove it, since eccemotus has been deprecated and therefore this attribute is no longer needed.

joachimmetz:
Change list to remove the strings_parsed attribute https://github.com/log2timeline/plaso/pull/2860

einfallstoll:
Thanks for the information. This field basically was my excuse to add the `xml_string_parsed` field (see initial comment) and why I was so confused that you would say my field was "not ok" because of code design, but the `strings_parsed` was allowed somehow.

Nevertheless, for forensic analysis we need the data within the EventData tag of the XML (which is basically the same information as in `strings`, and a bit worse in `strings_parsed`). However, as long as the data is in the XML it can't be indexed easily without further effort and this leads me to the question:

Would you rather remove `strings_parsed` without replacement (I would not like this) or would you agree to get the EventData out of the XML and in some separate fields? I though of parsing the XML EventData tag and migrate all Data tags to a top-level key-value field (e.g. `<Data Name="SubjectDomainName">EXAMPLE</Data>` becomes `"subject_domain_name": "EXAMPLE"`). That way I would have all the data I want, and you would get rid of some legacy code and keep the design principle of only having typed fields and not "untyped containers within containers".

joachimmetz:
As I indicated earlier, I would recommend moving this into an output module; maybe a kibana specific one. That does the break-down of the XML into a dict there instead of in the parser.

einfallstoll:
What speaks against putting the data into separate fields?

Btw. Kibana was just an example, we use both Timesketch and Kibana.

joachimmetz:
> What speaks against putting the data into separate fields?

You mean as in additional class attributes? Not as in a dict?

But why stop there? Why overload WinEvtxRecordEventData for many different types of events, why not create separate events as outlined in #158

einfallstoll:
> But why stop there? Why overload WinEvtxRecordEventData for many different types of events, why not create separate events as outlined in #158

Ok, I see, I'm basically at the same point as you were in 2015. No need to say it like this.

I don't think a single output module would be suitable, since I would want this to parsed in all output variants, not just a specific one. And I'm still convinced something like `strings_parsed` or `xml_string_parsed` wouldn't be such a bad solution despite being not very clean. But it's hard to come up with a better solution. Or would you prefer to write a single parser for each and every event type? 😄

Maybe you can tell me how you would access those fields right now? What external tools do you use?

joachimmetz:
> Ok, I see, I'm basically at the same point as you were in 2015. No need to say it like this.

Just for context I mean these as reflecting questions. More as a way of thinking out loud, so to speak. Please don't take these personal in any way, more an artifact of my cultural communication preference.

> Or would you prefer to write a single parser for each and every event type?

I could imagine a way of mapping Windows event log messages to other event data types that lives in the winevtx parser. With a bit of thought into the design this can likely be done in a way that allows easy configuration and not too much overhead.

> Maybe you can tell me how you would access those fields right now? What external tools do you use?

First, let's do a step back, what problem are you trying to solve? This might be a different one than we are trying to solve.

A solution that has come up is to add message strings http://blog.kiddaland.net/2015/04/windows-event-log-message-strings.html

einfallstoll:
To quickly find and analyze Windows Event Logs (especially in the context of a forensic analysis with a super timeline) we want to access and relate the Event Logs to other entries. If data such as the username or target domain are available in a class attribute (or even in a simple dictionary) we can use Timesketch and Kibana to visualize and filter them.

As long as the data is stored in the XML, an array or multiple values within a single string, there is no clean (or simple) way to access the data without further parsing.

In my opinion the EventData tag in the XML contains valuable information that can and should be make accessible somehow and would help us (and hopefully others) to perform analysis based on this.

I just think that there are too many Event Types to map them manually (e.g. like `strings_parsed` was composed), so that's why I want to do a dynamic approach and process the EventData tag in the XML string.

joachimmetz:
It sounds to me that what you're looking for is something similar to TimeSketch Analyzers. https://blog.timesketch.org/2019/12/13/new-ui-same-great-functionality-new-wrapping/

joachimmetz:
@einfallstoll were you planning to move this to an output module? or Timesketch? Or should I close the PR since this is not in line with the direction chosen for event data in plaso.

joachimmetz:
Closing this issue due to lack of response original reporter.

PR #2853, opened 2020-03-21
Added support for directory with formatter configuration files

joachimmetz:
Added support for directory with formatter configuration files

### DIFF
--- a/plaso/formatters/manager.py
+++ b/plaso/formatters/manager.py
@@ -3,6 +3,9 @@
 
 from __future__ import unicode_literals
 
+import glob
+import os
+
 from plaso.formatters import default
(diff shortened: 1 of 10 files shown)


plaso/formatters/manager.py
```diff
@@ -184,3 +221,15 @@ def RegisterFormatters(cls, formatter_classes):
     """
     for formatter_class in formatter_classes:
       cls.RegisterFormatter(formatter_class)
+
+  @classmethod
+  def Reset(cls):
+    """Resets the manager to the hardcoded formatter classes.
+
+    This method is used for during unit testing.
```
Onager: -for

PR #2859, opened 2020-03-22

joachimmetz:


joachimmetz:
@Onager unfortunately there is not a specific break down for which formatters are being changed. So if mainly Android formatters are being changed in this PR it is coincidental.

PR #2860, opened 2020-03-24
Removed unused strings_parsed attribute from winevtx event data

joachimmetz:
Removed unused strings_parsed attribute from winevtx event data


joachimmetz:
@Onager PTAL

PR #2862, opened 2020-03-25

joachimmetz:


data/formatters/generic.yaml
```diff
@@ -151,6 +161,51 @@ short_source: 'LOG'
...
+short_message:
+- '[nickname: {nickname}]'
+- '{text}'
+short_source: 'LOG'
+source: 'XChat Log File'
+---
+type: 'conditional'
+data_type: 'xchat:scrollback:line'
+message:
+- 'Time elapsed: {elapsed_seconds} seconds'
```
rgayon, line 193: I don't see elapsed_seconds in the original formatter, is this wai?

joachimmetz: ack, good catch

joachimmetz:
@rgayon PTAL, rogue line removed and rebased

PR #2884, opened 2020-04-15
Changes for libfsntfs and dfvfs update

joachimmetz:
Changes for libfsntfs and dfvfs update


joachimmetz:
TODO: determine why AppVeyor win64 tests are failing and fix the issue

Issue #2889, opened 2020-04-20
Parser attribute not serialized


Onager:
Debug log line:
`2020-04-15 07:26:43,455 [DEBUG] (MainProcess) PID:4299 <json_serializer> [ConvertDictToObject] unsupported attribute name: event.parser`

joachimmetz:
```
Traceback (most recent call last):
  File "plaso/tools/log2timeline.py", line 97, in <module>
    if not Main():
  File "plaso/tools/log2timeline.py", line 72, in Main
    tool.ExtractEventsFromSources()
  File "plaso/plaso/cli/log2timeline_tool.py", line 448, in ExtractEventsFromSources
    worker_memory_limit=self._worker_memory_limit)
  File "plaso/plaso/multi_processing/task_engine.py", line 812, in ProcessSources
    self._ProcessSources(source_path_specs, storage_writer)
  File "plaso/plaso/multi_processing/task_engine.py", line 343, in _ProcessSources
    self._ScheduleTasks(storage_writer)
  File "plaso/plaso/multi_processing/task_engine.py", line 449, in _ScheduleTasks
    self._MergeTaskStorage(storage_writer)
  File "plaso/plaso/multi_processing/task_engine.py", line 256, in _MergeTaskStorage
    maximum_number_of_containers=self._MAXIMUM_NUMBER_OF_CONTAINERS)
  File "plaso/plaso/storage/sqlite/merge_reader.py", line 278, in MergeAttributeContainers
    self._active_container_type, serialized_data)
  File "plaso/plaso/storage/interface.py", line 600, in _DeserializeAttributeContainer
    attribute_container = self._serializer.ReadSerialized(serialized_string)
  File "plaso/plaso/serializer/json_serializer.py", line 404, in ReadSerialized
    return cls.ReadSerializedDict(json_dict)
  File "plaso/plaso/serializer/json_serializer.py", line 423, in ReadSerializedDict
    json_object = cls._ConvertDictToObject(json_dict)
  File "plaso/plaso/serializer/json_serializer.py", line 252, in _ConvertDictToObject
    'supported.').format(attribute_name))
ValueError: Event data attribute value: return_value of type dict is not supported.
```

joachimmetz:
The psort-linux-hashes end to end test seems to be passing again. Can you indicate if this log statement was possibly due to an older storage file? We can update the storage file format version number.

Issue #1691, opened 2018-01-27
Add structural fix for session parser counter


joachimmetz:
session parser counter not being filled, work around in https://github.com/log2timeline/plaso/issues/1687, but this needs structural fix:

* [x] move the parser attribute to event data - https://github.com/log2timeline/plaso/pull/3639
* [x] add tests https://github.com/log2timeline/plaso/pull/3698
* [x] ~~rename parser to parser_chain~~ - moved to https://github.com/log2timeline/plaso/issues/3700

Onager:
We also need to work out what we actually want to count here - events, eventdatas, both?

Onager:
@joachimmetz Can you elaborate on what needs the structural fix here? It's not clear to me what the structural issue is.

joachimmetz:
@Onager good question, I'll need to jog my memory

joachimmetz:
A comment in https://codereview.appspot.com/338420043/diff/1/plaso/parsers/mediator.py that seem to have dropped over time:
```
# TODO: rename this to event.parser_chain or equivalent.
```

joachimmetz:
An idea might have been to add counters based on the full parser chain instead of the last parser in the chain. I don't recall at this time. So maybe:

* rename parser to parser_chain
* add tests

joachimmetz:
Address comment in https://github.com/log2timeline/plaso/blob/master/plaso/parsers/mediator.py#L510

* move the parser attribute to event data
* make sure parser counters are updated and working

joachimmetz:
Changes https://github.com/log2timeline/plaso/pull/3639 to address moving the parser attribute to event data

joachimmetz:
> We also need to work out what we actually want to count here - events, eventdatas, both?

Keeping what we count as-is, number of events extracted/generated by parser or parser plugin

Issue #1687, opened 2018-01-26
Pinfo JSON output broken


Onager:
Parsers counter output should look like this:
```
    "parsers_counter": {
      "__type__": "collections.Counter",
      "hachoir": 8,
      "pls_recall": 2,
      "filestat": 383066,
      "utmp": 58,
      "syslog": 11827,
      "pe": 2,
      "total": 402039,
      "dpkg": 7076
    },
```

But looks like this:
```
    "parsers_counter": {
      "total": 402029,
      "__type__": "collections.Counter",
      "N/A": 402029
    },
```

Onager:
This is probably due to the event data split, now that I think about it - events no longer have a data_type to examine.

Onager:
https://github.com/log2timeline/plaso/blob/294d4927a98f9eb39d4e38182575b2047eb7bdd5/plaso/storage/interface.py#L1059 looks like part of the problem, counters aren't being updated as expected.

Onager:
I think the best short term fix (bearing in mind #1658 and probably further, related changes will affect this) is to add the parser chain information to the event (it's already stored on the event data).

I'm not immediately sure we need to set `parser` on both the event and event_data, but I suspect so. We want to be able to trace which parser/plugin produced these objects.

joachimmetz:
merged changes proposed in https://codereview.appspot.com/338420043/

Onager:
Fix in https://codereview.appspot.com/338420043/ wasn't complete.

Onager:
Fixed in https://github.com/log2timeline/plaso/commit/05a88037e5adc662f4b588b9bed8e9afa4b78451

PR #2892, opened 2020-04-20
Added parser as an event attribute #2889

Onager:
Per https://github.com/log2timeline/plaso/issues/1687#issuecomment-360735449 I think we want to track the origin of both events and event_datas, so kept the attribute on both.

Also added some extra tests to hopefully catch similar issues in the future.

https://github.com/log2timeline/plaso/issues/1691 is possibly related to this, but I think we'll still want to track what created the event objects themselves.

### DIFF
--- a/plaso/containers/events.py
+++ b/plaso/containers/events.py
@@ -17,6 +17,7 @@ class EventData(interface.AttributeContainer):
     data_type (str): event data type indicator.
     offset (int): offset relative to the start of the data stream where
         the event data is stored.
+    parser (str): string identifying the parser that produced the event data.
     query (str): query that was used to obtain the event data.
   """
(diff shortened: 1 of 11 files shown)


joachimmetz:
From the perspective of other parts of the code such as the formatters and output modules, there are now 2 overlapping parser values. What is the root cause of https://github.com/log2timeline/plaso/issues/2889 ? Can't the corresponding `event.parser` be changed into `event_data.parser`?

joachimmetz:
Per chat `event.parser` might be still need for pinfo parser statistics as well.

joachimmetz:
psort-linux-hashes end to end test is running again, so the question is do we need to make these changes? or maybe update the storage version number instead.

Issue #2890, opened 2020-04-20
End-to-end tests failing on MacOS image due to increased strictness of serializer


joachimmetz:
```
Traceback (most recent call last):
  File "plaso/tools/log2timeline.py", line 97, in <module>
    if not Main():
  File "plaso/tools/log2timeline.py", line 72, in Main
    tool.ExtractEventsFromSources()
  File "plaso/plaso/cli/log2timeline_tool.py", line 448, in ExtractEventsFromSources
    worker_memory_limit=self._worker_memory_limit)
  File "plaso/plaso/multi_processing/task_engine.py", line 812, in ProcessSources
    self._ProcessSources(source_path_specs, storage_writer)
  File "plaso/plaso/multi_processing/task_engine.py", line 343, in _ProcessSources
    self._ScheduleTasks(storage_writer)
  File "plaso/plaso/multi_processing/task_engine.py", line 449, in _ScheduleTasks
    self._MergeTaskStorage(storage_writer)
  File "plaso/plaso/multi_processing/task_engine.py", line 256, in _MergeTaskStorage
    maximum_number_of_containers=self._MAXIMUM_NUMBER_OF_CONTAINERS)
  File "plaso/plaso/storage/sqlite/merge_reader.py", line 278, in MergeAttributeContainers
    self._active_container_type, serialized_data)
  File "plaso/plaso/storage/interface.py", line 600, in _DeserializeAttributeContainer
    attribute_container = self._serializer.ReadSerialized(serialized_string)
  File "plaso/plaso/serializer/json_serializer.py", line 404, in ReadSerialized
    return cls.ReadSerializedDict(json_dict)
  File "plaso/plaso/serializer/json_serializer.py", line 423, in ReadSerializedDict
    json_object = cls._ConvertDictToObject(json_dict)
  File "plaso/plaso/serializer/json_serializer.py", line 252, in _ConvertDictToObject
    'supported.').format(attribute_name))
ValueError: Event data attribute value: return_value of type dict is not supported.
```

joachimmetz:
Looks like related to bsm parser https://github.com/log2timeline/plaso/blob/3c128a329e820b7b4d0ef6f78f09d32e732273e0/plaso/parsers/bsm.py#L36

joachimmetz:
`ValueError: Event data attribute value: return_value of type dict is not supported: {'error': 'Success', 'token_status': 0, 'call_status': 0}.`

Onager:
[...]

PR #2897, opened 2020-04-23
Changes to no longer support byte and dict EventData values #2890

joachimmetz:
Changes to no longer support dict attribute container values #2890


joachimmetz:
Made some changes, I've reused InvalidEvent since that is already raised by this function. @Onager PTAL

Issue #2912, opened 2020-05-02
Improve message formatting of empty Run/RunOnce key


joachimmetz:
While testing 20200430 noticed an issue on one of the test images in the dynamic output:

```
2010-11-10T17:42:41.390625+00:00,
Content Modification Time,
REG,
Registry Key : Run Key,

data_type: windows:registry:run parser: winreg/windows_run key_path: HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Run pathspec: <dfvfs.path.tsk_path_spec.TSKPathSpec object at 0x7f7a1f48b730> filename: /WINDOWS/system32/config/default.sav inode: 3327 display_name: TSK:/WINDOWS/system32/config/default.sav hostname: WKS-WINXP32BIT sha256_hash: c62983a4f58a696a871c6195eb0bd3090539711e9a6f7e9a5fedfbd6895fc0a1,

winreg/windows_run,
TSK:/WINDOWS/system32/config/default.sav,
```

'datetime', 'timestamp_desc', 'source', 'source_long', 'message', 'parser', 'display_name', 'tag'

It appears message is not correctly formatted.

joachimmetz:
Seems to be contained to Run and RunOnce winreg_plugins

joachimmetz:
Will add a work around for now, but have a look at a structural fix:

* changing the fallback behavior of the basic formatter
* change run/run once entries to a list of strings

PR #2914, opened 2020-05-02
Changes to prevent fallback formatting of empty Run key event #2912

joachimmetz:
Changes to prevent fallback formatting of empty Run key event #2912

### DIFF
--- a/plaso/parsers/winreg_plugins/run.py
+++ b/plaso/parsers/winreg_plugins/run.py
@@ -84,7 +84,9 @@ class AutoRunsPlugin(interface.WindowsRegistryPlugin):
       entries.append(value_string)
 
     event_data = RunKeyEventData()
-    event_data.entries = ' '.join(sorted(entries)) or None
+    # If entries is empty make sure it is set to an empty string otherwise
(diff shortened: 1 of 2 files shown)


plaso/parsers/winreg_plugins/run.py
```diff
@@ -84,7 +84,9 @@ def ExtractEvents(self, parser_mediator, registry_key, **kwargs):
       entries.append(value_string)
 
     event_data = RunKeyEventData()
-    event_data.entries = ' '.join(sorted(entries)) or None
+    # If entries is empty make sure it is set to an empty sting otherwise
```
Onager, line 87: string, not sting

joachimmetz: Not sure what you mean can you be more specific.

joachimmetz:
@Onager can you be a bit more verbose in your comments

plaso/parsers/winreg_plugins/run.py
joachimmetz: per conversation you meant there was a typo in "sting"

joachimmetz:
@Onager per conversation typo addressed

Issue #109, opened 2015-02-12
add multi volume support


joachimmetz:
Depends on phased processing changes
- [x] ~~change source scanner and collector to handle multiple file system path specs~~
- [x] ~~add support to process multi volumes to the extraction frontend~~
- [x] ~~add support for encrypted volumes for now BDE~~
  - ~~https://codereview.appspot.com/251910043~~
  - ~~make sure to rescan BDE for VSS in BDE~~
  - ~~determine which credentials are actually available in the BDE volume~~
- [x] ~~[add support to select a range of partitions/volumes](https://github.com/log2timeline/plaso/issues/537)~~
- [x] ~~fix preprocessor to handle more than one system configuration~~
  - **moved to https://github.com/log2timeline/plaso/issues/2286**
- [x] ~~fix tools to handle more than one source~~
  - **moved to https://github.com/log2timeline/plaso/issues/2287**
- [x] ~~Store scanning options in VolumeScanOptions objects ?~~
  - ~~https://codereview.appspot.com/340700043/~~
  - ~~make VolumeShadowSnapshotsOptions more generic?~~
  - ~~move part of this functionality into dfvfs~~
  - **moved to https://github.com/log2timeline/dfvfs/issues/341**
- [x] ~~print overview of volumes to be processed~~
  - **moved to https://github.com/log2timeline/plaso/issues/4610**

Related:
- ~~https://github.com/log2timeline/plaso/issues/180~~
- ~~https://codereview.appspot.com/238390043~~
- ~~https://codereview.appspot.com/232710043~~
- ~~https://codereview.appspot.com/246600043~~

f-s-p:
+1

pettai:
Is this issue about being able to understand Linux LVM too? (I noted that LVM was mentioned in https://github.com/log2timeline/plaso/issues/180) Or should I open a new issue?

It doesn't seem to it work ATM:

-bash-4.3$ log2timeline.py plaso.dump clone.vmdk
[INFO] Data files will be loaded from /usr/share/plaso by default.
The following partitions were found:
Identifier	Offset (in bytes)	Size (in bytes)
p1		1048576 (0x00100000)	243.0MiB / 254.8MB (254803968 B)
p2		256901120 (0x0f500000)	19.8GiB / 21.2GB (21216886784 B)

Please specify the identifier of the partition that should be processed.
All partitions can be defined as: "all". Note that you can abort with Ctrl^C.
p2
[...]

PR #2916, opened 2020-05-03
Renamed ReadPreprocessingInformation to ReadSystemConfiguration #109

joachimmetz:
Renamed ReadPreprocessingInformation to ReadSystemConfiguration #109

### DIFF
--- a/plaso/storage/fake/writer.py
+++ b/plaso/storage/fake/writer.py
@@ -427,16 +427,14 @@ class FakeStorageWriter(interface.StorageWriter):
           task.identifier))
 
   # pylint: disable=unused-argument
-  def ReadPreprocessingInformation(self, knowledge_base):
-    """Reads preprocessing information.
+  def ReadSystemConfiguration(self, knowledge_base):
+    """Reads system configuration information.
 
-    The preprocessing information contains the system configuration which
-    contains information about various system specific configuration data,
-    for example the user accounts.
+    The system configuration contains information about various system specific
+    configuration data, for example the user accounts.
 
     Args:
-      knowledge_base (KnowledgeBase): is used to store the preprocessing
-          information.
+      knowledge_base (KnowledgeBase): is used to store the system configuration.
 
     Raises:
       IOError: if the storage type does not support writing preprocessing
--- a/plaso/storage/file_interface.py
+++ b/plaso/storage/file_interface.py
@@ -405,18 +405,16 @@ class StorageFileReader(interface.StorageReader):
     """
     return self._storage_file.HasWarnings()
 
-  def ReadPreprocessingInformation(self, knowledge_base):
-    """Reads preprocessing information.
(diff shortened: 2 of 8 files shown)


plaso/storage/fake/writer.py
```diff
@@ -524,6 +502,27 @@ def WriteSessionCompletion(self, aborted=False):
...
+    Raises:
+      IOError: if the storage type does not support writing preprocessing
+          information or when the storage writer is closed.
+      OSError: if the storage type does not support writing preprocessing
+          information or when the storage writer is closed.
+    """
+    self._RaiseIfNotWritable()
+
+    if self._storage_type != definitions.STORAGE_TYPE_SESSION:
+      raise IOError('Session configuration not supported by storage type.')
```
Onager, line 522: Please add the actual storage type to the error message here

plaso/storage/file_interface.py
```diff
@@ -1085,6 +1064,26 @@ def WriteSessionCompletion(self, aborted=False):
...
+    Raises:
+      IOError: if the storage type does not support writing preprocessing
+          information or when the storage writer is closed.
+      OSError: if the storage type does not support writing preprocessing
+          information or when the storage writer is closed.
+    """
+    self._RaiseIfNotWritable()
+
+    if self._storage_type != definitions.STORAGE_TYPE_SESSION:
+      raise IOError('Session configuration not supported by storage type.')
```
Onager, line 1083: Please add the actual storage type to the error message here

plaso/storage/fake/writer.py
```diff
@@ -447,7 +446,7 @@ def ReadPreprocessingInformation(self, knowledge_base):
     self._RaiseIfNotWritable()
 
     if self._storage_type != definitions.STORAGE_TYPE_SESSION:
-      raise IOError('Preprocessing information not supported by storage type.')
+      raise IOError('Session configuration not supported by storage type.')
```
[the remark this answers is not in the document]
joachimmetz: Technically this can only be task storage at the moment. So IMHO this does not add much value at the moment.

plaso/storage/fake/writer.py
joachimmetz: Same reply as previous comment.

plaso/storage/file_interface.py
joachimmetz: Same reply as previous comment.

joachimmetz:
@Onager PTAL, IMHO adding the storage type indication at this point (seeing we only have 2 types) does not add much value.

joachimmetz:
@Onager I'm having a second look at this and this might not be entirely the path I want to take. Since with the system configuration stored in the session, the ReadPreprocessingInformation might only be needed for backwards compatibility.

joachimmetz:
Re-aligned changes with https://github.com/log2timeline/plaso/pull/2936. @Onager PTAL

PR #2936, opened 2020-05-18
Added session configuration attribute container #109

joachimmetz:
Added session configuration attribute container #2946


plaso/containers/sessions.py
```diff
@@ -185,8 +237,8 @@ def __init__(self, identifier=None):
     self.timestamp = None
 
 
-class SessionStart(interface.AttributeContainer):
-  """Session start attribute container.
+class SessionConfiguration(interface.AttributeContainer):
+  """Session configuration attribute container.
```
Onager, line 241: Please add a description of what a session configuration is, and what sort of information should be stored in one.

joachimmetz: done

PR #2941, opened 2020-05-20
Changed pylint CI test to Ubuntu 20.04

joachimmetz:
Changed pylint CI test to Ubuntu 20.04


joachimmetz:
Updated

plaso/cli/helpers/analysis_plugins.py
```diff
@@ -61,6 +61,7 @@ def ParseOptions(cls, options, configuration_object):
 
     Raises:
       BadConfigObject: when the configuration object is of the wrong type.
+      BadConfigOption: when a non-existing analysis plugins are specified.
```
Onager, line 64: typo, should be "plugin" not "plugins"

joachimmetz: I'll remove the `a` to match the error string in the code.

Issue #434, opened 2015-12-12
Improve date estimation for parsers that use GetEstimatedYear


joachimmetz:
Improve date estimation for parsers that use GetEstimatedYear
- mac_appfirewall.py
- mac_securityd.py
- mac_wifi.py
- syslog.py
- google_logging.py 

Also see if `self._year_use` approach can be improved.

* [x] move year-less format handling into mix-in - https://github.com/log2timeline/plaso/pull/4262
* [x] Split artifact and event extraction across multiple processing phases
* [x] Add dynamic-time support for year-less date and time values - https://github.com/log2timeline/plaso/issues/4266

See if this addresses https://github.com/log2timeline/plaso/issues/2871

PR #2943, opened 2020-05-20
Added parser for Google log formatted log files

Onager:
Parser for "Google Log" formatted files, as produced by:
https://github.com/google/glog
https://github.com/golang/glog

etc.

### DIFF
--- a/data/formatters/generic.yaml
+++ b/data/formatters/generic.yaml
@@ -149,6 +149,17 @@ short_source: 'LOG'
 source: 'GoogleDriveSync Log File'
 ---
 type: 'conditional'
+data_type: 'googlelog:log'
+message:
+  - '{file_name}:'
+  - '{line_number}]'
+  - '{message}'
+short_message:
+  - '{message}'
+short_source: 'LOG'
+source: 'Google Log'
+---
+type: 'conditional'
 data_type: 'java:download:idx'
 message:
 - 'IDX Version: {idx_version}'
--- a/plaso/parsers/google_logging.py
+++ b/plaso/parsers/google_logging.py
@@ -19,21 +19,25 @@ from plaso.parsers import text_parser
 class GoogleLogEventData(events.EventData):
   """Google log file event data.
 
-  See: https://github.com/google/glog
+  See: https://github.com/google/glog. This format is also used by
+  Kubernetes, see https://github.com/kubernetes/klog
 
   Attributes:
     file_name (str): the name of the source file that logged the message.
     line_number (int): the line number in the source file where the logging
-      statement is.
+        statement is.
     message (str): the log message.
-    priority (str): the priority of the message - I, W, E or F.
-    thread_id (int): the ID of the thread that recorded the message.
+    priority (str): the priority of the message - I, W, E or F. These values
+        represent messages logged at INFO, WARNING, ERROR or FATAL severities,
+        respectively.
+    thread_identifier (int): the identifier of the thread that recorded the
+        message.
   """
 
   DATA_TYPE = 'googlelog:log'
 
   def __init__(self, data_type=DATA_TYPE):
-    """Initializes the event object.
+    """Initializes event data.
 
     Args:
       data_type (Optional[str]): event data type indicator.
@@ -43,7 +47,7 @@ class GoogleLogEventData(events.EventData):
     self.line_number = None
     self.message = None
     self.priority = None
-    self.thread_id = None
+    self.thread_identifier = None
 
 
 class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
@@ -57,8 +61,8 @@ class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
 
   # PyParsing components used to construct grammars for parsing lines.
   _PYPARSING_COMPONENTS = {
-      'priority':
-          (pyparsing.oneOf(['I', 'W', 'E', 'F']).setResultsName('priority')),
+      'priority': (
+          pyparsing.oneOf(['I', 'W', 'E', 'F']).setResultsName('priority')),
       'year': text_parser.PyparsingConstants.FOUR_DIGITS.setResultsName(
           'year'),
       'month_number': text_parser.PyparsingConstants.TWO_DIGITS.setResultsName(
@@ -73,15 +77,16 @@ class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
           'second'),
       'microsecond': pyparsing.Word(pyparsing.nums, exact=6).setParseAction(
           text_parser.PyParseIntCast).setResultsName('microsecond'),
-      'thread_id': pyparsing.Word(pyparsing.nums).setResultsName('thread_id'),
+      'thread_identifier': pyparsing.Word(pyparsing.nums).setResultsName(
+          'thread_identifier'),
       'file_name':
           (pyparsing.Word(pyparsing.printables.replace(':', '')).setResultsName(
               'file_name')),
-      'line_number':
-          (pyparsing.Word(pyparsing.nums).setResultsName('line_number')),
-      'message':
-          (pyparsing.Regex('.*?(?=($|\n[IWEF][0-9]{4}))', re.DOTALL).
-           setResultsName('message'))}
+      'line_number': (
+          pyparsing.Word(pyparsing.nums).setResultsName('line_number')),
+      'message': (
+          pyparsing.Regex('.*?(?=($|\n[IWEF][0-9]{4}))', re.DOTALL).
+          setResultsName('message'))}
 
   _PYPARSING_COMPONENTS['date'] = (
       _PYPARSING_COMPONENTS['month_number'] +
@@ -94,8 +99,8 @@ class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
 
   # Grammar for individual log event lines.
   _LOG_LINE = (
-      _PYPARSING_COMPONENTS['priority'] + _PYPARSING_COMPONENTS['date']
-      + _PYPARSING_COMPONENTS['thread_id'] +
+      _PYPARSING_COMPONENTS['priority'] + _PYPARSING_COMPONENTS['date'] +
+      _PYPARSING_COMPONENTS['thread_identifier'] +
       _PYPARSING_COMPONENTS['file_name'] + pyparsing.Suppress(':') +
       _PYPARSING_COMPONENTS['line_number'] + pyparsing.Suppress('] ') +
       _PYPARSING_COMPONENTS['message'] + pyparsing.lineEnd())
@@ -129,7 +134,7 @@ class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
   _SUPPORTED_KEYS = frozenset([key for key, _ in LINE_STRUCTURES])
 
   def __init__(self):
-    """Initializes a parser object."""
+    """Initializes a Google log formatted file parser."""
     super(GoogleLogParser, self).__init__()
 
     # Set the size of the file we need to read to verify it.
@@ -206,7 +211,8 @@ class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
 
     event_data = GoogleLogEventData()
     event_data.priority = self._GetValueFromStructure(structure, 'priority')
-    event_data.thread_id = self._GetValueFromStructure(structure, 'thread_id')
+    event_data.thread_identifier = self._GetValueFromStructure(
+        structure, 'thread_identifier')
     event_data.file_name = self._GetValueFromStructure(structure, 'file_name')
     event_data.line_number = self._GetValueFromStructure(
         structure, 'line_number')
--- a/plaso/parsers/google_logging.py
+++ b/plaso/parsers/google_logging.py
@@ -12,7 +12,6 @@ from plaso.containers import events
 from plaso.containers import time_events
 from plaso.lib import errors
 from plaso.lib import definitions
-from plaso.lib import timelib
 from plaso.parsers import manager
 from plaso.parsers import text_parser
 
@@ -24,15 +23,15 @@ class GoogleLogEventData(events.EventData):
   See: https://github.com/google/glog
 
   Attributes:
-    priority (str): the priority of the body - I, W, E or F.
-    thread_id (int): the ID of the thread that recorded the body.
-    file_name (str): the name of the source file that logged the body.
+    priority (str): the priority of the message - I, W, E or F.
+    thread_id (int): the ID of the thread that recorded the message.
+    file_name (str): the name of the source file that logged the message.
     line_number (int): the line number in the source file where the logging
       statement is.
-    body (str): the body logged.
+    message (str): the log message.
   """
 
-  DATA_TYPE = 'googlelog:body'
+  DATA_TYPE = 'googlelog:log'
 
   def __init__(self, data_type=DATA_TYPE):
     """Initializes the event object.
@@ -45,7 +44,7 @@ class GoogleLogEventData(events.EventData):
     self.thread_id = None
     self.file_name = None
     self.line_number = None
-    self.body = None
+    self.message = None
 
 
 class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
@@ -58,20 +57,14 @@ class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
 
   DESCRIPTION = 'Parser for handling Google log formatted files.'
 
-  _GREETING_START = 'Log file created at: '
-
-  # Our initial buffer length is the length of the string we verify with.
-  _INITIAL_BUFFER_SIZE = len(_GREETING_START)
-  # Once we're sure we're reading a valid file, we'll use a larger read buffer.
-  _RUNNING_BUFFER_SIZE = 5120
-
   # PyParsing Components used to construct grammars for parsing lines.
   _PYPARSING_COMPONENTS = {
       'priority':
           (pyparsing.oneOf(['I', 'W', 'E', 'F']).setResultsName('priority')),
       'year': text_parser.PyparsingConstants.FOUR_DIGITS.setResultsName(
           'year'),
-      'month': text_parser.PyparsingConstants.MONTH.setResultsName('month'),
+      'month_number': text_parser.PyparsingConstants.TWO_DIGITS.setResultsName(
+          'month_number'),
       'day': text_parser.PyparsingConstants.ONE_OR_TWO_DIGITS.setResultsName(
           'day'),
       'hour': text_parser.PyparsingConstants.TWO_DIGITS.setResultsName(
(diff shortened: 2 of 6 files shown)


plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+import pyparsing
+
+from plaso.containers import events
+from plaso.containers import time_events
+from plaso.lib import errors
+from plaso.lib import definitions
+from plaso.parsers import manager
+from plaso.parsers import text_parser
+
+
```
joachimmetz, line 18: `-1` white line

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+
+class GoogleLogEventData(events.EventData):
+  """Convenience class for events from a Google log file.
```
joachimmetz, line 21: Please change to "Google log file event data." 

"Convenience class"  has a specific meaning https://en.wiktionary.org/wiki/convenience_class

Did you mean "value object" https://en.wikipedia.org/wiki/Value_object ?

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+
+  See: https://github.com/google/glog
+
+  Attributes:
+    priority (str): the priority of the message - I, W, E or F.
```
joachimmetz, line 26: https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md#class-attributes

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+    self.file_name = None
+    self.line_number = None
+    self.message = None
+
+
+class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
+  """A parser for handling google log formatted files.
+
+  See https://google.io/docs/python/guides/logging
+
```
joachimmetz, line 54: `-1` white line

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+    super(GoogleLogEventData, self).__init__(data_type=data_type)
+    self.priority = None
+    self.thread_id = None
+    self.file_name = None
+    self.line_number = None
+    self.message = None
+
+
+class GoogleLogParser(text_parser.PyparsingMultiLineTextParser):
+  """A parser for handling google log formatted files.
```
joachimmetz, line 51: google => Google 

Also please use one form "Google-formatted log", "Google logs formatted"

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+
+  See https://google.io/docs/python/guides/logging
+
+  """
+  NAME = 'googlelog'
+
+  DESCRIPTION = 'Parser for handling Google log formatted files.'
+
+  # PyParsing Components used to construct grammars for parsing lines.
```
joachimmetz, line 60: drop the capital C
PyParsing Components => PyParsing components

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+          'day'),
+      'hour': text_parser.PyparsingConstants.TWO_DIGITS.setResultsName(
+          'hour'),
+      'minute': text_parser.PyparsingConstants.TWO_DIGITS.setResultsName(
+          'minute'),
+      'second': text_parser.PyparsingConstants.TWO_DIGITS.setResultsName(
+          'second'),
+      'microsecond': pyparsing.Word(pyparsing.nums, exact=6).setParseAction(
+          text_parser.PyParseIntCast).setResultsName('microsecond'),
+      'thread_id': pyparsing.Word(pyparsing.nums).setResultsName('thread_id'),
```
joachimmetz, line 78: id => identifier 

Also see: https://github.com/log2timeline/l2tdocs/blob/master/process/Style-guide.md#naming

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+  # last.
+  LINE_STRUCTURES = [
+      ('log_entry', _LOG_LINE),
+      ('greeting_start', pyparsing.Literal(_GREETING_START)),
+      ('greeting', _GREETING)]
+
+  _SUPPORTED_KEYS = frozenset([key for key, _ in LINE_STRUCTURES])
+
+  def __init__(self):
+    """Initialize a parser object."""
```
joachimmetz, line 134: Initialize => Initializes

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+    super(GoogleLogParser, self).__init__()
+
+    # Set the size of the file we need to read to verify it.
+    self._buffer_size = self._INITIAL_BUFFER_SIZE
+    self._maximum_year = 0
+    # The year to use for events. The initial year is stored in the log file
+    # greeting.
+    self._year_use = 0
```
joachimmetz, line 142: Consider using `_current_year` or just `_year` ?

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+    # The month the last observed event occurred.
+    self._last_month = 0
+
+  def _UpdateYear(self, mediator, month):
+    """Updates the year to use for events, based on last observed month.
+
+    Args:
+      mediator (ParserMediator): mediates the interactions between
```
joachimmetz, line 150: This is the description used in other parsers (except for syslog):
```
      parser_mediator (ParserMediator): mediates interactions between parsers
          and other components, such as storage and dfvfs.
```

Please change it, I'll create a separate PR to address the other syslog docstrings

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+    """Extract useful information from the logfile greeting.
+
+    Args:
+      structure (pyparsing.ParseResults): elements parsed from the file.
+    """
+    self._year_use = self._GetValueFromStructure(structure, 'year')
+    self._last_month = self._GetValueFromStructure(structure, 'month_number')
+
+  def _ParseLine(self, parser_mediator, structure):
+    """Process a single log event into a GoogleLogEvent.
```
joachimmetz, line 179: Change "log event" => "log line"

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+      self._year_use = mediator.GetEstimatedYear()
+    if not self._maximum_year:
+      self._maximum_year = mediator.GetLatestYear()
+
+    if not self._last_month:
+      self._last_month = month
+      return
+
+    # TODO: Check whether out of order events are possible
+    if self._last_month > (month + 1):
```
joachimmetz, line 164: Why `month + 1` ? I assume `self._last_month` contains a value from 1 upto and including 12
and month as well. So if the logfile would jump from Dec 2018 to Nov 2019 the year would stay the same?

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,264 @@
...
+      parser_mediator (ParserMediator): mediates interactions between
+          parsers and other components, such as storage and dfvfs.
+      lines (str): one or more lines from the text file.
+
+    Returns:
+      bool: True if this is the correct parser, False otherwise.
+    """
+    if not lines.startswith(self._GREETING_START):
+      return False
+
```
joachimmetz, line 256: If the "greeting" contains a year and month value let's check if these are valid as well.

joachimmetz: maybe validate the entire date?

tests/parsers/google_logging.py
```diff
@@ -0,0 +1,101 @@
...
+
+import unittest
+
+from plaso.lib import errors
+from plaso.parsers import google_logging
+
+from tests.parsers import test_lib
+
+
+class GooglelogTest(test_lib.ParserTestCase):
```
joachimmetz, line 15: Let's keep the name of this class consistent with most other parsers and rename to:  GoogleLogParserTest

tests/parsers/google_logging.py
```diff
@@ -0,0 +1,101 @@
...
+  """Tests for the Google logging parser"""
+
+  def testEventsParsed(self):
```
joachimmetz, line 18: Can we add tests for the individual methods as well. So that we know these are properly tested for error cases as well.

I'm wondering why this parser unittest is deviating so much from other ones? Is this a left over from earlier work?

tests/parsers/google_logging.py
```diff
@@ -0,0 +1,101 @@
...
+        'logging_functional_test_helper.py: 65] This line is log level 0')
+    expected_short = 'This line is log level 0'
+    self._TestGetMessageStrings(event_data, expected_string, expected_short)
+
+  def testRaisesUnableToParseForInvalidFiles(self):
+    """Test that attempting to parse an invalid file should raise an error."""
+    parser = google_logging.GoogleLogParser()
+    knowledge_base_values = {'year': 2020}
+
+    for invalid_file in ['access.log']:
```
joachimmetz, line 94: remove the for loop it adds no value

tests/parsers/google_logging.py
```diff
@@ -0,0 +1,101 @@
...
+    self.CheckTimestamp(event.timestamp, '2019-12-31 23:59:59.000002')
+
+    event_data = self._GetEventDataOfEvent(storage_writer, event)
+    expected_string = (
+        'logging_functional_test_helper.py: 65] This line is log level 0')
+    expected_short = 'This line is log level 0'
+    self._TestGetMessageStrings(event_data, expected_string, expected_short)
+
+  def testRaisesUnableToParseForInvalidFiles(self):
+    """Test that attempting to parse an invalid file should raise an error."""
```
joachimmetz, line 90: add check to see if access.log exists otherwise this test will likely fail

plaso/parsers/google_logging.py
Onager: The parameter in the library is consistently referred to as an 'id' and never as an 'identifier': https://github.com/google/glog/blob/5b4fb63d277795eea3400e3e6af542f3b765f2d2/src/glog/logging.h.in#L293 etc.

I think it's better to match the source here, so keeping this as 'thread_id'

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Done - "Google log" is the log format.

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Thats correct, this is the same logic as used in the syslog parser, where events can sometimes occur out of order. In your example, this code assumes that if an event from November follows an event from December 2018, that the event occurred in 2018, not 2019.

plaso/parsers/google_logging.py
Onager: Done

plaso/parsers/google_logging.py
Onager: Is there a particular false positive case you're concerned about? Adding the validation is going to complicate the logic quite a bit, and I think the likelihood and impact of starting to parse a file with a different format are very low.

tests/parsers/google_logging.py
Onager: Done

tests/parsers/google_logging.py
Onager: Done

tests/parsers/google_logging.py
Onager: Done

tests/parsers/google_logging.py
Onager: Changed this to be more like the other parser tests.

plaso/parsers/google_logging.py
joachimmetz: Ack, added to https://github.com/log2timeline/plaso/issues/434

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,262 @@
...
+  # last.
+  LINE_STRUCTURES = [
+      ('log_entry', _LOG_LINE),
+      ('greeting_start', pyparsing.Literal(_GREETING_START)),
+      ('greeting', _GREETING)]
+
+  _SUPPORTED_KEYS = frozenset([key for key, _ in LINE_STRUCTURES])
+
+  def __init__(self):
+    """Initializes a parser object."""
```
joachimmetz, line 132: Please change to: Initializes a Google log formatted file parser

plaso/parsers/google_logging.py
joachimmetz: Why is this "better"? Why is this unique for this specific format? Why is it not clear from thread_identifier that this field is referenced?

Please add your rationale why to deviate from previous established guidelines.

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,262 @@
...
+      _PYPARSING_COMPONENTS['hour'] + pyparsing.Suppress(':') +
+      _PYPARSING_COMPONENTS['minute'] + pyparsing.Suppress(':') +
+      _PYPARSING_COMPONENTS['second'] + pyparsing.Optional(
+          pyparsing.Suppress('.') +
+          _PYPARSING_COMPONENTS['microsecond']))
+
+  # Grammar for individual log event lines.
+  _LOG_LINE = (
+      _PYPARSING_COMPONENTS['priority'] + _PYPARSING_COMPONENTS['date']
+      + _PYPARSING_COMPONENTS['thread_id'] +
```
joachimmetz, line 98: Why the `+` at the start of the line, where everywhere else it is at the end?
Please change to make this consistent.

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,262 @@
...
+
+  See https://google.io/docs/python/guides/logging
+  """
+  NAME = 'googlelog'
+
+  DESCRIPTION = 'Parser for handling Google log formatted files.'
+
+  # PyParsing components used to construct grammars for parsing lines.
+  _PYPARSING_COMPONENTS = {
+      'priority':
```
joachimmetz, line 60: Please put `(` on the end of the line so you don't need 1 additional space for continuation indentation, which is also more consistent with the other definitions in this dict.

tests/parsers/google_logging.py
```diff
@@ -0,0 +1,66 @@
...
+    expected_short = 'This line is log level 0'
+    self._TestGetMessageStrings(event_data, expected_string, expected_short)
+
+   # Test a multiline event.
+    multi_line_event = events[2]
+    multi_line_event_data = self._GetEventDataOfEvent(
+        storage_writer, multi_line_event)
+    multi_line_message = multi_line_event_data.message
+
+    self.assertIn('\n', multi_line_message)
```
joachimmetz, line 48: Maybe make this check more specific and use split('\n') and check the number of resulting lines.

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,262 @@
...
+  """Google log file event data.
+
+  See: https://github.com/google/glog
+
+  Attributes:
+    file_name (str): the name of the source file that logged the message.
+    line_number (int): the line number in the source file where the logging
+      statement is.
+    message (str): the log message.
+    priority (str): the priority of the message - I, W, E or F.
```
joachimmetz, line 29: Please add a quick explanation what 'I', 'W' etc represent.

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,262 @@
...
+    thread_id (int): the ID of the thread that recorded the message.
```
joachimmetz, line 30: thread_id => thread_identifier

data/formatters/generic.yaml
```diff
@@ -149,6 +149,17 @@ short_source: 'LOG'
 source: 'GoogleDriveSync Log File'
 ---
 type: 'conditional'
+data_type: 'googlelog:log'
+message:
+  - '{file_name}:'
+  - '{line_number}]'
```
joachimmetz, line 155: There is a `]` in this message part but no `[` is that intended? Do you also want add context what these values represent e.g. `Line number: {line_number}` ?

plaso/parsers/google_logging.py
joachimmetz: So `Log file created at: ` IMHO is not that unique. I can imagine that other log formats use this as well. Hence the request to add additional validation.

plaso/parsers/google_logging.py
```diff
@@ -0,0 +1,262 @@
...
+
+  # Order is important here, as the structures are checked against each line
+  # sequentially, so we put the most common first, and the most expensive
+  # last.
+  LINE_STRUCTURES = [
+      ('log_entry', _LOG_LINE),
+      ('greeting_start', pyparsing.Literal(_GREETING_START)),
+      ('greeting', _GREETING)]
+
+  _SUPPORTED_KEYS = frozenset([key for key, _ in LINE_STRUCTURES])
```
joachimmetz, line 129: list comprehension is likely not needed here.
=> `_SUPPORTED_KEYS = frozenset(LINE_STRUCTURES.keys())`

plaso/parsers/google_logging.py
joachimmetz: Could you do something with the additional header lines:
```
Running on machine: plasotest1
Binary: Built on May 14 2019 17:08:01 (1368576481)
Binary: Built for gcc-4.X.Y
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
```

test_data/googlelog_test.INFO
```diff
@@ -0,0 +1,10 @@
+Log file created at: 2019/07/18 06:07:40
+Running on machine: plasotest1
+Binary: Built on May 14 2019 17:08:01 (1368576481)
```
joachimmetz, line 3: Should we generate an event for the build date?

plaso/parsers/google_logging.py
joachimmetz: ack, I see, treacherous `, _`

data/formatters/generic.yaml
joachimmetz: IMHO not required, it is now user configurable, people can change it if they don't like the defaults. But might be worth a follow up conversation about some unification here like the event data attributes.

plaso/parsers/google_logging.py
joachimmetz: ack

plaso/parsers/google_logging.py
joachimmetz: Did a rough search but most of the tools seem to use glog under the hood. Also
```
Binary: Built on May 14 2019 17:08:01 (1368576481)
Binary: Built for gcc-4.X.Y"
```
Seem to be not required lines

Issue #2926, opened 2020-05-15
Elasticsearch output error: Root mapping definition has unsupported parameters


utkonos:
**Description of problem:**

According to the commit logs, elasticsearch 7 is supported, but after running with a fresh install, the output plugin fails with an error.

**Command line and arguments:**

```
psteal.py -o elastic --server localhost --port 9200 --index_name test1 --raw_fields --source image.img -w test1.plaso
```

**Source data:**

I used a VirtualBox VM with a fresh install of windows 7 32bit. Drive size 16GB. Converted vdi image to raw with this command:
```
VBoxManage clonehd Win7x32.vdi image.img --format raw
```

**Plaso version:**

plaso-20200430 installed from GIFT PPA
Elasticsearch 7.7.0

**Operating system Plaso is running on:**

Ubuntu Server 20.04

**Installation method:**

```
#!/bin/bash
## Set environment
export DEBIAN_FRONTEND=noninteractive

## Update Ubuntu
apt-get update
apt-get upgrade -y
apt-get autoremove -y
apt-get autoclean -y

## Install Plaso
add-apt-repository ppa:gift/stable
apt-get update
apt-get install -y plaso-tools
apt-get upgrade -y

## Add Plaso user
adduser --gecos '' --disabled-password plaso
```
Elasticsearch installed like this:
```
wget -qO - https://artifacts.elastic.co/GPG-KEY-elasticsearch | sudo apt-key add -
echo "deb https://artifacts.elastic.co/packages/7.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-7.x.list
apt-get update
apt-get install -y elasticsearch
apt-get install -y kibana
service elasticsearch start
service kibana start
systemctl enable elasticsearch
systemctl enable kibana
```

**Debug output/tracebacks:**

```
Traceback (most recent call last):
  File "/usr/bin/psteal.py", line 110, in <module>
    if not Main():
  File "/usr/bin/psteal.py", line 87, in Main
    tool.AnalyzeEvents()
  File "/usr/lib/python3/dist-packages/plaso/cli/psteal_tool.py", line 218, in AnalyzeEvents
    analysis_engine.ExportEvents(
  File "/usr/lib/python3/dist-packages/plaso/multi_processing/psort.py", line 997, in ExportEvents
    output_module.WriteHeader()
  File "/usr/lib/python3/dist-packages/plaso/output/elastic.py", line 72, in WriteHeader
[...]

PR #2953, opened 2020-05-22
Changes for Elasticsearch 7 and raw fields output #2926

joachimmetz:
Changes for Elasticsearch 7 and raw fields output #2953


joachimmetz:
@berggren can you have a look at the changes in the last commit if that matches your understanding of Elasticsearch 7

Issue #2956, opened 2020-05-22
Move analyzer extra event attributes and path specification to event data stream


joachimmetz:
The attributes generated by analyzers (hasher, yara scan, etc.) are currently [set](https://github.com/log2timeline/plaso/blob/a7365709efc09874bcdff86df89f4b7f01975bb6/plaso/engine/worker.py#L232) on each [event data object](https://github.com/log2timeline/plaso/blob/a7365709efc09874bcdff86df89f4b7f01975bb6/plaso/parsers/mediator.py#L452) by the parser mediator.

Since analyzers operate on data stream, having analyzer attributes on every event data object is unnecessary storage overhead. Event data stream would make a better fit however the currently:

* event data stream once generated is not expected to be changed
* event data stream currently represents an individual file (system) entry not an individual data stream, which are used (by the task manager) as unit of processing.
  * If data stream should become a unit of processing how to handle file entry-based events (file_stat), since a single file entry can have multiple data streams.

Possible options:

1. Have task manager pass individual data streams as tasks instead of file entries.
  * Store the analyzer attributes separately in the task storage and merge these into the event source after processing.
 * Handle not duplicating file entry-based events.
2. Create an attribute container to represent a data stream (EventDataStream). Have event data reference this attribute container.

For now 2 seems the most straight forward solution, with the downside that this is yet another attribute container to de-refference for psort.

joachimmetz:
* Add EventDataStream attribute container https://github.com/log2timeline/plaso/pull/2957
* Change storage to support EventDataStream attribute container
* Change parser mediator to add EventDataStream attribute container
* Have analyzers fill EventDataStream attributes

PR #2957, opened 2020-05-22
Added event data stream attribute container #2956

joachimmetz:
Added event data stream attribute container #2956


joachimmetz:
@Onager also see https://github.com/log2timeline/plaso/issues/2956 for context.

PR #2960, opened 2020-05-22
Changes to storage for event data stream attribute containers #2956

joachimmetz:
Changes to storage for event data stream attribute containers #2956

### DIFF
--- a/plaso/storage/file_interface.py
+++ b/plaso/storage/file_interface.py
@@ -308,6 +308,25 @@ class StorageFileReader(interface.StorageReader):
     """
     return self._storage_file.GetEventDataByIdentifier(identifier)
 
+  def GetEventDataStreams(self):
+    """Retrieves the event data streams.
+
+    Returns:
+      generator(EventDataStream): event data stream generator.
+    """
+    return self._storage_file.GetEventDataStreams()
+
+  def GetEventDataStreamByIdentifier(self, identifier):
+    """Retrieves a specific event data stream.
+
+    Args:
+      identifier (AttributeContainerIdentifier): event data identifier.
+
+    Returns:
+      EventDataStream: event data stream or None if not available.
+    """
+    return self._storage_file.GetEventDataStreamByIdentifier(identifier)
+
   def GetEvents(self):
     """Retrieves the events.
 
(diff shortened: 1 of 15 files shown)


plaso/storage/file_interface.py
```diff
@@ -308,6 +308,25 @@ def GetEventDataByIdentifier(self, identifier):
     """
     return self._storage_file.GetEventDataByIdentifier(identifier)
 
+  def GetEventDataStream(self):
+    """Retrieves the event data stream.
```
Onager, line 312: Change to "Retrieves the event data streams". Do this for all other implementations of this method as well.

plaso/storage/file_interface.py
```diff
@@ -677,6 +713,17 @@ def GetEventDataByIdentifier(self, identifier):
     """
     return self._storage_file.GetEventDataByIdentifier(identifier)
 
+  def GetEventDataStreamByIdentifier(self, identifier):
+    """Retrieves specific event data stream.
```
Onager, line 717: Change to "Retrieves a specific event data stream". Please change this for all other implementations of this method as well.

plaso/storage/file_interface.py
```diff
@@ -308,6 +308,25 @@ def GetEventDataByIdentifier(self, identifier):
     """
     return self._storage_file.GetEventDataByIdentifier(identifier)
 
+  def GetEventDataStream(self):
```
Onager, line 311: Rename to "GetEventDataStreams" to be consistent with other methods (GetWarnings, GetEvents, etc.). This will affect all subclasses, calls etc.

plaso/storage/sqlite/sqlite_file.py
```diff
@@ -646,6 +687,24 @@ def AddEvent(self, event, serialized_data=None):
...
+      event_data (EventData): event data.
+      serialized_data (Optional[bytes]): serialized form of the event data.
+
+    Raises:
+      IOError: when the storage file is closed or read-only.
+      OSError: when the storage file is closed or read-only.
+    """
+    self._RaiseIfNotWritable()
+
+    # The serialized data is not used, as this method modifies the attribute
```
Onager, line 703: Comment: this to me suggests that attaching the EventDataStream to the event vs. the EventData is a better way to approach this. Events are cheap to serialize and deserialize, as they have a relatively small number of fixed fields. EventData are typically much larger, and have a variable number of fields.

Having to reserialize EventData when merging to the session store negates much of the benefit of doing any serialized data caching at all.

plaso/storage/file_interface.py
joachimmetz: good catch

plaso/storage/file_interface.py
joachimmetz: done

plaso/storage/file_interface.py
joachimmetz: ack, fixed 2 instances of the missing `a`

plaso/storage/sqlite/sqlite_file.py
joachimmetz: Ack, let's chat more about this. There are more events than event data so the costs/benefits might even out. I opt to continue with this path, we can definitely look into gains of changing this at a later date. Especially if we're considering serializing to native SQLite types,

PR #2961, opened 2020-05-22
Changes to attribute containers to serialize protected attributes

joachimmetz:
Changes to attribute containers to serialize protected attributes

### DIFF
--- a/plaso/containers/interface.py
+++ b/plaso/containers/interface.py
@@ -35,10 +35,15 @@ class AttributeContainer(object):
   the container type, for example the container type "event" identifiers
   an event object.
 
-  Attributes are public class members of an serializable type. Protected
-  and private class members are not to be serialized.
+  Attributes are public class members of an serializable type. Protected and
+  private class members are not to be serialized, with the exception of those
+  defined in _SERIALIZABLE_PROTECTED_ATTRIBUTES.
   """
   CONTAINER_TYPE = None
+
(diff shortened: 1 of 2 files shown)


joachimmetz:
The context for these changes preparations to move storage internal only (attribute) container attributes like `event.event_data_row_identifier` to protected attribute names like `event._event_data_row_identifier`

plaso/containers/interface.py
```diff
@@ -111,19 +115,20 @@ def GetAttributeValuesString(self):
       str: comparable string of the attribute values.
     """
     attributes = []
-    for attribute_name, attribute_value in sorted(self.__dict__.items()):
+    for attribute_name, attribute_value in self.__dict__.items():
```
joachimmetz, line 118: Note to self, add call to sorted()

joachimmetz:
The rational is mainly to prevent these "special" attributes leaking into "regular/public" attribute space. To make the split between data (public attribute) and metadata (protected attributes) more clear. We can do without 

> My concern with these changes is that they add a lot of extra complexity to these methods, and the reason for doing so isn't clear to me.

The additional complexity is minimal, 1x additional checks for certain defined protected attribute names in a 3x places; That's it. The storage needs to treat them the same if they are public or protected.

> Given this seems to be part of a large set of changes to storage, maybe its worth putting together a short doc with where this is all headed.

I think a doc might be overkill here, I opt to chat about it and see if I can make things more clear for you.

PR #2965, opened 2020-05-23
Changes to enforce SQLite event data row identifier

joachimmetz:
Changes to enforce SQLite event data row identifier


joachimmetz:
Note to self: this will break backward compatibility with older storage files that don't use event data. Determine which format version event data was introduced and set that as the minimum read requirement.

joachimmetz:
Looks like event data was introduced in the SQlite storage file in https://github.com/log2timeline/plaso/commit/540bd2296e2aaf34d3ae0b0645cfb447c6436310 hence using 20170707 as the minimum read version.

PR #2968, opened 2020-05-23
Changes to parser mediator to support EventDataStream #2956

joachimmetz:
Changes to parser mediator to support EventDataStream #2956

### DIFF
--- a/plaso/engine/worker.py
+++ b/plaso/engine/worker.py
@@ -17,6 +17,7 @@ from dfvfs.resolver import resolver as path_spec_resolver
 from plaso.analyzers import hashing_analyzer
 from plaso.analyzers import manager as analyzers_manager
 from plaso.containers import event_sources
+from plaso.containers import events
 from plaso.engine import extractors
 from plaso.engine import logger
 from plaso.lib import definitions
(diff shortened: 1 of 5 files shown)


plaso/engine/worker.py
```diff
@@ -218,22 +222,21 @@ def _AnalyzeFileObject(self, mediator, file_object):
 
       data = file_object.read(maximum_read_size)
 
-    display_name = mediator.GetDisplayName()
+    analyzer_attributes = {}
```
Onager, line 225: Remove this and directly set attributes on event_data_stream instead.

plaso/engine/worker.py
```diff
@@ -122,25 +123,27 @@ def __init__(self, parser_filter_expression=None):
...
+    attributes that are added to produced event objects. Note that some
+    file systems allow directories to have data streams, such as NTFS.
 
     Args:
-      mediator (ParserMediator): mediates the interactions between
-          parsers and other components, such as storage and abort signals.
       file_entry (dfvfs.FileEntry): file entry whose data stream is to be
           analyzed.
       data_stream_name (str): name of the data stream.
+      display_name (str): human readable of the file entry currently being
```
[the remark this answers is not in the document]
joachimmetz: done

plaso/engine/worker.py
joachimmetz: I have this as an after care point of https://github.com/log2timeline/plaso/issues/2956. I opt to do this in a different PR, to not change everything in a single PR.

joachimmetz:
@Onager PR has been rebased and responded/addressed comments, PTAL

plaso/engine/worker.py
Onager: I think your aftercare point is a bit different, I meant changing what's currently on like 241 to something like `setattr(event_data_stream, result.attribute_name, result.attribute_value)`. The aftercare point that I could identify from #2956 that I think you're talking about is `To do - determine if analyzers should set EventDataStream attributes directly?` which is a more involved change.
I'm just asking to remove the extra loop in this method.

joachimmetz: ah, yeah that is different than what I interpreted, I'll change it

joachimmetz: done

joachimmetz:
@Onager PTAL

PR #2973, opened 2020-05-24
Changes to DSV output modules to use shared base class

joachimmetz:
Changes to DSV output modules to use shared base class

### DIFF
--- a/plaso/output/dynamic.py
+++ b/plaso/output/dynamic.py
@@ -1,17 +1,18 @@
 # -*- coding: utf-8 -*-
-"""Contains a formatter for a dynamic output module for plaso."""
+"""Dynamic selected delimiter separated values output module."""
 
 from __future__ import unicode_literals
 
 from plaso.lib import errors
 from plaso.lib import timelib
-from plaso.output import interface
+from plaso.output import formatting_helper
 from plaso.output import logger
 from plaso.output import manager
+from plaso.output import shared_dsv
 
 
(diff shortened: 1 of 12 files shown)


plaso/output/dynamic.py
```diff
@@ -1,17 +1,18 @@
 # -*- coding: utf-8 -*-
-"""Contains a formatter for a dynamic output module for plaso."""
+"""Dynamic selected delimeter separated values output module."""
```
Onager, line 2: Delimiter, not delimeter

plaso/output/formatting_helper.py
```diff
@@ -0,0 +1,170 @@
...
+
+from __future__ import unicode_literals
+
+from plaso.lib import errors
+
+
+class FieldFormattingHelper(object):
+  """Output module field formatting helper."""
+
+  # Maps the name of a fields to a a callback function that formats
```
Onager, line 12: Please rewrite this comment to something like: "Maps the name of a field to callback function that formats the field value"

joachimmetz:
Rebased and addressed merge conflicts.

plaso/output/dynamic.py
joachimmetz: done

plaso/output/formatting_helper.py
joachimmetz: done

joachimmetz:
Addressed review comments.

Issue #2985, opened 2020-05-28
New bencode.py release breaks tests


joachimmetz:
There appears to be a new bencode.py release, that currently breaks the tests

https://travis-ci.com/github/log2timeline/plaso/jobs/340809463#L4127

```
======================================================================
FAIL: testProcess (parsers.bencode_plugins.transmission.BencodeTest)
Tests the Process function.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/plaso/tests/parsers/bencode_plugins/transmission.py", line 24, in testProcess
    self.assertEqual(storage_writer.number_of_events, 3)
AssertionError: 0 != 3
```

joachimmetz:
Stop gap measure to prevent the test from failing https://github.com/log2timeline/plaso/pull/2986

joachimmetz:
Looks the keys in the OrderedDict of the decoded values are bytes in bencode 3 where previously they were strings.

PR #2989, opened 2020-05-28
Changes to dependencies to use bencode.py 2.x #2985

joachimmetz:
Changes to dependencies to use bencode.py 2.x #2985


joachimmetz:
@Onager for context these are changes generated with https://github.com/log2timeline/l2tdevtools/pull/835

PR #3018, opened 2020-06-06
Changed end-to-end tests to compare log2timeline with psort and psteal output

joachimmetz:
Changed end-to-end tests to compare log2timeline with psort and psteal output


config/end-to-end.ini
```diff
@@ -11,22 +11,24 @@ case=extract_and_output
 extract_options=--no-vss
 output_file=extract_and_output1.log
```
Onager, line 12: Please change this to match the reference_output_file name

config/end-to-end.ini
```diff
@@ -11,22 +11,24 @@ case=extract_and_output
...
 output_format=dynamic
-reference_output_file=test_data/end_to_end/extract_and_output1.log
+reference_output_file=test_data/end_to_end/extract_and_output.log
 source=test_data/vsstest.qcow2
 
 [extract_and_output2]
 case=extract_and_output
 extract_options=--no-vss --text C:
 output_file=extract_and_output2.log
```
Onager, line 20: Please change this to match the reference_output_file name

config/end-to-end.ini
```diff
@@ -11,22 +11,24 @@ case=extract_and_output
...
 output_format=dynamic
-reference_output_file=test_data/end_to_end/extract_and_output2.log
+reference_output_file=test_data/end_to_end/extract_and_output_with_text_prepend.log
 source=test_data/vsstest.qcow2
 
 [extract_and_output3]
 case=extract_and_output_with_psteal
+extract_options=--no-vss
 output_file=extract_and_output3.log
```
Onager, line 28: Please change this to match the reference_output_file name

config/end-to-end.ini
joachimmetz: No, since that would cause the `extract_and_output3` test to overwrite this file

`test_data/end_to_end/extract_and_output.log` is used for both the `extract_and_output1` (psort) and `extract_and_output3` (psteal) tests. Hence the description.

config/end-to-end.ini
joachimmetz: It is actually easier to have the file named corresponding to the test, so it is clear in the test result directory that this file belongs to the extract_and_output2 test

config/end-to-end.ini
joachimmetz: IMHO no, since that would cause the extract_and_output3 test to overwrite this file in the test results directory

test_data/end_to_end/extract_and_output.log is used for both the extract_and_output1 (psort) and extract_and_output3 (psteal) tests. Hence the description.

joachimmetz:
@Onager provided a bit of context on why the requested changes IMHO are not a good idea

PR #3021, opened 2020-06-06
Removed attributes from event data that can be derived from path spec

joachimmetz:
Removed attributes from event data that can be derived from path spec


joachimmetz:
Rebased and addressed merge issues.

PR #3053, opened 2020-06-09
Additional changes for event counter tests to check data types

joachimmetz:
Additional changes for event counter tests to check data types


joachimmetz:
@Onager looks like one test update fell between merges, this PR contains a correction for that

PR #3064, opened 2020-06-11
Removed dependency check for cffi

joachimmetz:
Removed dependency check for cffi


joachimmetz:
Merged this since this is currently affecting Docker CI tests.

Issue #2067, opened 2018-08-02
psort: l2tcsv output module ignores the timezone setting


yukinobu-n:
**Description of problem:**

In plaso-20180630, it seems `l2tcsv` output module of `psort` ignores the timezone setting which is specified by `-z` option.

In case of using default output module, `psort` treats the timezone setting correctly.

From the following test, it is considered that this issue was occurred after plaso-20180127.

**Command line and arguments:**

**Test 1 (l2tcsv):** With plaso-20180630, I ran the following command including `-o l2tcsv` option. 

```
log2timeline --parsers winevt,winevtx ./winevt.db /path/to/eventlogs && psort -z Japan -w eventlog_timeline.csv -o l2tcsv ./winevt.db
```

Then I saw the first time on the timeline is `07/19/2018,06:27:51,Japan`. But the correct time is 15:27:51 (JST). 06:27:51 seems like UTC time. `-z Japan` option seems to be ignored.

**Test 2 (default):** With plaso-20180630, after I removed db and timeline, I ran the following command which is not including any `-o ` option. 

```
log2timeline --parsers winevt,winevtx ./winevt.db /path/to/eventlogs && psort -z Japan -w eventlog_timeline.csv ./winevt.db
```

Then I saw the first time on the timeline is `2018-07-19T15:27:51.267307+09:00`. This is the correct time.

**Test 3 (20180127):** With plaso-20180127, after I removed db and timeline, I ran the following command that is exactly same as Test 1.

```
log2timeline --parsers winevt,winevtx ./winevt.db /path/to/eventlogs && psort -z Japan -w eventlog_timeline.csv -o l2tcsv ./winevt.db
```

Then I saw the first time on the timeline is `07/19/2018,15:27:51,Japan`. This is the correct time.

**Source data:**

Eventlogs: [eventlogs.zip](https://github.com/log2timeline/plaso/files/2252228/eventlogs.zip)

**Plaso version:**

plaso-20180630-amd64

**Operating system Plaso is running on:**

Windows 7 SP1 (64bit)

**Installation method:**

I downloaded plaso-20180630-amd64.zip (sha1: `0e10b5a99c99bdd04fd8a5dcbc7e4d96f9a982c8`) and I unzipped it to my local folder.

**Debug output/tracebacks:**

Debug output in Test 1: [debug_out.zip](https://github.com/log2timeline/plaso/files/2252227/debug_out.zip)

Onager:
[...]

Issue #910, opened 2016-08-11
Migrate plaso timestamp to dfdatetime


joachimmetz:
- [x] ~~Migrate plaso timestamp to [dfdatetime](https://github.com/log2timeline/dfdatetime)~~
  - ~~https://codereview.appspot.com/315300043/~~
  - ~~https://codereview.appspot.com/318190043/~~
  - ~~https://codereview.appspot.com/319930043/~~
  - ~~https://codereview.appspot.com/304440043/~~
- [x] ~~deprecate TextEvent~~
- [x] ~~Moved definitions from eventdata.py to definitions.py and removed eventdata.py~~
- [x] ~~deprecate use of ShellItemFileEntryEvent~~
  - ~~https://codereview.appspot.com/317680043~~
- [x] ~~deprecate PlistEvent~~
  - ~~https://codereview.appspot.com/323050043/~~
- [x] ~~replace FromTimeParts~~
  - ~~sccm add TimeElementsInMicroseconds to dfdatetime~~
  - ~~symantec~~
  - ~~syslog~~
- [x] ~~Removed FromPythonDatetime and FromPosixTime~~
  - ~~https://codereview.appspot.com/347740043/~~
- [x] ~~replace use of timelib in file entry filter~~
  - ~~https://codereview.appspot.com/341160043/~~
  - ~~https://codereview.appspot.com/348740043~~
- [x] ~~replace CopyToDatetime in output modules~~
  - ~~https://codereview.appspot.com/342080043/~~
  - ~~updated l2tbinaries and GIFT PPA~~
- [x] ~~Remove timelib CopyToPosix function in output modules~~
  - ~~https://codereview.appspot.com/343150043/~~
  - ~~updated l2tbinaries and GIFT PPA~~
  - ~~https://codereview.appspot.com/343890043/~~
- [x] ~~replace GetNow https://github.com/log2timeline/plaso/pull/3074~~
- [x] ~~move CopyFromString into tests~~
  - ~~https://github.com/log2timeline/plaso/pull/3080~~
  - ~~https://github.com/log2timeline/plaso/pull/3081~~
- [x] ~~replace PythonDatetimeEvent by dfdatetime.TimeElements.CopyFromDatetime~~
  - ~~https://github.com/log2timeline/dfdatetime/issues/104~~
  - ~~https://github.com/log2timeline/dfdatetime/pull/164~~
  - ~~https://github.com/log2timeline/plaso/pull/3082~~
[...]

PR #3071, opened 2020-06-12
Changes for dynamic and l2tcsv output time zone adjustment #2067

joachimmetz:
Changes for dynamic and l2tcsv output time zone adjustment #2067

Note to self:

* update https://github.com/log2timeline/plaso/pull/3088 after merge


joachimmetz:
@rgayon that is correct these change address the issue described #2067 but the path outlined in #910 will eventually get back replacing this with a dfDateTime aligned solution.

PR #3073, opened 2020-06-12
Removed timelib RoundToSeconds #910

joachimmetz:
Removed timelib RoundToSeconds #910

### DIFF
--- a/plaso/output/shared_elastic.py
+++ b/plaso/output/shared_elastic.py
@@ -3,9 +3,10 @@
 
 from __future__ import unicode_literals
 
-import os
 import logging
+import os
 
+from dfdatetime import posix_time as dfdatetime_posix_time
 from dfvfs.serializer.json_serializer import JsonPathSpecSerializer
 
 try:
@@ -14,9 +15,9 @@ except ImportError:
   elasticsearch = None
 
 from plaso.lib import errors
-from plaso.lib import timelib
 from plaso.output import interface
 from plaso.output import logger
+
 
 # Configure the Elasticsearch logger.
 if elasticsearch:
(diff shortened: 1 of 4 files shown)


plaso/output/shared_elastic.py
```diff
@@ -170,17 +171,19 @@ def _GetSanitizedEventValues(self, event, event_data, event_tag):
...
-    except TypeError as exception:
-      logger.warning((
-          'Unable to round timestamp {0!s}. error: {1!s}. '
-          'Defaulting to 0').format(event.timestamp, exception))
-      attribute_value = 0
-
-    attribute_value = timelib.Timestamp.CopyToIsoFormat(
-        attribute_value, timezone=self._output_mediator.timezone)
-    event_values['datetime'] = attribute_value
+    # TODO: determine why the string representation needs to be rounded in
```
joachimmetz, line 174: @berggren @kiddinn @Onager any idea why the string representation needs to be in seconds precision?

plaso/output/shared_elastic.py
```diff
@@ -170,17 +171,19 @@ def _GetSanitizedEventValues(self, event, event_data, event_tag):
...
+    # seconds.
+    number_of_seconds, fraction_of_second = divmod(
+        event.timestamp, definitions.MICROSECONDS_PER_SECOND)
+
+    if fraction_of_second > 500000:
```
rgayon, line 179: Since we have a macro for MICROSECONDS_PER_SECOND, should 500000 be replaced by definitions.MICROSECONDS_PER_SECOND/2  .

joachimmetz: Ack, I'll change it to a class constant `_MICROSECONDS_PER_HALF_SECOND`

plaso/output/shared_elastic.py
joachimmetz: This was introduced in https://github.com/log2timeline/plaso/commit/4d7b40ba21189d0f13c55435cf5e0a0e1a08bae0

joachimmetz:
Note to self, RoundToSeconds can likely be remove entirely and return dfDateTime's ISO 8601 string instead.

joachimmetz:
Per conversation @kiddinn let's remove round to seconds

plaso/output/shared_elastic.py
joachimmetz: Discussed with @kiddinn and @berggren there is no good reason to keep this behavior, removing round to seconds behavior entirely.

plaso/output/shared_elastic.py
joachimmetz: code was removed instead

rgayon:
Approved, but it looks like tests need a different formatting for the timezone part

joachimmetz:
> Trying to unapprove

Ack, having a look at the tests. Unfortunately something has an outage at the moment

PR #3074, opened 2020-06-13
Moved timelib GetNow and GetCurrentYear into mediators #910

joachimmetz:
Moved timelib GetNow and GetCurrentYear into mediators #910


joachimmetz:
> I suppose GetNow() wasn't used anywhere else in the code

Correct

PR #3081, opened 2020-06-13
Moved timelib CopyFromString into tests #910

joachimmetz:
Moved timelib CopyFromString into tests #910


tests/lib/timelib.py
```diff
@@ -11,120 +11,18 @@
...
-class TimeLibTest(unittest.TestCase):
-  """Tests for timestamp."""
-
-  def testCopyFromString(self):
-    """Tests the CopyFromString function."""
-    timestamp = timelib.Timestamp.CopyFromString('2012-06-27')
-    expected_timestamp = 1340755200000000
-    self.assertEqual(timestamp, expected_timestamp)
-
-    with self.assertRaises(ValueError):
```
rgayon, line 24: Don't we still want to test that bad timestamps will raise expectedly?

joachimmetz: Not sure what you mean but the CopyFromString method is removed here.

rgayon: oh nevermind, I expect these tests are now on the dfdatetime side, which is why I don't see them in this PR

Issue #2617, opened 2019-06-04
CLI tools add unattended mode


joachimmetz:
To run plaso CLI tools e.g. from turbina, add an unattended mode.

behavior of unattended mode is to fail if input from user is required.

taishi8117:
When `--credential` is used with an incorrect credential, Plaso currently hangs and waits for the correct one from stdin, which makes it hard to integrate with a tool like turbinia.

joachimmetz:
Please read https://en.wikipedia.org/wiki/Hang_(computing) for the meaning of a hang.

joachimmetz:
Related issue: https://github.com/log2timeline/plaso/issues/968

PR #3094, opened 2020-06-16
Added unattended mode for non-interactive usage of tools #2617

joachimmetz:
Added unattended mode for non-interactive usage of tools #2617

### DIFF
--- a/plaso/cli/storage_media_tool.py
+++ b/plaso/cli/storage_media_tool.py
@@ -191,6 +191,10 @@ class StorageMediaTool(tools.CLITool):
         return selected_volume_identifiers
 
     if len(volume_identifiers) > 1:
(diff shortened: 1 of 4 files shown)


config/docker/plaso-switch.sh
```diff
@@ -10,23 +10,23 @@
 #
 # or to run it on actual data:
 # mkdir -p /data/sources    # put the files to parse here
-# mkdir -p /data/results    # plaso storage file will appear hear
+# mkdir -p /data/results    # a Plaso storage file will appear hear
```
rgayon, line 13: can you fix "hear" => "here"?

joachimmetz: thx, good catch

Issue #2894, opened 2020-04-21
improve documentation of -z option of tools


joachimmetz:
Different plaso tools have a `-z` time zone option. However they have different meanings, improve the help description to include what time zone is affected, that of the input or output

joachimmetz:
split original `-z` into:
* `-z` preferred time zone for extraction
* `--output-time-zone` for output format/modules that support it

PR #3100, opened 2020-06-17
Updated output time zone in the psort documentation #2894

joachimmetz:
Updated output time zone in the psort documentation #2894

### DIFF
--- a/docs/sources/user/Using-psort.md
+++ b/docs/sources/user/Using-psort.md
@@ -11,8 +11,8 @@ Looking for [tips and tricks](Using-psort.md#how-do-i)?
 To see a list of all available parameters you can pass to psort use ``-h`` or
 ``--help``.
 
-Basic usage of **psort** is straightforward, just provide it with a storage file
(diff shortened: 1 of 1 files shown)


docs/sources/user/Using-psort.md
```diff
@@ -1,30 +1,39 @@
...
-The simplest way to run the tool is simply provide it with a storage file.
+The most basic way to use **psort** is to provide it with a storage file and
+output file, for example:
 
 ```
 $ psort.py -w test.log test.plaso
 ```
 
-This will use the default output module and print out to STDOUT a list of all extracted events, merging detected duplicate events. All timestamps on the output will be in UTC.
+**Note that as of 1.5.0 psort no longer supports wrting output to stdout.**
```
pyllyukko, line 21: ☝️ Typo

joachimmetz: thx for flagging, note that it helps to point out which word is incorrectly written.

PR #3102, opened 2020-06-18
Changed end-to-end script to hash files exported files

joachimmetz:
Changed end-to-end script to hash files exported files

### DIFF
--- a/tests/end-to-end.py
+++ b/tests/end-to-end.py
@@ -10,6 +10,7 @@ import abc
 import argparse
 import configparser
 import difflib
+import hashlib
 import logging
 import os
 import shutil
@@ -463,7 +464,7 @@ class StorageFileTestCase(TestCase):
       temp_directory (str): name of a temporary directory.
 
     Returns:
-      bool: True if he output files are identical.
+      bool: True if the output files are identical.
     """
     output_file_path = os.path.join(temp_directory, test_definition.output_file)
 
@@ -496,6 +497,7 @@ class StorageFileTestCase(TestCase):
             line = line.replace('C:\\tmp\\test\\test_data\\', '')
             line = line.replace('C:\\\\tmp\\\\test\\\\test_data\\\\', '')
             output_list.append(line)
+
           differences = list(difflib.unified_diff(
               reference_output_list, output_list,
               fromfile=reference_output_file_path, tofile=output_file_path))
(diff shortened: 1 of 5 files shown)


tests/end-to-end.py
```diff
@@ -1147,6 +1151,89 @@ def __init__(
...
+            self._test_references_path, reference_hashes_file_path)
+
+      if not os.path.exists(reference_hashes_file_path):
+        logging.error('No such reference hashes file: {0:s}'.format(
+            reference_hashes_file_path))
+        return False
+
+      with open(reference_hashes_file_path, 'r') as reference_hashes_file:
+        with open(hashes_file_path, 'r') as hashes_file:
+          # Hack to remove paths in the hashes that are different when running
```
rgayon, line 1180: It looks like on windows the path & hashes are different than what is expected.
SHould we have different reference hash files per platform?

tests/end-to-end.py
```diff
@@ -1147,6 +1151,89 @@ def __init__(
...
+            data = file_object.read(self._READ_BUFFER_SIZE)
+            while data:
+              hash_context.update(data)
+              data = file_object.read(self._READ_BUFFER_SIZE)
+
+          relative_file_path = file_path[len(temp_directory):]
+          hashes_file_object.write('{0:s}\t{1:s}\n'.format(
+              hash_context.hexdigest(), relative_file_path))
+
+    if os.path.exists(hashes_file_path):
```
rgayon, line 1234: should we raise if hashes_file_path doesn't exist, or do we expect _CreateHashesFile to be called with no an empty {temp_directory}/export?

tests/end-to-end.py
joachimmetz: hashes should be the same, paths are different. I currently reformat the paths.

tests/end-to-end.py
joachimmetz: IMHO no need to raise here, if the path is missing this will fail the compare step.

tests/end-to-end.py
```diff
@@ -1159,13 +1151,92 @@ def __init__(
...
 
+  def _CompareHashesFile(self, test_definition, temp_directory):
+    """Compares the hashes file with a reference hashes file.
+
+    Args:
+      test_definition (TestDefinition): test definition.
+      temp_directory (str): name of a temporary directory.
+
+    Returns:
+      bool: True if he hashes files are identical.
```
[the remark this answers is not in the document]
joachimmetz: @rgayon what do you mean with this?

joachimmetz:
Note to self have a look why the Windows tests are failing.

tests/end-to-end.py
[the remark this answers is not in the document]
joachimmetz: nvm figured it out

tests/end-to-end.py
joachimmetz: I see what you mean, I've changed the code a bit.

PR #3104, opened 2020-06-18
Have docs Makefile use Python build directory

joachimmetz:
Have docs Makefile use Python build directory


joachimmetz:
Dropping this in favor or https://github.com/log2timeline/plaso/issues/3111

Issue #3145, opened 2020-07-02
psort-studentpc1-services end-to-end test failing with QueueFull


joachimmetz:
a previous instance of this issue https://github.com/log2timeline/plaso/issues/3001

```
2020-06-28 00:54:56,187 [INFO] (MainProcess) PID:4328 <data_location> Determined data location: /tmp/workspace/psort-linux-services/data
Traceback (most recent call last):
  File "./tools/psort.py", line 99, in <module>
    if not Main():
  File "./tools/psort.py", line 76, in Main
    tool.ProcessStorage()
  File "/tmp/workspace/psort-linux-services/plaso/cli/psort_tool.py", line 556, in ProcessStorage
    worker_memory_limit=self._worker_memory_limit)
  File "/tmp/workspace/psort-linux-services/plaso/multi_processing/psort.py", line 947, in AnalyzeEvents
    storage_writer, analysis_plugins, event_filter=event_filter)
  File "/tmp/workspace/psort-linux-services/plaso/multi_processing/psort.py", line 308, in _AnalyzeEvents
    event_queue.PushItem((event, event_data, event_data_stream))
  File "/tmp/workspace/psort-linux-services/plaso/engine/zeromq_queue.py", line 457, in PushItem
    raise errors.QueueFull
plaso.lib.errors.QueueFull
```

* [x] fix issue https://github.com/log2timeline/plaso/pull/3146
* [x] add end-to-end test that can be run on CI https://github.com/log2timeline/plaso/pull/3148
  * ~~reduce analysis module (worker process) time out, default is 15 mins~~

joachimmetz:
```
Traceback (most recent call last):
  File "plaso/multi_processing/analysis_process.py", line 246, in _ProcessEvent
    self._analysis_plugin.ExamineEvent(
  File "plaso/analysis/windows_services.py", line 281, in ExamineEvent
    service = WindowsService.FromEventData(event_data, event_data_stream)
  File "plaso/analysis/windows_services.py", line 137, in FromEventData
    event_data.start_type, event_data.object_name, source,
AttributeError: 'EventData' object has no attribute 'object_name'
```

PR #3147, opened 2020-07-04
Changes to set worker timeout from command line #3145

joachimmetz:
Changes to set worker timeout from command line #3145

### DIFF
--- a/plaso/cli/log2timeline_tool.py
+++ b/plaso/cli/log2timeline_tool.py
@@ -99,7 +99,6 @@ class Log2TimelineTool(extraction_tool.ExtractionTool):
     self._status_view_mode = status_view.StatusView.MODE_WINDOW
     self._stdout_output_writer = isinstance(
         self._output_writer, tools.StdoutOutputWriter)
-    self._worker_memory_limit = None
 
     self.dependencies_check = True
     self.list_hashers = False
@@ -401,7 +400,10 @@ class Log2TimelineTool(extraction_tool.ExtractionTool):
     if single_process_mode:
       extraction_engine = single_process_engine.SingleProcessEngine()
     else:
-      extraction_engine = multi_process_engine.TaskMultiProcessEngine()
+      extraction_engine = multi_process_engine.TaskMultiProcessEngine(
+          number_of_worker_processes=self._number_of_extraction_workers,
+          worker_memory_limit=self._worker_memory_limit,
+          worker_timeout=self._worker_timeout)
 
     # If the source is a directory or a storage media image
     # run pre-processing.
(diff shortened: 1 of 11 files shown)


plaso/cli/log2timeline_tool.py
```diff
@@ -401,7 +400,10 @@ def ExtractEventsFromSources(self):
     if single_process_mode:
       extraction_engine = single_process_engine.SingleProcessEngine()
     else:
-      extraction_engine = multi_process_engine.TaskMultiProcessEngine()
+      extraction_engine = multi_process_engine.TaskMultiProcessEngine(
```
Onager, line 403: Rather than adding args to the engine constructor, please add the worker_* values to the [processing_configuration](https://github.com/log2timeline/plaso/blob/35082c09e525fac88daa2b522d9d546f7d051184/plaso/engine/configurations.py#L172) object that gets passed to ProcessSources().

This is where all the other processing-specific args are, with apparently a couple of exceptions that should also move. Having two different paths for configuring an engine to process something adds extra complexity, and I don't see any advantage for having these different paths.

plaso/cli/psort_tool.py
```diff
@@ -239,9 +240,16 @@ def _ParseProcessingOptions(self, options):
 
     if worker_memory_limit and worker_memory_limit < 0:
       raise errors.BadConfigOption(
-          'Invalid worker memory limit value cannot be negative.')
+          'Invalid worker memory limit value cannot be less than 0.')
```
Onager, line 243: Please include the invalid value in the error message here. For example: "Invalid memory limit -1. Limit cannot be less than 0."

plaso/cli/psort_tool.py
```diff
@@ -239,9 +240,16 @@ def _ParseProcessingOptions(self, options):
...
+
+    worker_timeout = getattr(options, 'worker_timeout', None)
+
+    if worker_timeout is not None and worker_timeout <= 0.0:
+      raise errors.BadConfigOption(
+          'Invalid worker timeout value must be larger than 0.0 minutes.')
```
Onager, line 249: Please include the invalid value in the error message here: "Invalid memory limit -1. Limit cannot be less than 0."

plaso/lib/definitions.py
```diff
@@ -14,8 +14,13 @@
     COMPRESSION_FORMAT_NONE,
     COMPRESSION_FORMAT_ZLIB])
 
+# Default worker process memory limit of 2 GiB.
 DEFAULT_WORKER_MEMORY_LIMIT = 2048 * 1024 * 1024
 
+# Consider a worker process inactive after 15 minutes of no activity
```
Onager, line 20: Change this to just "of no status updates".

plaso/cli/log2timeline_tool.py
joachimmetz: The reasoning behind moving this to the constructor was that these values don't change once set and are engine specific.

plaso/cli/psort_tool.py
joachimmetz: done

plaso/cli/psort_tool.py
joachimmetz: done

plaso/lib/definitions.py
joachimmetz: done

joachimmetz:
@Onager PTAL, the reason for not having the number of workers and time in the config is that they don't change once set. Also these values are for the foreman/main process only and don't need to be communicated to the workers.

Onager:
I think these values should still be in the processing configuration for consistency, but we can address it in a follow up PR.

PR #3148, opened 2020-07-04
Added extract and analyze end-to-end test

joachimmetz:
Added extract and analyze end-to-end test


joachimmetz:
@Onager could you rubberstamp the PR as well, thx

PR #3171, opened 2020-07-26
Changes to user installation documentation

joachimmetz:
Changes to user installation documentation


joachimmetz:
Another user not knowing that Docker also works on Windows.

Issue #3177, opened 2020-07-29
If hashers is none, path_spec/display name are absent from result


william-billaud:
**Description of problem:**

If a disk is analyzed with --hashers none, path_spec/display_name is absent (or corrupted idk)

**Command line and arguments:**

```
log2timeline.py --partitions 2  --no-vss -f ./20_prefetch.yaml --hashers none  ./pf_no_hash.plaso  /data/disk.raw
psort -o xlsx -w pf_no_hash.xslx ./pf_no_hash.plaso 
```

```
log2timeline.py --partitions 2  --no-vss -f ./20_prefetch.yaml  ./pf.plaso  /data/disk.raw
psort -o xlsx -w pf.xslx ./pf.plaso 
```
where 20_prefetch.yaml is 

```
description: Prefetch files.
type: include
path_separator: '\'
paths:
- '%SystemRoot%\\Prefetch\\.+[.]pf'
```

in pf_no_hash.xlsx , event other than filestat won't have the field display name, eg : 

datetime | timestamp_desc | source | source_long | message | parser | display_name | tag
-- | -- | -- | -- | -- | -- | -- | --
2016-07-14 13:32:57,614 | Previous Last Time Executed | LOG | WinPrefetch | Prefetch [TIWORKER.EXE] was executed - run count 5 path hints: \WINDOWS\WINSXS\X86_MICROSOFT-WINDOWS-SERVICINGSTACK_31BF3856AD364E35_10.0.10586.0_NONE_39C65E1DB9443C5F\TIWORKER.EXE hash: 0x45CD89AD volume: 1 [serial number: 0xB6E74FF7, device path: \VOLUME{01d1dde0e6d6158b-b6e74ff7}] | prefetch | - | -
-- | -- | -- | -- | -- | -- | -- | --

If the output is elasticsearch, path_spec will be missing.
In pf.xlsx evrything will be normal

**Source data:**

https://www.enisa.europa.eu/ftp/ENISA-LOT3-Evidence.zip
(disk.raw is inside the vmdk

**Plaso version:**

20200717

**Operating system Plaso is running on:**

Ubuntu

**Installation method:**

For example:
* installed from [GiFT PPA][https://launchpad.net/~gift] stable track

joachimmetz:
thx for the report this looks like an issue with path spec not being set if there are no analyzers.

joachimmetz:
[...]

PR #3179, opened 2020-07-29
Added end-to-end test with analyzers disabled #3177

joachimmetz:
Added end-to-end test with analyzers disabled #3177

### DIFF
--- a/config/end-to-end.ini
+++ b/config/end-to-end.ini
@@ -37,6 +37,14 @@ output_format=dynamic
 reference_output_file=test_data/end_to_end/extract_and_output.log
 source=test_data/vsstest.qcow2
 
+[extract_and_output_no_hashers1]
+case=extract_and_output
+extract_options=--no-vss --hashers=none
+output_file=extract_and_output_no_hashers1.log
+output_format=dynamic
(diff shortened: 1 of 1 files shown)


config/end-to-end.ini
```diff
@@ -37,6 +37,14 @@ output_format=dynamic
 reference_output_file=test_data/end_to_end/extract_and_output.log
 source=test_data/vsstest.qcow2
 
+[extract_and_output_no_hahsers1]
```
rgayon, line 40: s/hs/sh/

config/end-to-end.ini
```diff
@@ -37,6 +37,14 @@ output_format=dynamic
...
+case=extract_and_output
+extract_options=--no-vss --hashers=none
+output_file=extract_and_output_no_hahsers1.log
```
rgayon, line 43: same

config/end-to-end.ini
joachimmetz: ack, good catch

config/end-to-end.ini
joachimmetz: done

joachimmetz:
@rgayon feedback addressed, PTAL

PR #3184, opened 2020-08-04
Fix formatting error in Using-psort.md

stephenkreusch:

Fix documentation formatting in Using-psort.md

## Description:

https://plaso.readthedocs.io/en/latest/sources/user/Using-psort.html doesn't render properly at "Here the filter cached_file_size is 43".


joachimmetz:
@stephenkreusch thx for the correction

PR #3194, opened 2020-08-13
Added default mappings for Elasticsearch output modules

### LINKED ISSUE #3170: Elasticsearch could fail to ingest output from esedb/srum
**Description of problem:**

When uploading data to elasticsearch/timesketch via psort.py, elastic could fail to ingest most of esedb/srum event ( and also windows:registry:winlogon event).
This case append when the first srum event send to elastic have an application en user_identifier that were not resolved against the  SruDbIdMapTable.
In this case elasticsearch map internally the user_identifier and application field as long integer and failed to ingest data where these field are text.

**Command line and arguments:**

-

**Source data:**

could not provide

**Plaso version:**

For example 20200717

**Operating system Plaso is running on:**

Ubuntu

**Installation method:**

* installed from [GiFT PPA][https://launchpad.net/~gift] stable track

This issue could be resolved by two different way (I have tested both) :
- forcing the mapping of these to field when the index is created
`
{
            "properties": {
                "application": {
                    "type": "text",
                    "fields": {"keyword": {"type": "keyword"}},
                },
                "user_identifier": {
                    "type": "text",
                    "fields": {"keyword": {"type": "keyword"}},
                },
            }
        }
`
- just cast these value to string when they are created : ( replace line 333 of https://github.com/log2timeline/plaso/blob/master/plaso/parsers/esedb_plugins/srum.py : 
`record_value = identifier_mappings.get(record_value, record_value)` -> `record_value = str(identifier_mappings.get(record_value, record_value))`

Imo the second option is the best but I let you see which one you prefer

william-billaud:
Create specific mapping for Elasticsearch to prevent type collision that generate the lost of some events

Create a mapping for elasticsearch to prevent data lost when data are indexed.

## Description:

As explained in issue #3170 , when elastic ingest data, it create a mapping that assign type to field.
Some data_type have field with the same name but of different type.
In this case elasticsearch could silently fail to ingest data, causing the lose of some event.

The most frequent type collision is when a field could be an int or a string (eg: file_reference is an int in the fs:ntfs:usn_change, but a string in windows:shell_item:file_entry). In this case if an integer is ingested first, elastic will fail to ingest event where the field is a string (impossible to case).

By forcing the mapping we prevent elastic to assign these fields as integer, so it can ingest both strings in integers.

The main difference between a field marked as an integer or as a string is that you can make some specific aggregation (eg: average, sum etc...), but imo these aggregation are not important for these fields.

  

**Related issue (if applicable):** fixes #3170

## Notes:
review](https://github.com/log2timeline/l2tdocs/blob/master/process/Code%20review%20process.asciidoc). This makes sure

## Checklist:
  - [X] No new [new dependencies](https://plaso.readthedocs.io/en/latest/sources/developer/Adding-a-new-dependency.html) are required or l2tdevtools has been updated
  - [ ] Reviewer assigned


joachimmetz:
@kiddinn friendly ping

plaso/output/elastic.py
```diff
@@ -13,7 +13,70 @@ class ElasticsearchOutputModule(shared_elastic.SharedElasticsearchOutputModule):
 
   NAME = 'elastic'
   DESCRIPTION = 'Saves the events into an Elasticsearch database.'
-
+  DEFAULT_MAPPING = {
+      'properties': {
```
kiddinn, line 17: can you make sure that all quotes are single (') instead of double ("), to be consistent with the rest of the code base

plaso/output/elastic.py
```diff
@@ -13,7 +13,70 @@ class ElasticsearchOutputModule(shared_elastic.SharedElasticsearchOutputModule):
...
+          },
+          "user_identifier": {
+              "type": "text",
+              "fields": {"keyword": {"type": "keyword"}},
+          },
+          "version": {
+              "type": "text", "fields": {"keyword": {"type": "keyword"}},
+          },
+      }
+  }
```
kiddinn, line 79: I'm wondering... how often will this list change? is it possible to create this more robustly, what happens when a new parser/plugin comes in that has the same attribute name as someone else, then we need to have a way to check whether or not we need to change this mapping, so I'm wondering whether we can have at least some sort of maintenance script that can do the check and add to this list?

plaso/output/elastic.py
```diff
@@ -13,7 +13,70 @@ class ElasticsearchOutputModule(shared_elastic.SharedElasticsearchOutputModule):
...
+              "type": "text",
+              "fields": {"keyword": {"type": "keyword"}},
+          },
+          "user_identifier": {
+              "type": "text",
+              "fields": {"keyword": {"type": "keyword"}},
+          },
+          "version": {
+              "type": "text", "fields": {"keyword": {"type": "keyword"}},
+          },
```
kiddinn, line 77: since the mapping is more or less the same, I'm wondering whether we should have this DEFAULT_MAPPING simpler and then to generate the actual mapping.
```
DEFAULT_MAPPING = {
   'version': 'text',
   'user_identifier': 'text',
...
}
```

Then in the mappings we can do

```
mappings = {}
for key, type_string in self.DEFAULT_MAPPING.items():
  mappings[key] = {
      'type': type_string, 'fields': {'keyword': {'type': 'keyword'}
  }
```

something like that, then the original dict is more easily readable, and perhaps easier to maintain

kiddinn (commented):
some comments

kiddinn:
sorry for the late review, but PTAL @william-billaud

plaso/output/elastic.py
william-billaud: I don't think that generate the mapping from an alternative notation would make it easier to maintain.
The advantage of the current solution is that it fits the elasticsearch syntax. Using an intermediate notation would make it harder to make the link between the code and the mapping that will be applied.

Furthermore, it could be interesting in the future to create a more specific mapping (e.g: specific tokenizer for path, registry key, (https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pathhierarchy-tokenizer.html), store the length of some filename, etc...). That would allow more powerful/faster query.

(Btw, currently it easy to make it by creating your own index before upload with psort)
Keep the elasticsearch syntax would also be more friendly for someone that want to extend the mapping with his own customization.

plaso/output/elastic.py
william-billaud: I have already made a script  to detect these collision. I have no longer access to it but can recreate it/ ask to get it back. this script basically parse the comment from parser and extract the field name and type for each data type. Once every parser have been parsed, it detect if some field with the same name have different type.

It's not perfect because the comment could be inacurrate (E.g for the windows:srum:application_usage data_type, the application field is specified as a string, but in some case it could also be an integer) 

I don't think this list would change really often, because most of time if field have the same name, they contains the same type of data

kiddinn: ok, we should probably think about having a script of some sort checked in as well, to be able to maintain this.

plaso/output/elastic.py
kiddinn: ACK and yes I agree, we should start looking more into different type of tokenization for these, the same goes for dfdatetime objects later in the future, that may represent `fuzzy` timestamps (as in will contain a timestamp and an accuracy, essentially making it a range) or ranges.

plaso/output/elastic.py
kiddinn: can you resolve this one

kiddinn (commented):
still waiting on the quotes

joachimmetz:
@william-billaud any specific reason for closing the PR?

kiddinn:
I think this PR should be committed, was this an accidental close, or?

william-billaud:
Just make a failed rebase...(that automatically close pr)

kiddinn (commented):
ok, this gets a LGTM from me, @joachimmetz can you take a look as well?

joachimmetz:
ack, I'll have a look later today, some other stuff that needs taking care of first

william-billaud:
btw here is the script that I used to detect these collision
https://gist.github.com/william-billaud/3c3dd2e4f20fde1d2f298b0174313493

It did not take care of containers, so it could miss some type collision

joachimmetz:
@william-billaud looks like I broke the tests on the dev track by updating some of the dependencies. I'll fix that first and then look at merging your PR. From an initial glance I have no immediate concerns.

joachimmetz:
tests should be fixed, rerunning.

joachimmetz (approved):
LGTM

joachimmetz:
@william-billaud thx for the changes, @kiddinn thx for the review

Issue #2785, opened 2019-12-04
Move source types to output modules


Onager:
Consider moving source types from formatters to output modules

joachimmetz:
Source is used by the following output formats:
* dynamic
* l2t_csv
* tln
* elasticsearch, timesketch

IMHO we can remove source from elasticsearch, timesketch and dynamic but likely need to keep it for backwards compatibility with l2t_csv and tln. A possible solution is to have some configuration (in code or file) that maps event data type to source with a default of LOG.

joachimmetz:
@kiddinn who indicated that source might be use TS analyzers

joachimmetz:
Move sources to a sources field formatter that reads the configuration from file (csv?).
```
data_type,source_long,source_short
```

joachimmetz:
The file stat formatter has a "customized" (long) source formatter https://github.com/log2timeline/plaso/blob/master/plaso/formatters/file_system.py#L98.

@kiddinn what were the design ideas behind this?

kiddinn:
As chatted offline, unsure about the origin of this, safe to remove imho

joachimmetz:
Step 1 to remove customized sources fields https://github.com/log2timeline/plaso/pull/3214

joachimmetz:
Step 2 move source field definitions to configuration file

joachimmetz:
Todo step 3, remove source field definitions from formatters

joachimmetz:
Todo step 4 update documentation

joachimmetz:
Not removing for now but have moved this to an output config file

PR #3215, opened 2020-08-26
Changes to move source types to output modules #2785

joachimmetz:
Changes to move source types to output modules #2785

### DIFF
--- a/data/sources.config
+++ b/data/sources.config
@@ -1,0 +1,192 @@
+data_type	short_source	source
+android:event:call	LOG	Android Call History
+android:event:last_resume_time	LOG	Android App Usage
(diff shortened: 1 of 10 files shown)


data/sources.csv
```diff
@@ -0,0 +1,192 @@
+data_type	short_source	source
```
rgayon, line 1: Is this a "CSV" when it's tab separated? :)

rgayon (approved):
Nit

data/sources.csv
joachimmetz: what do you prefer? `sources.conf` instead?

joachimmetz: changed to `sources.config`

rgayon (approved):

Issue #3087, opened 2020-06-14
Add data format information/description to parsers and parser plugins


joachimmetz:
Add data format information/description to parsers and parser plugins that can be used to supported formats information (also see https://github.com/log2timeline/plaso/issues/3036)

https://github.com/log2timeline/plaso/pull/3085 introduces a script to generate supported formats information. To not have to replicate information about parsers and parser plugins into the script:

* Add data format information/description to parsers and parser plugins

PR #3225, opened 2020-09-07
Replaced description by data format of various parser plugins #3087

joachimmetz:
Replaced description by data format of various parser plugins #3087

### DIFF
--- a/utils/export_supported_formats.py
+++ b/utils/export_supported_formats.py
@@ -9,10 +9,14 @@ import argparse
 import collections
 import importlib
 import inspect
+import os
 import pkgutil
 import sys
 
 from urllib import parse as urllib_parse
+
+from dtfabric import reader as dtfabric_reader
+from dtfabric import registry as dtfabric_registry
 
 import plaso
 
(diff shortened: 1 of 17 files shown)


utils/export_supported_formats.py
```diff
@@ -294,14 +275,29 @@ def _GetDataFormatInformationFromPackage(self, package):
...
+
+            dtfabric_file = os.path.join(package_path, ''.join([name, '.yaml']))
+            if os.path.exists(dtfabric_file):
+              definitions_registry = (
+                  dtfabric_registry.DataTypeDefinitionsRegistry())
+              definitions_reader = (
+                  dtfabric_reader.YAMLDataTypeDefinitionsFileReader())
+
+              try:
+                definitions_reader.ReadFile(definitions_registry, dtfabric_file)
```
rgayon, line 294: I don't see any code using definitions_registry. Is this normal?

If another PR is coming, can you add an explicit TODO?

joachimmetz: Ack, this is what I meant with `# TODO: determine url.` but agree that this is not very explicit ;) I'll update the TODO

joachimmetz: updated the TODO

joachimmetz:
@rgayon PTAL

rgayon (approved):
I see codefactor is not happy with the try/Exception. Maybe you want to ignore

joachimmetz:
> I see codefactor is not happy with the try/Exception. Maybe you want to ignore

thx I'll ignore, I don't find a lot what codefactor flags actionable and therefore that helpful. The broad except is there to have the script continue, since this is a utility script the impact is minimal.

PR #3292, opened 2020-11-06
Changes to tox Python 3.9 Travis CI test

joachimmetz:
Changes to tox Python 3.9 Travis CI test

To do:
* determine why 3.9 test is failing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,10 +83,10 @@ jobs:
     python: 3.8
     services:
     - docker
-  - name: "Ubuntu Focal (20.04) (Docker) with Python 3.9 (tox)"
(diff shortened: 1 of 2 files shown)


rgayon (commented):
It looks like libbde won't build with 3.9

joachimmetz:
> It looks like libbde won't build with 3.9

For some reason it tries to build against Python 3.8

rgayon:
The problem now is we need to build packages for Groovy

joachimmetz:
> The problem now is we need to build packages for Groovy

Not necessarily this is a test that uses tox, that uses PyPI not the GIFT PPA

but my changes did not seem to do the trick, I'll marked this as blocked until I've figured out the root cause

joachimmetz:
Since we're no longer using Travis CI I'll close this

Issue #3301, opened 2020-11-15
text parser: improve handling of decode errors


joachimmetz:
Per: https://github.com/log2timeline/plaso/issues/3298 some input might be malformed. To better handle such corrupt input (1) define an encoding error handler to escape the incorrectly encoded data and (2) track a parse error

* [x] ~~Change dfVFS text file helper to support an error handler https://github.com/log2timeline/dfvfs/pull/508~~
* [x] Change text parser to define and use an encoding error handler

For now do not allow a decode error during validation.

PR #3302, opened 2020-11-15
Changes to text parser to handle decode errors #3301

joachimmetz:
Changes to text parser to handle decode errors #3301

### DIFF
--- a/plaso/parsers/text_parser.py
+++ b/plaso/parsers/text_parser.py
@@ -7,6 +7,7 @@ parser.
 """
 
 import abc
+import codecs
 
 import pyparsing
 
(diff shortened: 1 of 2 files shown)


plaso/parsers/text_parser.py
```diff
@@ -179,6 +180,33 @@ def __init__(self):
...
+      raise TypeError('Unsupported exception type.')
+
+    if self._parser_mediator:
+      self._parser_mediator.ProduceExtractionWarning(
+          'error decoding 0x{0:02x} at offset: {1:d}'.format(
+              exception.object[exception.start],
+              self._current_offset + exception.start))
+
+    escaped = '\\x{0:2x}'.format(exception.object[exception.start])
+    return (escaped, exception.start + 1)
```
Onager, line 209: Shouldn't this be exception.**end** (no +1) instead? I assume start and end will be the same after a 1-byte error, but if there's multiple bytes of invalid data, wouldn't we want to resume parsing at the end of that?

plaso/parsers/text_parser.py
```diff
@@ -179,6 +180,33 @@ def __init__(self):
...
+    codecs.register_error('text_parser_handler', self._EncodingErrorHandler)
+
+  def _EncodingErrorHandler(self, exception):
+    """Encoding error handler.
+
+    Args:
+      exception [UnicodeDecodeError]: exception.
+
+    Returns:
+      tuple[str, int]: replacement string and number of bytes to skip.
```
Onager, line 194: The docs for [register_error() ](https://docs.python.org/3/library/codecs.html#codecs.register_error) say that the int here is where encoding should continue, not the number of bytes that should be skipped. Please correct the docstring here.

plaso/parsers/text_parser.py
```diff
@@ -179,6 +180,33 @@ def __init__(self):
...
+    if not isinstance(exception, UnicodeDecodeError):
+      raise TypeError('Unsupported exception type.')
+
+    if self._parser_mediator:
+      self._parser_mediator.ProduceExtractionWarning(
+          'error decoding 0x{0:02x} at offset: {1:d}'.format(
+              exception.object[exception.start],
+              self._current_offset + exception.start))
+
+    escaped = '\\x{0:2x}'.format(exception.object[exception.start])
```
Onager, line 208: This only handles the case of a 1-byte error, are you sure you want 1 exception message per invalid byte?

Onager (changes requested):

plaso/parsers/text_parser.py
joachimmetz: I'm not sure, I took the conservative approach to account for single byte data that might case multi byte decoding errors

plaso/parsers/text_parser.py
joachimmetz: changed

plaso/parsers/text_parser.py
joachimmetz: I took the conservative approach to account for single byte data that might case multi byte decoding errors. We could limit the number of warning that will be generated if you expect a lot of encoding errors.

joachimmetz:
@Onager PTAL

Issue #3287, opened 2020-11-04
Double check if parser and plugins that process input in local time pass time zone to event


joachimmetz:
Per https://github.com/log2timeline/plaso/issues/3280 double check if parser and plugins that process input in local time pass time zone to event

Also add unit tests to catch this

joachimmetz:
Parsers that use: parser_mediator.timezone

* apt_history.py
* msiecf.py
* setupapi.py
* sophos_av.py
* syslog.py
* vsftpd.py
* winfirewall.py
* winjob.py
* winreg_plugins/ccleaner.py
* xchatlog.py

joachimmetz:
mostly done, bumbing long tail to next release

PR #3310, opened 2020-11-16
Added tests for parser preferred time zone handling #3287

joachimmetz:
Added tests for parser preferred time zone handling #3287

### DIFF
--- a/tests/parsers/xchatlog.py
+++ b/tests/parsers/xchatlog.py
@@ -17,6 +17,22 @@ class XChatLogUnitTest(test_lib.ParserTestCase):
   def testParse(self):
     """Tests the Parse function."""
     parser = xchatlog.XChatLogParser()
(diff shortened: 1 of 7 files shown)


tests/parsers/xchatlog.py
```diff
@@ -14,9 +14,25 @@
 class XChatLogUnitTest(test_lib.ParserTestCase):
   """Tests for the xchatlog parser."""
 
-  def testParse(self):
+  def testParseWith(self):
```
rgayon, line 17: *testParseWithTimeZone

rgayon (approved):
Just 1 nit

tests/parsers/xchatlog.py
joachimmetz: good catch

joachimmetz: this should be just testParse

Issue #3312, opened 2020-11-19
remove elasticsearch 6 support it is near end of life


joachimmetz:
Per: https://www.elastic.co/support/eol

Elasticsearch | EOL date | Maintained until
-- | -- | --
6.8.x | 2020-11-20 | 8.0.0

* [x] ~~disable elasticsearch 6 end-to-end tests~~
* [x] remove elasticsearch 6 end-to-end tests
* [x] remove elasticsearch-py 6 test dependency from GIFT PPA (~~testing~~, ~~dev~~)
* [x] remove elasticsearch-py 6 from codebase https://github.com/log2timeline/plaso/pull/3313

PR #3313, opened 2020-11-19
Removed Elasticsearch 6 support #3312

joachimmetz:
Removed Elasticsearch 6 support #3312

### DIFF
--- a/plaso/output/shared_elastic.py
+++ b/plaso/output/shared_elastic.py
@@ -182,7 +182,16 @@ class SharedElasticsearchOutputModule(interface.OutputModule):
     self._url_prefix = None
 
   def _Connect(self):
-    """Connects to an Elasticsearch server."""
+    """Connects to an Elasticsearch server.
+
+    Raises:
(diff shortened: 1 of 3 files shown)


plaso/output/shared_elastic.py
```diff
@@ -182,7 +182,16 @@ def __init__(self, output_mediator):
...
 
   def _Connect(self):
-    """Connects to an Elasticsearch server."""
+    """Connects to an Elasticsearch server.
+
+    Raises:
+      RuntimeError: if the Elasticsearch version is not supported or the server
+          cannot be reached.
+    """
+    if 6 < elasticsearch.__version__[0] < 8:
```
rgayon, line 191: IIUC, the only ES version which will trigger the RuntimeError is 7, while it looks like we want to support versions > 6

joachimmetz: good catch, changed.

joachimmetz:
@rgayon PTAL

Issue #1879, opened 2018-05-10
elasticsearch output module: add option to load a custom mappings file 


joachimmetz:
Per https://github.com/log2timeline/plaso/issues/1156 add an option to load a custom mappings file (JSON) for the elasticsearch output module.

Consider how to test this:
* e.g. add ES connectivity test that tests the mappings as well?

Onager:
Not going to make the June release, bumping to next release.

Onager:
Not making September release, removing milestone.

goorgoorlou:
Hi,
Will this feature be added in the next milestone (>Nov 2018)?  (As the problem of #1156 still exist).
Thank you!

joachimmetz:
resurfacing this since it is also needed to be able to combine ES and TS output modules

PR #3314, opened 2020-11-19
Added options to set custom Elasticsearch mappings #1879

joachimmetz:
Added options to set custom Elasticsearch mappings #1879

### DIFF
--- a/data/elasticsearch.mappings
+++ b/data/elasticsearch.mappings
@@ -1,0 +1,77 @@
+{
+  "properties": {
+    "application": {
+      "type": "text",
+      "fields": {
+        "keyword": {"type": "keyword"}}
+    },
+    "data": {
+      "type": "text",
+      "fields": {"keyword": {"type": "keyword"}}
+    },
+    "doc_type": {
+      "type": "text",
+      "fields": {"keyword": {"type": "keyword"}}
+    },
+    "event_type": {
+      "type": "text",
+      "fields": {"keyword": {"type": "keyword"}}
+    },
+    "exit_status": {
+      "type": "text",
+      "fields": {"keyword": {"type": "keyword"}}
+    },
+    "facility": {
+      "type": "text",
+      "fields": {"keyword": {"type": "keyword"}}
+    },
+    "file_reference": {
+      "type": "text",
+      "fields": {"keyword": {"type": "keyword"}}
+    },
+    "file_size": {
+      "type": "text",
+      "fields": {"keyword": {"type": "keyword"}}
+    },
+    "flags": {
+      "type": "text",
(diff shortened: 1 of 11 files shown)


tests/cli/helpers/elastic_output.py
```diff
@@ -22,43 +22,46 @@ class ElasticSearchOutputArgumentsHelperTest(
...
 
 Test argument parser.
 
 optional arguments:
-  --ca_certificates_file_path CA_CERTIFICATES_FILE_PATH
+  --ca_certificates_file_path PATH, --ca-certificates-file-path PATH
                         Path to a file containing a list of root certificates
                         to trust.
-  --doc_type DOCUMENT_TYPE
+  --doc_type DOCUMENT_TYPE, --doc-type DOCUMENT_TYPE
```
berggren, line 39: Since ES 7, doc_type has been deprecated. We should consider removing it from Plaso as well. Let's do that in a followup PR though. Maybe add a TODO here?

data/elasticsearch.mappings
```diff
@@ -0,0 +1,77 @@
+{
```
berggren, line 1: Where does this come from? Do we generate this mapping? It's just a subset of the attributes avaialble so I don't see the need for this. ES should map correctly by default in these cases.

berggren (changes requested):
Overall looks good, a couple of initial comments.

tests/cli/helpers/elastic_output.py
joachimmetz: I've created an isssue https://github.com/log2timeline/plaso/issues/3415, thx for flagging

data/elasticsearch.mappings
joachimmetz: this comes from the elastic output module https://github.com/log2timeline/plaso/pull/3314/files/f6652638bb3da6c69917cd8142108d907a56677d#diff-48a52c9b262a7ecbed3950483191eeec5492f0983c87eb7e580e5a6656623efeL17 I think @kiddinn implemented this back in the day

joachimmetz: Correction this comes from https://github.com/log2timeline/plaso/commit/6a5f6e918203bfa126f4d0572bbeb5fed67b208a

joachimmetz: This change was made to work-around bigint issue seen with some entries.

joachimmetz:
@berggren PTAL

data/elasticsearch.mappings
berggren: Ah, I see. Ok, makes sense.

berggren (approved):
LGTM

Issue #1293, opened 2017-07-09
add LVM support


joachimmetz:
Originally part of https://github.com/log2timeline/plaso/issues/109

- [x] add LVM support
  - Requires https://github.com/log2timeline/dfvfs/issues/89

PR #3317, opened 2020-11-20
Added LVM support #1293

joachimmetz:
Added LVM support #1293

### DIFF
--- a/plaso/cli/storage_media_tool.py
+++ b/plaso/cli/storage_media_tool.py
@@ -17,6 +17,7 @@ from dfvfs.lib import definitions as dfvfs_definitions
 from dfvfs.lib import errors as dfvfs_errors
 from dfvfs.path import factory as path_spec_factory
 from dfvfs.volume import apfs_volume_system
+from dfvfs.volume import lvm_volume_system
 from dfvfs.volume import tsk_volume_system
 from dfvfs.volume import vshadow_volume_system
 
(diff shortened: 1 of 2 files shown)


plaso/cli/storage_media_tool.py
```diff
@@ -203,6 +212,59 @@ def _GetAPFSVolumeIdentifiers(self, scan_node):
...
+    """Determines the LVM volume identifiers.
+
+    Args:
+      scan_node (dfvfs.SourceScanNode): scan node.
+
+    Returns:
+      list[str]: LVM volume identifiers.
+
+    Raises:
+      SourceScannerError: if the format of or within the source is not
```
rgayon, line 225: I don't understand the comment here

plaso/cli/storage_media_tool.py
```diff
@@ -684,6 +777,48 @@ def _PromptUserForAPFSVolumeIdentifiers(
...
+    Args:
+      volume_system (dfvfs.LVMVolumeSystem): volume system.
+      volume_identifiers (list[str]): volume identifiers including prefix.
+
+    Returns:
+      list[str]: selected volume identifiers including prefix or None.
+    """
+    print_header = True
+    while True:
+      if print_header:
```
rgayon, line 793: Why not just print the header outside the "while" loop?
print_header is not set to true ever again

joachimmetz: good question, print_header was mainly used because to keep it consistent with other similar sections in this file, but there seems no need for it, I've removed it in the other places as well.

plaso/cli/storage_media_tool.py
joachimmetz: reworded

joachimmetz:
@rgayon PTAL

rgayon (approved):

PR #3320, opened 2020-11-20
Added --no-hashes option to not generate hashes.json in image_export tool

rgayon:
Added --no-hashes option to not generate hashes.json in image_export tool.

I assume this might prevent some bugs if the tool is used to collect a file called 'hashes.json' that is at the root of the image disk.

This also helps some automated tools (like Turbinia) to not implement dedicated logic to ignore that file


plaso/cli/image_export_tool.py
```diff
@@ -670,6 +671,10 @@ def ParseArguments(self, arguments):
             'exported files and duplicates are skipped. Use this option to '
             'include duplicate files in the export.'))
 
+    argument_parser.add_argument(
+        '--no-hashes', dest='no_hashes', action='store_true', default=False,
```
joachimmetz, line 675: Please add `--no_hashes` option as well

joachimmetz:
@rgayon please rebase with HEAD

joachimmetz (approved):
LGTM

PR #3346, opened 2020-12-20
Clean up of message formatter functions

joachimmetz:
Clean up of message formatter functions


plaso/formatters/winreg.py
```diff
@@ -16,39 +15,55 @@ class WinRegistryGenericFormatter(interface.EventFormatter):
...
     if 'key_path' in event_values:
       format_string = self.FORMAT_STRING
     else:
       format_string = self.FORMAT_STRING_ALTERNATIVE
 
-    return self._FormatMessages(
-        format_string, self.FORMAT_STRING_SHORT, event_values)
+    return self._FormatMessage(format_string, event_values)
+
+  def GetMessageShort(self, event_values):
```
rgayon, line 44: Is this different than the GetMessageShort() in plaso/formatters/interface.py ?

plaso/output/formatting_helper.py
```diff
@@ -332,6 +348,30 @@ def _FormatUsername(self, event, event_data, event_data_stream):
...
+    message_string = None
+    if event_data.source_name and event_data.message_identifier:
+      windows_event_message = self._output_mediator.GetWindowsEventMessage(
+          event_data.source_name, event_data.message_identifier)
+      if windows_event_message:
+        try:
+          message_string = windows_event_message.format(*event_data.strings)
+        except IndexError:
+          # Unable to create the message string.
+          pass
```
rgayon, line 371: Should we still try to return something?
IIUC, here we do have a windows_event_message but just not found a way to format it

plaso/formatters/winreg.py
joachimmetz: yes due to `FORMAT_STRING_ALTERNATIVE` which is a left over from earlier days only used by this formatter

plaso/output/formatting_helper.py
joachimmetz: ack, I'll add a todo to address this in a later change

rgayon (approved):

Issue #3042, opened 2020-06-09
Make event values tests more data driven


joachimmetz:
In line of https://github.com/log2timeline/plaso/issues/2937 make the event values tests more data driven.

Clean up:
* Remove need for CheckTimestamp
* Remove need for `event = events[` usage

Close issue after https://github.com/log2timeline/plaso/pull/3414 is merged continuation of clean up in https://github.com/log2timeline/plaso/issues/3380

joachimmetz:
First iteration in https://github.com/log2timeline/plaso/pull/3043

PR #3352, opened 2020-12-22
Changes to make event values tests data driven #3042

joachimmetz:
Changes to make event values tests data driven #3042


rgayon:
Looks like this needs to be synced to #3351

joachimmetz:
rebased with HEAD @rgayon PTAL

rgayon (approved):

PR #3362, opened 2020-12-24
Added path artifact attribute container

joachimmetz:
Added path artifact attribute container

### DIFF
--- a/plaso/containers/artifacts.py
+++ b/plaso/containers/artifacts.py
@@ -219,6 +219,176 @@ class OperatingSystemArtifact(ArtifactAttributeContainer):
     return False
 
 
+class PathArtifact(ArtifactAttributeContainer):
+  """Path artifact attribute container.
+
+  Attributes:
+    data_stream (str): name of a data stream.
+    path_segment_separator (str): path segment separator.
+    path_segments (list[str]): path segments.
+  """
+  CONTAINER_TYPE = 'path'
+
+  def __init__(self, data_stream=None, path=None, path_segment_separator='/'):
+    """Initializes a path artifact.
+
+    Args:
+      data_stream (Optional[str]): name of a data stream.
+      path (Optional[str]): a path.
+      path_segment_separator (Optional[str]): path segment separator.
+    """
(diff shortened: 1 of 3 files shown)


plaso/containers/artifacts.py
```diff
@@ -219,6 +219,176 @@ def IsEquivalent(self, other):
...
+    super(PathArtifact, self).__init__()
+    self.data_stream = data_stream
+    self.path_segment_separator = path_segment_separator
+    self.path_segments = self._SplitPath(path, path_segment_separator)
+
+  def __eq__(self, other):
+    """Determines if the path is equal to other.
+
+    Args:
+      other (str): path to compare against.
```
rgayon, line 249: Shouldn't we allow comparing PathArtifact objects with eachother?

plaso/containers/artifacts.py
```diff
@@ -219,6 +219,176 @@ def IsEquivalent(self, other):
...
+  """Path artifact attribute container.
+
+  Attributes:
+    data_stream (str): name of a data stream.
+    path_segment_separator (str): path segment separator.
+    path_segments (list[str]): path segments.
+  """
+  CONTAINER_TYPE = 'path'
+
+  def __init__(self, data_stream=None, path=None, path_segment_separator='/'):
```
rgayon, line 232: Should the default path segment separator be OS independant?

plaso/containers/artifacts.py
joachimmetz: we should, I was planning to add this as a later stage.

plaso/containers/artifacts.py
joachimmetz: How do you see/imagine this? My thoughts were to set this on a per file system level or to maybe set this in the output.

rgayon: I have no idea. If this is just used internally by Plaso then I guess it doesn't really matter, I just wanted to point this out in case it was overlooked

rgayon (approved):

PR #3400, opened 2021-01-05
Renamed warning to extraction warning

joachimmetz:
Renamed warning to extraction warning

### DIFF
--- a/plaso/storage/interface.py
+++ b/plaso/storage/interface.py
@@ -242,17 +242,18 @@ class BaseStore(object):
         self._CONTAINER_TYPE_EVENT_TAG, event_tag,
         serialized_data=serialized_data)
 
-  def AddWarning(self, warning, serialized_data=None):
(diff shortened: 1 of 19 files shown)


plaso/storage/interface.py
```diff
@@ -22,7 +22,7 @@ class BaseStore(object):
   Attributes:
     format_version (int): storage format version.
     serialization_format (str): serialization format.
-    storage_type (str): storage type.
+    storage_typg (str): storage type.
```
rgayon, line 25: I wonder what sequence of vi shortcuts would make that typo =)

rgayon (approved):
just one typo

plaso/storage/interface.py
joachimmetz: likely more a factor of me randomly hitting the keyboard at some strange time of day, changed

PR #3442, opened 2021-01-18
Changes for Timesketch output mappings file

joachimmetz:
Changes for Timesketch output mappings file

### DIFF
--- a/config/jenkins/greendale/output_timesketch.ini
+++ b/config/jenkins/greendale/output_timesketch.ini
@@ -2,5 +2,5 @@
 case=output
 output_file=studentpc1.csv
 output_format=timesketch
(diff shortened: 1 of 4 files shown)


config/jenkins/greendale/output_timesketch.ini
```diff
@@ -2,5 +2,5 @@
 case=output
 output_file=studentpc1.csv
 output_format=timesketch
-output_options=--elastic_mappings data/timesketch.mappings --name test --username test
```
rgayon, line 5: Just to confirm removing the elastic mapping args is WAI

rgayon (approved):
LGTM, just confirm removing the elastic mapping args is WAI

PR #3448, opened 2021-01-22
Fix missing spaces in ssh Message strings

obsidianforensics:
## Fix missing spaces in ssh Message strings
I noticed some message strings were missings spaces:

> Successful login of user: ryanfrom 10.1.1.1:53570using authentication method: passwordssh pid: 1234

Small PR to add those in.

## Checklist:
  - [ ] Reviewer assigned

### DIFF
--- a/data/formatters/generic.yaml
+++ b/data/formatters/generic.yaml
@@ -635,10 +635,10 @@ short_message:
 type: 'conditional'
 data_type: 'syslog:ssh:failed_connection'
 message:
-- 'Unsuccessful connection of user: {username}'
+- 'Unsuccessful connection of user: {username} '
 - 'from {address}:'
-- '{port}'
-- 'using authentication method: {authentication_method}'
+- '{port} '
+- 'using authentication method: {authentication_method} '
 - 'ssh pid: {pid}'
 separator: ''
 short_message:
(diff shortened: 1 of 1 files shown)


data/formatters/generic.yaml
```diff
@@ -635,10 +635,10 @@ short_message:
...
 message:
-- 'Unsuccessful connection of user: {username}'
+- 'Unsuccessful connection of user: {username} '
 - 'from {address}:'
-- '{port}'
-- 'using authentication method: {authentication_method}'
+- '{port} '
+- 'using authentication method: {authentication_method} '
 - 'ssh pid: {pid}'
 separator: ''
```
joachimmetz, line 643: should we remove the custom separator instead and just use default of space?

obsidianforensics: I think then we'd have a space between IP and port
> ...from 10.0.0.1: 4566 using... 

I'm not too familiar with the formatters, but I'm assuming there's a reason each variable is on its own line. Otherwise, we could combine address:port on one line, then use the default space separator.

joachimmetz: > I'm not too familiar with the formatters, but I'm assuming there's a reason each variable is on its own line.

mainly historical, was sufficient at the time. I can make a note to enhance the message formatters in this aspect. There is more formatting work in the pipeline

joachimmetz: more info can be found here: https://plaso.readthedocs.io/en/latest/sources/user/Output-and-formatting.html#message-formatting

but the tl;dr is that the individual string (parts/segment) will be combined using the separator. A string part will be present if the value is available.

joachimmetz (approved):
LGTM

Issue #3449, opened 2021-01-22
psort: +1.4M events with the same timestamp exceed 4G memory limit


joachimmetz:
This should be and edge case but a Plaso storage file with +1.4M events with the same timestamp (rare and unlikely to be very useful in analysis) causes psort to exceed the 4G memory limit due to growth of the psort event heap. Adding a limit e.g. max 500k events on the heap prevents exceeding the memory limit. Determine if such a cap has additional side effects.

Reason for the +1.4M events with the same timestamp are default "limitations" in the bodyfile format

joachimmetz:
Lowered the limit to 100k in https://github.com/log2timeline/plaso/pull/3531 seeing that the base memory consumption can be different for different installations of Python.

PR #3450, opened 2021-01-23
Added limit in psort for events with the same timestamp #3449

joachimmetz:
Added limit in psort for events with the same timestamp #3449


plaso/multi_processing/psort.py
```diff
@@ -476,7 +476,8 @@ def _ExportEvent(
       deduplicate_events (Optional[bool]): True if events should be
           deduplicated.
     """
-    if event.timestamp != self._export_event_timestamp:
+    if (event.timestamp != self._export_event_timestamp or
+        self._export_event_heap.number_of_events > 500000):
```
joachimmetz, line 480: Move 500000 to a class constant

Issue #3172, opened 2020-07-27
FR: Have ElasticSearch output module export filename for all so it can be searched


davidrudduck:
**Description of request:**
Prior to 20200717, if we were investigating an incident and wanting to deep dive on a particular users artifacts/events, we would use an elasticsearch query like "filename: username" or "filename: username*" to filter down to a particular folder path like C:\Users\username" or "C:\Users\username.domain"

As this field is no longer being populated this is not possible - and the path_spec field is being populated via a JSON value instead of being broken out into sub fields, which is not as easily searchable.

If the data in path_spec was broken out into sub fields (path_spec.type, path_spec.location, etc) instead of being lumped in one field as JSON it would re-enable the ability to filter down based on the path of the artefact(s) we want to show events on.

**Plaso version:**

For example 20200717

**Operating system Plaso is running on:**

Ubuntu 18.04 on WSLv1

**Installation method:**
via PPA

** Thoughts / suggestions **

The alternative workaround for this would be requiring ElasticSearch pipelines to be built to break up the JSON values in the pathspec field; or alternatively just break pathspec out into pathspec.type, pathspec.location and pathspec.indicator rather than dropping the entire JSON into a single value which then needs to be processed to work properly

davidrudduck:
To explain the workflow, when we are doing windows forensics, we will often start by trying to find the login events (typically by filtering down to windows event logs and the event id's relevant to logon events). 

Once we've identified the accounts that we believe have been compromised, and the logon/logoff times of those sessions, we will then filter down on the artefacts that come from that users folder

so if "john.domain" is the user that we believe was compromised, we will run a query in elasticsearch to show all artifacts below C:\Users\john.domain for the period of the session.

Previously we used the filename field to do this by just using the query "filename: john.domain" (or sometimes "filename: Users AND filename: john*".

We are unable to use this type of query on path_spec due to the JSON formatting of values stored within it. Lucene and KQL returns zero results.

davidrudduck:
Currently "path_sec" gets populated with JSON in ElasticSearch like such: 

{"__type__": "PathSpec", "location": "/mnt/e/Cases/CaseName/Plaso/lcl-backupserver.body", "type_indicator": "OS"}

If I were building an ingestion pipeline for ElasticSearch, I would look at this JSON and break it down into it's individual elements.

[...]

PR #3463, opened 2021-02-01
Added additional fields support to elastic output module #3172

kiddinn:

Changing the shared_elastic output module to allow you to define extra fields to add to each event as it gets indexed into Elastic.

## Description:

The reason is a move for Timesketch to stop using the Timesketch output module in favor of the Elastic one. This would mean that the TS specific things that are done in the output module are going to be moved to TS itself, and all that psort does is to ingest data into ES.

However in order for Timesketch to be able to identify each event as belonging to a particular timeline new events need to be added to each document that is indexed. This PR provides the option to do so.

## Checklist:
  - [x] Reviewer assigned

### DIFF
--- a/plaso/cli/helpers/elastic_output.py
+++ b/plaso/cli/helpers/elastic_output.py
@@ -61,7 +61,7 @@ class ElasticSearchOutputArgumentsHelper(interface.ArgumentsHelper):
             'Export string fields that will not be analyzed by Lucene.'))
 
     argument_group.add_argument(
-        '--additional', '--additional-fields', dest='additional_fields',
+        '--additional_fields', '--additional-fields', dest='additional_fields',
         action='store', default='', help=(
             'JSON string with a dict of fields and values to add to each '
             'event that will be indexed by Elastic.'))
@@ -130,21 +130,21 @@ class ElasticSearchOutputArgumentsHelper(interface.ArgumentsHelper):
     additional_fields_string = cls._ParseStringOption(
         options, 'additional_fields')
 
+    additional_fields = {}
     if additional_fields_string:
       try:
         additional_fields = json.loads(additional_fields_string)
-      except (json.JSONDecodeError, ValueError) as exc:
+      except (json.JSONDecodeError, ValueError) as exception:
         additional_fields = {}
         logger.warning(
-            'Unable to parse the extra fields, with error: {0!s}'.format(exc))
+            'Unable to parse the extra fields, with error: {0!s}'.format(
+                exception))
 
       if not isinstance(additional_fields, dict):
         logger.warning(
-            'Extra fields needs to be a dictionary, not [{0!s}]'.format(
+            'Extra fields needs to be a dictionary, not "{0!s}"'.format(
                 type(additional_fields)))
         additional_fields = {}
-    else:
-      additional_fields = {}
 
     ca_certificates_path = cls._ParseStringOption(
         options, 'ca_certificates_file_path')
--- a/plaso/cli/helpers/elastic_output.py
+++ b/plaso/cli/helpers/elastic_output.py
@@ -61,6 +61,12 @@ class ElasticSearchOutputArgumentsHelper(interface.ArgumentsHelper):
             'Export string fields that will not be analyzed by Lucene.'))
 
     argument_group.add_argument(
+        '--extra_fields', '--extra-fields', dest='extra_fields',
+        action='store', default='', help=(
+            'JSON string with a dict of fields and values to add to each '
+            'event that will be indexed by Elastic.'))
+
+    argument_group.add_argument(
         '--elastic_mappings', '--elastic-mappings', dest='elastic_mappings',
         action='store', default=None, metavar='PATH', help=(
             'Username to use for Elasticsearch authentication.'))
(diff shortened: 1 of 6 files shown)


plaso/cli/helpers/elastic_output.py
```diff
@@ -60,6 +60,12 @@ def AddArguments(cls, argument_group):
         default=cls._DEFAULT_RAW_FIELDS, help=(
             'Export string fields that will not be analyzed by Lucene.'))
 
+    argument_group.add_argument(
```
joachimmetz, line 63: Can you change extra to additional to match dynamic output module terminology
https://github.com/log2timeline/plaso/blob/master/plaso/cli/helpers/dynamic_output.py#L40

kiddinn: done

plaso/cli/helpers/elastic_output.py
```diff
@@ -60,6 +60,12 @@ def AddArguments(cls, argument_group):
...
+        '--additional', '--additional-fields', dest='additional_fields',
+        action='store', default='', help=(
+            'JSON string with a dict of fields and values to add to each '
```
joachimmetz, line 66: Please add an example or documentation of what value is expected here

plaso/cli/helpers/elastic_output.py
```diff
@@ -121,6 +127,24 @@ def ParseOptions(cls, options, output_module):
     elastic_user = cls._ParseStringOption(options, 'elastic_user')
     elastic_password = cls._ParseStringOption(options, 'elastic_password')
     use_ssl = getattr(options, 'use_ssl', False)
+    additional_fields_string = cls._ParseStringOption(
+        options, 'additional_fields')
+
+    if additional_fields_string:
+      try:
+        additional_fields = json.loads(additional_fields_string)
+      except (json.JSONDecodeError, ValueError) as exc:
```
joachimmetz, line 136: style nit: do not use abbreviations like `exc`

plaso/cli/helpers/elastic_output.py
```diff
@@ -121,6 +127,24 @@ def ParseOptions(cls, options, output_module):
...
+        additional_fields = {}
+        logger.warning(
+            'Unable to parse the extra fields, with error: {0!s}'.format(exc))
+
+      if not isinstance(additional_fields, dict):
+        logger.warning(
+            'Extra fields needs to be a dictionary, not [{0!s}]'.format(
```
joachimmetz, line 143: for consistency don't use `[{0!s}]` but `: "{0!s}"`

plaso/cli/helpers/elastic_output.py
```diff
@@ -121,6 +127,24 @@ def ParseOptions(cls, options, output_module):
...
+                type(additional_fields)))
+        additional_fields = {}
+    else:
+      additional_fields = {}
```
joachimmetz, line 147: put `additional_fields = {}` before `if additional_fields_string:` and remove else

plaso/output/shared_elastic.py
```diff
@@ -323,6 +324,9 @@ def _InsertEvent(self, event, event_data, event_data_stream, event_tag):
     event_values = self._GetSanitizedEventValues(
         event, event_data, event_data_stream, event_tag)
 
+    if self._additional_fields:
+      event_values.update(self._additional_fields)
```
joachimmetz, line 328: I think we have to make sure there are no duplicate field names here

plaso/output/shared_elastic.py
```diff
@@ -360,6 +364,18 @@ def Close(self):
 
     self._client = None
 
+  def SetAdditionalFields(self, additional_fields):
+    """Set the additional field.
```
joachimmetz, line 368: style consistency nit: Set => Sets

plaso/output/shared_elastic.py
```diff
@@ -360,6 +364,18 @@ def Close(self):
...
+
+    Args:
+      additional_fields (dict): A dict with additional fields and values that
```
joachimmetz, line 371: remove `A dict with`, change `dict` to `dict[str, object]`

why do the values need to be passed here and not just the field names ?

plaso/cli/helpers/elastic_output.py
joachimmetz: Also why JSON and not just a string of field names?

kiddinn: well, this is a draft PR, and will not be set to a PR until I've done some testing on the TS side, and written the code there.

Then this will all be better explained, I just added this here so that you would see what the intention was, and how it looks

plaso/output/shared_elastic.py
kiddinn: again, this is a draft PR, and I'm just testing this, this is not the final version

plaso/output/shared_elastic.py
kiddinn: it absolutely needs to be a key/value pair, otherwise this would not be needed at all...

kiddinn (commented):
still a draft

plaso/output/shared_elastic.py
joachimmetz: I understand that it is still a draft. I'm also trying to understand the changes to prevent having them to change a lot later on.

Could you give me some context on why both key and value are needed?

plaso/cli/helpers/elastic_output.py
kiddinn: done

plaso/cli/helpers/elastic_output.py
kiddinn: done

plaso/cli/helpers/elastic_output.py
kiddinn: done

plaso/output/shared_elastic.py
kiddinn: consistent with many of the other function name, changing them to match...

plaso/output/shared_elastic.py
kiddinn: done

and why do values, because that's the goal here, to add a key/value pair to each event that is ingested, to make timeline filtering work in TS.

explained later

kiddinn:
This is fine, let's ship this

joachimmetz (approved):
LGTM

Issue #3485, opened 2021-02-07
Too few arguments for string format. Format "Unable to parse trigger: {0:d} with error: {2!s}" requires at least 3, but 2 are provided.


jaegeral:
https://github.com/log2timeline/plaso/blob/f90b58d1fae5a60688f1f1bae55f6dbf23be669b/plaso/parsers/winjob.py#L237

If I am not completely off this should be ```{1!s}``` in ```error:```

PR #3486, opened 2021-02-07
Update winjob.py

jaegeral:
https://github.com/log2timeline/plaso/issues/3485

LGTM semmle told me this should be two arguments and not three as expected by {2!}.


joachimmetz (approved):
LGTM

Issue #3490, opened 2021-02-08
elastic_ts: change default mappings file


joachimmetz:
For the elastic_ts output module change the default mappings file path to `/etc/timesketch/plaso.mappings`

PR #3491, opened 2021-02-08
Changed elastic_ts output to use /etc/timesketch/plaso.mappings #3490

joachimmetz:
Changed elastic_ts output to use /etc/timesketch/plaso.mappings #3490


kiddinn (approved):
LGTM

PR #3530, opened 2021-03-25
Lowered psort event heap limit

joachimmetz:
Lowered psort event heap limit


joachimmetz:
Duplicate of https://github.com/log2timeline/plaso/pull/3531

PR #3563, opened 2021-04-22
Add parsing of additional fields for Chrome file download records

obsidianforensics:
Add parsing of additional fields for Chrome file download records. For earlier Chrome, add in "state" value. For later Chrome, also add in danger_type, end_time, interrupt_reason, opened, and start_time.

## Checklist:
  - [ ] Reviewer assigned


obsidianforensics:
I'd like to also add translations of some of the values - what the interrupt_reason code means, state mean, and so on. I'm not quite sure where to do that - I imagine in a formatter, but I couldn't find a good example. Can you point me to one?

joachimmetz:
If `interrupt_reason` is an enumeration try adding a formatter enumeration helper e.g. https://github.com/log2timeline/plaso/blob/main/data/formatters/browser.yaml#L83

Also see: https://plaso.readthedocs.io/en/latest/sources/user/Output-and-formatting.html#enumeration-helpers

obsidianforensics:
Perfect, thanks. All I could find was https://github.com/log2timeline/plaso/blob/main/plaso/formatters/chrome.py, which wasn't quite right, but those helpers look like exactly what I need. Thanks.

obsidianforensics:
I've added the enumeration_helpers, but am not seeing the message string being updated? Is there somewhere else that needs to be changed as well?

obsidianforensics:
I figured it out - I was reviewing my output in Timesketch, which wasn't using my modified formatter for the message string. Ran psort and it all showed up. This is ready for review.

plaso/parsers/sqlite_plugins/chrome_history.py
```diff
@@ -15,8 +15,14 @@ class ChromeHistoryFileDownloadedEventData(events.EventData):
   """Chrome History file downloaded event data.
 
   Attributes:
+    danger_type (int): enum indicating if the download was dangerous.
+    end_time (int): when the download ended.
```
joachimmetz, line 19: Why not make this a separate event? Also what does the int represent?

plaso/parsers/sqlite_plugins/chrome_history.py
```diff
@@ -15,8 +15,14 @@ class ChromeHistoryFileDownloadedEventData(events.EventData):
...
     full_path (str): full path where the file was downloaded to.
+    interrupt_reason (int): enum indicating why the download stopped.
+    opened (int): if the downloaded file was opened from the browser.
     received_bytes (int): number of bytes received while downloading.
+    start_time (int): when the download started.
```
joachimmetz, line 24: Why not make this a separate event? Also what does the int represent?

plaso/parsers/sqlite_plugins/chrome_history.py
```diff
@@ -15,8 +15,14 @@ class ChromeHistoryFileDownloadedEventData(events.EventData):
   """Chrome History file downloaded event data.
 
   Attributes:
+    danger_type (int): enum indicating if the download was dangerous.
```
joachimmetz, line 18: Please don't abbreviate and write enumeration (also see https://github.com/log2timeline/l2tdocs/blob/main/process/Style-guide.md#docstrings)

IMHO you can leave out the fact that this is an enum (same applies to the other values)

plaso/parsers/sqlite_plugins/chrome_history.py
```diff
@@ -15,8 +15,14 @@ class ChromeHistoryFileDownloadedEventData(events.EventData):
...
+    end_time (int): when the download ended.
     full_path (str): full path where the file was downloaded to.
+    interrupt_reason (int): enum indicating why the download stopped.
+    opened (int): if the downloaded file was opened from the browser.
     received_bytes (int): number of bytes received while downloading.
+    start_time (int): when the download started.
+    state (int): state of the download (finished, cancelled, etc).
```
joachimmetz, line 25: I would opt:
```
the state of the download, such as finished or cancelled.
```

data/formatters/browser.yaml
```diff
@@ -61,11 +61,82 @@ short_message:
...
 type: 'conditional'
 data_type: 'chrome:history:file_downloaded'
+enumeration_helpers:
+- input_attribute: 'danger_type'
+  output_attribute: 'danger_type_string'
+  default_value: 'UNKNOWN'
+  # Also see: https://cs.chromium.org/chromium/src/components/download/public/common/download_danger_type.h
+  values:
+    0: 'Not Dangerous'
+    1: 'Dangerous (eg: a pdf or extension from places other than gallery)'
```
joachimmetz, line 71: nit: eg => e.g.

plaso/parsers/sqlite_plugins/chrome_history.py
joachimmetz: maybe:
```
assessment of the danger of the downloaded content.
```

Maybe indicate what this assessment is based on? Safe Browsing?

joachimmetz (changes requested):
Some questions about your changes

plaso/parsers/sqlite_plugins/chrome_history.py
obsidianforensics: All these descriptions were copied from the linked Chromium source files. I generally try to alter the "source of truth" as little as possible, but I can change to match plaso's style.

plaso/parsers/sqlite_plugins/chrome_history.py
obsidianforensics: start_time is currently used as the event time; I thought with the additional of end_time it was more clear to have start_time as an attribute as well, even if it's duplicative. I expanded the description.

plaso/parsers/sqlite_plugins/chrome_history.py
obsidianforensics: Changed

plaso/parsers/sqlite_plugins/chrome_history.py
obsidianforensics: Updated description. I don't feel it's that useful as its own event, but rather useful context for the download event as a whole, so adding as an attribute seemed appropriate.

joachimmetz: this deviates from Plaso's current approach, especially if for later usage in automated analysis.

plaso/parsers/sqlite_plugins/chrome_history.py
```diff
@@ -27,8 +38,14 @@ def __init__(self):
     """Initializes event data."""
     super(ChromeHistoryFileDownloadedEventData, self).__init__(
         data_type=self.DATA_TYPE)
+    self.danger_type = None
+    self.end_time = None
```
joachimmetz, line 42: let's make these events and represent them with the correct dfDateTime object

plaso/parsers/sqlite_plugins/chrome_history.py
joachimmetz: happy to provide more detail in a chat if that helps to clarify the current approach

joachimmetz: I assume you wish to track the association (https://github.com/log2timeline/plaso/issues/3566) between the 2 events and represent this as part of the output?

plaso/parsers/sqlite_plugins/chrome_history.py
obsidianforensics: If we go down that road, what do you think for the attributes then for each event? I see two approaches:

**Option 1**: Make two new events: `chrome:history:file_download:start` and `chrome:history:file_download:end`. Have the `:end` event contain most of the current `chrome:history:file_downloaded` attributes, as one would expect things like `received_bytes`, `state`, `full_path`, and so on to be known for a finished download, not at the start (when no bytes have been downloaded). For the `:start` event, have minimal attributes (maybe just `url` and the implicit timestamp?) 

The issue with this is when a download doesn't complete, it seems the `end_time` value in the SQLite DB is not populated, even though interesting fields like `interrupt_reason` and `state` are.  Since there isn't a timestamp, we couldn't create a `:end` record and we'd miss this info. We could flip and have the `:start` event have all the fields and the `:end` one have just a timestamp, but that seems wrong and confusing (adding future information to an event).

**Option 2**: Make two events, but have their attributes be basically the same. An `:end` record missing means the download didn't finish (and no info is missing, as it's all in the `:start` event). I feel this could be confusing for the user and easy to miss (requires the user to look for the absence of a record - and know to look for it). This also clutters the timeline with near-duplicate events for each download if the download finishes.

I personally favor Option 3: **keep it as one event** (but use dfDateTime objects for `start_time` and `end_time` so it's easier for a person to understand). 

What do you think?

obsidianforensics: Ah, just noticed your other comment on this, re: associated events. I think that would be interesting (for way more than just download events). That isn't built yet though, right? Maybe in the meantime implement Option 2 from above, then 
 "associate" them when that feature is added? 

I'm interested to see how those those events could be shown in Timesketch as well, could be pretty neat.

joachimmetz: Option 2 is mostly in line what we currently do, associated events is something we could work towards. I opt to brainstorm about this.

joachimmetz (approved):
LGTM

Issue #3611, opened 2021-05-17
asl parser change extraction warning into recovery warning


joachimmetz:
Extraction warning seen in the end-to-end tests on the dean-macbook Greendale test image.

```
********************************* Warning: 20 **********************************
           Message : last log entry offset does not match value in file header.
      Parser chain : asl_log
Path specification : type: OS, location:
                     dean-macbook.E01
                   : type: EWF
                   : type: TSK_PARTITION, location: /p2, part index: 5, start
                     offset: 0x12c06000
                   : type: APFS_CONTAINER, location: /apfs1, volume index: 0
                   : type: APFS, identifier: 901001, location:
                     /private/var/log/DiagnosticMessages/2019.10.03.asl
--------------------------------------------------------------------------------
```

Should this be a recovery warning? Since is a warning about metadata being inconsistent not about a failure during extraction.

PR #3613, opened 2021-05-17
Have ASL parser generate file creation event #3611

joachimmetz:
Have ASL parser generate file creation event #3611


Onager (changes requested):
Please add a test for an ASL log in a "dirty" state.

rgayon (approved):

joachimmetz:
> Please add a test for an ASL log in a "dirty" state.

Done

PR #3632, opened 2021-05-23
Made winreg parser more resilient to corrupt values

joachimmetz:
Made winreg parser more resilient to corrupt values


joachimmetz:
related to corrupt scenario https://github.com/libyal/libregf/blob/main/documentation/Windows%20NT%20Registry%20File%20(REGF)%20format.asciidoc#78-integer-value-data-too-small

PR #3936, opened 2021-11-22
Moved session analysis report counter to processing status

joachimmetz:


rgayon:
Sorry for the maybe naive question but is WAI that _PrintAnalysisReportsDetails gets removed?

rgayon (approved):
QQ about _PrintAnalysisReportsDetails, otherwise LGTM

joachimmetz:
@rgayon `_PrintAnalysisReportsDetails` is not used by psteal, since it does not support analysis modules, so yes. If we want to introduce that functionality, IMHO let's keep that in 1 single source file that is used by both psort and psteal

PR #3971, opened 2021-12-04
Moved event tag functionality into analysis mediator

joachimmetz:


joachimmetz:
this change is not feasible, closing without merging

PR #4144, opened 2022-06-19
Add timezone handling in dpkg

### LINKED ISSUE #4143: dkpg parser: ignoring time zone configuration
TL;DR - the dpkg parser does not properly handle the --timezone flag

The below is a really long explanation of how I came to the above conclusion and includes a really not-fun detour revolving around the "etcetera" time zones.

**Description of problem:**

I was using plaso to create a timeline from multiple logs. Specifically: dpkg and syslog.
I noticed that the dpkg logs were not timelined correctly because dpkg logs have no timezone (and syslog does).
I discovered the --timezone flag but that didn't resolve the issue.

At this point I have taken a deep dive into the issue.
Here's what I have: 

I'm using the following sample logs for testing: [logs.zip](https://github.com/log2timeline/plaso/files/8936136/logs.zip)
They use different parsers: dpkg, sophos_av, syslog, vsftpd
All 4 logs are set to 11:00am. syslog logs contain the timezone, and I have set it to UTC-1.
The expectation is that any log's timestamp that is correctly adjusted will be 12:00 UTC.
I ran log2timeline.py --storage_file standard.plaso logs and then psort standard.plaso -w standard.csv
I got the expected output: syslog at 12:00 and the others at 11:00.

When I added --timezone Etc/GMT-1 things started getting interesting.
Syslog: remained at 12:00Z
dpkg: remained at 11:00Z (unexpected, should be +1 hour)
sophos_av and vsftpd: fell back to 10:00Z (unexpected, should be +2 hours)

I ran another experiment where I changed the timezone in syslog to GMT+1, and passed Etc/GMT+1 to log2timeline.
Syslog: was now 10:00Z (correct)
dpkg: remained at 11:00Z (unexpected, should be -1 hour)
sophos_av and vsftpd: advanced to 12:00Z (unexpected, should be -2 hours)

So it seems there are 2 bugs:
1) Timezone adjustments are inverted (unless I'm misunderstanding how to use the feature).
2) The dpkg parser is ignoring the timezone argument outright.

For issue 1, ~~I don't want to dig further into the code until it is confirmed that this behaviour is not intended.~~ I kept digging and the behaviour comes via pytz whose behaviour comes specifically from the IANA timezone database. The etc timezones are actually the opposite of what one would expect. Ex. Etc/GMT-4 *adds* 4 hours to UTC. https://data.iana.org/time-zones/tzdb/etcetera https://bugs.launchpad.net/pytz/+bug/1777308

For issue 2, I will submit a PR so that dpkg does what sophos and vsftp do.

**Plaso version:**

20220428

**Operating system Plaso is running on:**

Ubuntu under WSL

**Installation method:**

pip install

**Debug output/tracebacks:**

[debug_logs.zip](https://github.com/log2timeline/plaso/files/8936226/debug_logs.zip)

plague006:

dpkg currently ignores the --timezone flag. 

## Description:

These changes mirror code seen in sophos_av [here](https://github.com/log2timeline/plaso/blob/720d553af5b7f4c7ad4e35dcd2a71a802608dbfa/plaso/parsers/sophos_av.py#L89) and [here](https://github.com/log2timeline/plaso/blob/720d553af5b7f4c7ad4e35dcd2a71a802608dbfa/plaso/parsers/sophos_av.py#L101) and vsftpd [here](https://github.com/log2timeline/plaso/blob/720d553af5b7f4c7ad4e35dcd2a71a802608dbfa/plaso/parsers/vsftpd.py#L95) and [here](https://github.com/log2timeline/plaso/blob/720d553af5b7f4c7ad4e35dcd2a71a802608dbfa/plaso/parsers/vsftpd.py#L106) to have dpkg properly apply the time zone.

Fixes #4143

## Checklist:
  - [X] No new [new dependencies](https://plaso.readthedocs.io/en/latest/sources/developer/Adding-a-new-dependency.html) are required or l2tdevtools has been updated
  - [x] Reviewer assigned

### DIFF
--- a/plaso/parsers/dpkg.py
+++ b/plaso/parsers/dpkg.py
@@ -154,6 +154,7 @@ class DpkgParser(text_parser.PyparsingSingleLineTextParser):
 
       date_time = dfdatetime_time_elements.TimeElements(time_elements_tuple=(
           year, month, day_of_month, hours, minutes, seconds))
+      date_time.is_local_time = True
 
     except (TypeError, ValueError):
       parser_mediator.ProduceExtractionWarning(
@@ -169,7 +170,9 @@ class DpkgParser(text_parser.PyparsingSingleLineTextParser):
     event_data.body = body_text
 
     event = time_events.DateTimeValuesEvent(
-        date_time, definitions.TIME_DESCRIPTION_ADDED)
+        date_time, definitions.TIME_DESCRIPTION_ADDED,
+        time_zone=parser_mediator.timezone
+        )
     parser_mediator.ProduceEventWithEventData(event, event_data)
 
   def VerifyStructure(self, parser_mediator, line):
--- a/plaso/parsers/dpkg.py
+++ b/plaso/parsers/dpkg.py
@@ -171,8 +171,7 @@ class DpkgParser(text_parser.PyparsingSingleLineTextParser):
 
     event = time_events.DateTimeValuesEvent(
         date_time, definitions.TIME_DESCRIPTION_ADDED,
-        time_zone=parser_mediator.timezone
-        )
+        time_zone=parser_mediator.timezone)
     parser_mediator.ProduceEventWithEventData(event, event_data)
 
   def VerifyStructure(self, parser_mediator, line):


joachimmetz (approved):
LGTM

PR #4168, opened 2022-07-21
Add parser for PowerShell transcript files

FabFaeb:

This PR adds a parser for PowerShell transcript files.

## Description:

If enabled (e.g. via [Start-Transcript](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.host/start-transcript)) PowerShell transcript files record all or part of a PowerShell session to a text file.
This PR adds a parser for these transcript files. 
Unfortunately the format can vary depending on system language and configuration. 
I tried to write this parser as general as possible and tested it with transcripts from different systems.

## Notes:
review](https://github.com/log2timeline/l2tdocs/blob/main/process/Code%20review%20process.asciidoc). This makes sure

## Checklist:
  - [x] Reviewer assigned


data/formatters/generic.yaml
```diff
@@ -864,3 +864,14 @@ type: 'basic'
 data_type: 'zeitgeist:activity'
 message: '{subject_uri}'
 short_message: '{subject_uri}'
+---
+type: 'conditional'
+data_type: 'ps:transcript:event'
```
jleaniz, line 869: We should rename 'ps' to 'powershell' to remove any ambiguity.

data/sources.config
```diff
@@ -110,6 +110,7 @@ plist:key	PLIST	Plist Entry
 PLSRecall:event	PLSRecall	PL/SQL Developer Recall file
 popularity_contest:log:event	LOG	Popularity Contest Log
 popularity_contest:session:event	LOG	Popularity Contest Session
+ps:transcript:event LOG PowerShell Transcript Event
```
jleaniz, line 113: We should rename 'ps' to 'powershell' to remove any ambiguity.

plaso/parsers/__init__.py
```diff
@@ -45,6 +45,7 @@
 from plaso.parsers import plist
 from plaso.parsers import popcontest
 from plaso.parsers import pls_recall
+from plaso.parsers import ps_transcript
```
jleaniz, line 48: We should rename 'ps' to 'powershell' to remove any ambiguity.
Maybe powershell_transcript ?

plaso/parsers/ps_transcript.py
```diff
@@ -0,0 +1,213 @@
...
+      remote management protocol version.
+    runas_user (str): User context of execution.
+    serialization_version (str): Serialization method version.
+    username (str): User that executed command.
+    version (str): PowerShell version.
+    ws_man_stack_version (str): WS-Management stack version
+    username (str): User that executed command.
+  """
+
+  DATA_TYPE = 'ps:transcript:event'
```
jleaniz, line 41: We should rename 'ps' to 'powershell' to remove any ambiguity.

plaso/parsers/ps_transcript.py
```diff
@@ -0,0 +1,213 @@
...
+    self.serialization_version = None
+    self.username = None
+    self.version = None
+    self.ws_man_stack_version = None
+
+
+class PSTranscriptParser(interface.FileObjectParser):
+  """Parses events from PowerShell transcript files."""
+
+  NAME = 'ps_transcript'
```
jleaniz, line 67: ps ==> powershell

plaso/parsers/ps_transcript.py
```diff
@@ -0,0 +1,213 @@
...
+                    11:'compatible_versions',
+                    12:'build_version',
+                    13:'clr_version',
+                    14:'ws_man_stack_version',
+                    15:'remoting_protocol_version',
+                    16:'serialization_version'}
+
+  def CreateEventsFromTranscripts(self, transcripts, parser_mediator,
+   header_data):
+    """ Get commandlines from transcripts
```
jleaniz, line 95: Remove extra whitespace after comment quotes.

plaso/parsers/ps_transcript.py
```diff
@@ -0,0 +1,213 @@
...
+          and other components, such as storage and dfvfs.
+      header_data (PSTranscriptEventData): prefilled PSTranscriptEventData
+          object with metainfos.
+    """
+
+    for transcript in transcripts:
+      # create a copy of the pre-filled event object
+      event_data = copy.deepcopy(header_data)
+      event = None
+      timestamp_pattern = re.compile(r": (\d{14})$")
```
jleaniz, line 110: Per style guide, should use ' instead of " to quote strings.

tests/parsers/ps_transcript.py
```diff
@@ -0,0 +1,71 @@
...
+    # Test a regular event.
+    expected_event_values = {
+        'build_version': '10.0.17763.1852',
+        'clr_version': '4.0.30319.42000',
+        'command': 'PS C:\\Windows\\system32> whoami; msedgewin10\\ieuser',
+        'compatible_versions': '1.0, 2.0, 3.0, 4.0, 5.0, 5.1.17763.1852',
+        'date_time': '2022-07-21 02:37:49',
+        'data_type': 'ps:transcript:event',
+        'edition': 'Desktop',
+        'host_application': 'C:\\Windows\\System32\\WindowsPowerShell\\v1.0'\
```
jleaniz, line 44: Per style guide, if possible please avoid using \ to split strings.
Maybe try this, or similar:
```
'host_application': 
    'C:\\Windows\\System32\\WindowsPowerShell\\v1.0'\\powershell.exe',
```

plaso/parsers/ps_transcript.py
```diff
@@ -0,0 +1,213 @@
...
+      line_bytes = file_bytes.split(b'\n')
+    lines = [line.decode('utf-8') for line in line_bytes]
+    sep_indices = []
+    header_data = PSTranscriptEventData()
+    for index, line in enumerate(lines):
+      if index in self.LINE_INFO_DICT:
+        param = self.LINE_INFO_DICT[index]
+        splitted = line.split(': ')
+        if len(splitted) != 2:
+          parser_mediator.ProduceExtractionWarning('line {0!s} seems invalid '\
```
jleaniz, line 188: Per style guide, if possible please avoid using \ to split strings.

plaso/parsers/ps_transcript.py
```diff
@@ -0,0 +1,213 @@
...
+    self.process_id = None
+    self.remoting_protocol_version = None
+    self.runas_user = None
+    self.serialization_version = None
+    self.username = None
+    self.version = None
+    self.ws_man_stack_version = None
+
+
+class PSTranscriptParser(interface.FileObjectParser):
```
jleaniz, line 64: One question about the choice of interface here. Is there a particular reason you are not using one of the text_parser classes (e.g. PyparsingMultiLineTextParser) and a pyparsing grammar? Transcript files seem to be UTF-8 encoded text files. It might make it easier to parse and verify the multi line transcript events.

plaso/parsers/ps_transcript.py
```diff
@@ -0,0 +1,213 @@
...
+    ws_man_stack_version (str): WS-Management stack version
+    username (str): User that executed command.
+  """
+
+  DATA_TYPE = 'ps:transcript:event'
+
+  def __init__(self):
+    """Initializes event data."""
+
+    super(PSTranscriptEventData, self).__init__(data_type=self.DATA_TYPE)
```
jleaniz, line 46: PS ==> PowerShell

plaso/parsers/ps_transcript.py
jleaniz: I think the example in the documentation uses interface.FileObjectParser because the example deals with a binary parser. In this case, i think a pyparsing parser makes more sense. @joachimmetz thoughts?

jleaniz (changes requested):
@FabFaeb thanks for your contribution and the new parser!. I left some comments for you. I would like to hear your thoughts regarding the parser interface that was used. Other than that, some minor style guide comments.

FabFaeb:
Hi @jleaniz - thanks for your fast review! 
I took care of your style guide comments and hopefully fixed all occurrences. If you happen to find any more: don't hesitate to give me a hint.

Regarding the parser interface:
I experimented with the `PyparsingMultiLineTextParser` in the beginning but was not able to make it work. My problem was that the structure of the transcript files tends to vary quite a bit. E.g. the detection of "is this line still a command or not" seemed very complicated with the TextParsers since these lines can contain basically everything - but that may also be due to me being inexperienced with writing Plaso Parsers...
In the end, I felt that it was easier to map the lines to their respective field when being able to "view" the whole file at once which is why I chose to use the `FileObjectParser`.

jleaniz:
Thanks for addressing my comments. Regarding the choice of parser interface, I think that's a fair reason to use ```FileObjectParser```.

jleaniz (approved):

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+
+    data = file_object.read(self._HEADER_READ_SIZE)
+    if not data == b'\xef\xbb\xbf**********************':
+      raise errors.WrongParser('Not a PowerShell transcript file')
+
+    # The current offset of the file-like object needs to point at
+    # the start of the file to parse the data correctly.
+    file_object.seek(0, os.SEEK_SET)
+    file_bytes = file_object.read()
```
jleaniz, line 175: @FabFaeb Could this be a problem if the file is quite large in size? Would it be better to read the file in chunks or using a memory-mapped object?

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,214 @@
...
+    # If transcript file was opened in editor, line breaks may have been changed
+    if b'\r\n' in file_bytes:
```
jleaniz, line 177: ```b'\r\n'``` seems to be used in several places throughout the method, could we use a variable name like _NEWLINE or similar?

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,214 @@
...
+      line_bytes = file_bytes.split(b'\r\n')
+    else:
+      line_bytes = file_bytes.split(b'\n')
+    lines = [line.decode('utf-8') for line in line_bytes]
+    sep_indices = []
```
jleaniz, line 182: ```sep_indices``` could we choose a more descriptive name for this variable?  Looks like maybe ```newline_indices``` might be a better name for this?

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,214 @@
...
+            self.LINE_INFO_DICT[index]))
+          continue
+        # pre-fill data for later objects
+        setattr(header_data, param, splitted[1])
+      if line == self.SEPARATOR:
+        sep_indices.append(index)
+    # the file always starts with a transcript
+    transcript_starts = [0]
+    for i, sep_index in enumerate(sep_indices):
+      # every second separator marks a new transcript
```
jleaniz, line 200: What if a transcript contains ':' in the text? Would this assumption still hold?

test_data/powershell_transcript.txt
```diff
@@ -0,0 +1,48 @@
+﻿**********************
```
jleaniz, line 1: If possible, it may be good to add some of the edge cases/weirder examples of transcripts that you mentioned.

jleaniz (changes requested):
Some additional comments/questions.

plaso/parsers/powershell_transcript.py
FabFaeb: Fixed in [1c5ce11](https://github.com/log2timeline/plaso/pull/4168/commits/1c5ce112ec91b6392e2ed80257f6c31c8bd36ca3)

plaso/parsers/powershell_transcript.py
FabFaeb: Fixed in [1c5ce11](https://github.com/log2timeline/plaso/pull/4168/commits/1c5ce112ec91b6392e2ed80257f6c31c8bd36ca3)

plaso/parsers/powershell_transcript.py
FabFaeb: A ':' in a transcript should not change anything about this assumption as far as I can see. The ':' character is only used for splitting key and value in the assumed header lines. If you still think that this is an issue, could you give an example?

test_data/powershell_transcript.txt
FabFaeb: How should I add those? Can I just commit a `powershell_transcript_2.txt` to the test_data folder? I was worried that only one example per parser is allowed.

plaso/parsers/powershell_transcript.py
FabFaeb: Personally, I have never seen a transcript file that was larger than a few MBs - but I can see your point.

I have thought about the reading in chunks, but I found that the unpredictable log format is again a problem. There are no save "boundaries" on which I could separate the chunks. So, very likely, part 1 of a transcript would be read in chunk 1 and part 2 in chunk 2. This would make the parsing of the file into separate "transcripts" (which are then used to generate the events) a whole lot more complicated.

The file_object that is handed to this method is a `dfvfs.FileIO` type. I have not been able to find an example how I could use this as mmapped object. Do you have any reference/example how I could integrate this here?

The simplest (although not the most beautiful) fix I could think of would be to check the size of the file_object and just stop parsing if it is larger than XXX MB - just to have a fail-safe in place.

What would be your favorite approach here?

FabFaeb:
@jleaniz thanks again for your review! 
I implemented a few of your comments and elaborated on the other ones directly under your comment.

test_data/powershell_transcript.txt
Onager: You can add as many source files as you need to test the parser - +1 to having some edge cases, it would be especially helpful to have an example in a non-English locale

plaso/parsers/powershell_transcript.py
Onager: Unfortunately, I think the ideal fix here is more structural - inheriting from PyParsing multi- or single- line parser, and using the existing machinery we have for handling text files in general.

@FabFaeb Could you take a look at that and see what you think? You could take a look at the SCCM parser for some inspiration

Onager: I had a few minutes, so I started poking at what the pyparsing grammar might look like - https://gist.github.com/Onager/b0162b2c1753e88ffa91923c8a254315 might also provide some inspiration

Onager (changes requested):
Per other comments,:

- please look at using pyparsingmultilineparser to implement this parser
- please add some more test cases, particularly a localized version of the log file, as well as one with multiple commands in the transcript
- Consider this adding this parser to default windows preset, in data/presets.yaml for "win7"

FabFaeb:
@Onager thanks for your review. As I commented a while ago I tried to inplement this parser initially with the multilineparser but couldn't quite make it work.
 I will take a look at the multilineparser once again and see if I can find a way to get it working this time. Thanks also for your gist - if that's fine I would just reach out in case of further questions.
I will also happily provide more sample data for testing.

Unfortunately, all of this might take some time - so if you prefer that, I could open another PR once I am done.

Onager:
Apologies for missing the earlier discussion about using pyparsing - I'll take a look and see if I can spot any solutions. Definitely feel free to reach out directly if you have questions.

FabFaeb:
Hi @Onager  and @jleaniz,
I found some time to look at this again and managed to rewrite the parser based on pyparsingmultilineparser.
Please take a look when you have some time and let me know what you think.

A few preliminary remarks:
- Unfortunately, it proved pretty hard to write a pyparsing grammar that would parse all lines correctly by itself. You can see an example for this in the new `powershell_transcript_ger.txt` test file: While line 35 is actually a `transcript_line`, it looks like a `metadata_line` for the parser. For this reason I had to add some logic to "re-classify" lines in some cases (see lines 309ff. in the parser).
- Per line 300ff. in the parser, the last line of a transcript will currently not be parsed correctly, if this transcript is not closed cleanly (which it will be most of the time). To do this correctly, the parser would need to know, when it got the last line of a text file. I have not found a way to accomplish this - maybe one of you has an idea?

I'm looking forward to your review!

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+
+  # key always start with an uppercase character
+  _METADATA_KEY = (pyparsing.Word(pyparsing.alphas.upper(), ' ' +
+    pyparsing.alphas) + _COLON)
+  # negative lookahead needed for correct identification
+  _METADATA_VALUE = (~_METADATA_KEY +
+    pyparsing.Word(pyparsing.printables, ' ' + pyparsing.printables))
+
+  _METADATA_LINE = (
+    _METADATA_KEY +
```
sydp, line 84: style nit - in these cases where you have an open parenthesis, use a hanging indent of 4 spaces

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+
+  def _GetTimestampFromString(self, value):
+    """Parse a timestamp string an return a TimeElements event.
+
+    Args:
+      value (str): String containing a timestamp.
+
+    Returns:
+      dfdatetime_time_elements.TimeElements: TimeElements, if string
+        was valid, None otherwise
```
sydp, line 124: use a 4 space hanging indent

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+    """
+    if len(value) == 14:
+      time_elements_tuple = int(value[:4]),\
```
sydp, line 127: style nit - use parenthesis around these instead of backslash line continuation and use appropriate indentation.

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+                              int(value[4:6]),\
+                              int(value[6:8]),\
+                              int(value[8:10]),\
+                              int(value[10:12]),\
+                              int(value[12:])
+      try:
+        start_time = dfdatetime_time_elements.TimeElements(
+          time_elements_tuple=time_elements_tuple)
```
sydp, line 135: style nit - 4 space hanging indent

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+    self.key_structure.clear()
+
+    new_event = PowerShellTranscriptEventData()
+    transcript_text = ''
+    metadata_lines = []
+    event_time = None
+    for key, structure in key_structure:
+      if key == 'transcript_line':
+        transcript_text = transcript_text + structure + '; '
+      if key == 'metadata_line':
```
sydp, line 167: elif?

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+
+    Args:
+      parser_mediator (ParserMediator): mediates interactions between parsers
+          and other components, such as storage and dfvfs.
+      lines (str): one or more lines from the text file.
+
+    Returns:
+      bool: True if this is the correct parser, False otherwise.
+    """
+    if lines[:25] == 'ï»¿' + self._SEPARATOR:
```
sydp, line 332: use a class constant _ENCODING = 'utf-8-sig' to avoid the BOM check here

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+      if 'Windows PowerShell' in lines:
```
sydp, line 333: rather than nesting could you AND these checks?

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+  _METADATA_LINE = (
+    _METADATA_KEY +
+    _COLON +
+    pyparsing.Optional(_METADATA_VALUE) +
+    pyparsing.LineEnd().suppress()
+  )
+
+  _TRANSCRIPT_LINE = (
+    ~_METADATA_LINE +
+    pyparsing.Word(' ' + '\t' + '\r' +
```
sydp, line 92: style nit - use vertical alignment

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+
+  _METADATA_LINE = (
+    _METADATA_KEY +
+    _COLON +
+    pyparsing.Optional(_METADATA_VALUE) +
+    pyparsing.LineEnd().suppress()
+  )
+
+  _TRANSCRIPT_LINE = (
+    ~_METADATA_LINE +
```
sydp, line 91: style nit - 4 space hanging indent

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+    pyparsing.Word(' ' + '\t' + '\r' +
+      pyparsing.printables + pyparsing.alphas8bit +
+      pyparsing.punc8bit, exclude_chars='**********************') +
+    pyparsing.LineEnd().suppress()
+  )
+
+  _SEPARATOR_LINE = (
+    pyparsing.Literal(_SEPARATOR)
+    + pyparsing.LineEnd().suppress()
```
sydp, line 100: i opt to move the "+" to the line above

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+          'since transcript text could not be set'.format(key_structure))
+      return False
+    new_event.command = transcript_text
+    # create event if at least the most relevant fields are set
+    if (event_time is not None and new_event.command is not None
+      and new_event.command != ''):
+      parser_mediator.ProduceEventWithEventData(event_time, new_event)
+    else:
+      parser_mediator.ProduceExtractionWarning('skipping transcript {0!s} - '
+        'relevant event values could not be set'.format(key_structure))
```
sydp, line 268: style nit - this line should be indented so it is vertically aligned, but a better layout might be hanging indents.

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+        definitions.TIME_DESCRIPTION_START)
+
+    if transcript_text == '':
+      parser_mediator.ProduceExtractionWarning('skipping transcript {0!s} - '
+          'since transcript text could not be set'.format(key_structure))
+      return False
+    new_event.command = transcript_text
+    # create event if at least the most relevant fields are set
+    if (event_time is not None and new_event.command is not None
+      and new_event.command != ''):
```
sydp, line 264: style nit - indent to align with event_time

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+      # pyparsing.StringEnd() did not match correctly
+      # __del__ will not be called in regular use cases
+      # tl;dr I currently have no idea how to fix this
+      self._CreateEventFromTranscript(parser_mediator)
+    else:
+      # overwrite key, since a transcript_line can never be followed directly
+      # by a metadata_line (and vice versa) and the pyparsing could be wrong
+      # (since a transcript_line can basically contain everything)
+      if (key == 'metadata_line' and len(self.key_structure) > 0
+        and self.key_structure[-1][0] == 'transcript_line'):
```
sydp, line 314: style nit - add spaces to align here and below

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+  """Tests for the PowerShell Transcript parser"""
+
+  def testParseEng(self):
+    """Tests the parse function on an example file with english locale."""
+    parser = powershell_transcript.PowerShellTranscriptParser()
+    storage_writer = self._ParseFile(
+        ['powershell_transcript.txt'], parser)
+
+    number_of_events = storage_writer.GetNumberOfAttributeContainers(
+      'event')
```
sydp, line 23: add 2 spaces

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+    expected_event_values = {
+        'build_version': '10.0.17763.1852',
+        'clr_version': '4.0.30319.42000',
+        'command': 'PS C:\\Windows\\system32> whoami; msedgewin10\\ieuser; ',
+        'compatible_versions': '1.0, 2.0, 3.0, 4.0, 5.0, 5.1.17763.1852',
+        'date_time': '2022-07-21 02:37:49',
+        'data_type': 'powershell:transcript:event',
+        'edition': 'Desktop',
+        'host_application':
+          'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
```
sydp, line 45: add 2 spaces

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+    del parser
+
+  def testParseGer(self):
+    """Tests the parse function on an example file with german locale."""
+    parser_ger = powershell_transcript.PowerShellTranscriptParser()
+    storage_writer_ger = self._ParseFile(
+        ['powershell_transcript_ger.txt'], parser_ger)
+
+    number_of_events = storage_writer_ger.GetNumberOfAttributeContainers(
+      'event')
```
sydp, line 65: add 2 spaces

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+
+    number_of_warnings = storage_writer_ger.GetNumberOfAttributeContainers(
+        'recovery_warning')
+    self.assertEqual(number_of_warnings, 0)
+
+    events_ger = list(storage_writer_ger.GetSortedEvents())
+    expected_event_values_ger_0 = {
+        'build_version': '10.0.19041.1682',
+        'clr_version': '4.0.30319.42000',
+        'command': 'Die Aufzeichnung wurde gestartet.'+
```
sydp, line 82: use parenthesis instead of concatenation

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+          ' STRG-C; PS C:\\Users\\User> TerminatingError():'+
+          ' "Die Pipeline wurde beendet."; >> TerminatingError():'+
+          ' "Die Pipeline wurde beendet."; PS C:\\Users\\User>'+
+          ' Get-Content .\\myfile.txt; VERBOSE: Simple Test; ',
+        'compatible_versions': '1.0, 2.0, 3.0, 4.0, 5.0, 5.1.19041.1682',
+        'date_time': '2022-08-24 12:21:11',
+        'data_type': 'powershell:transcript:event',
+        'edition': 'Desktop',
+        'host_application':
+          'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
```
sydp, line 100: add 2 spaces

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+        'process_id': '18716',
+        'remoting_protocol_version': '2.3',
+        'runas_user': 'DE\\User',
+        'serialization_version': '1.1.0.1',
+        'username': 'DE\\User',
+        'version': '5.1.19041.1682',
+        'ws_man_stack_version': '3.0'}
+
+    self.CheckEventValues(storage_writer_ger, events_ger[0],
+      expected_event_values_ger_0)
```
sydp, line 111: vertical alignment

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+
+    expected_event_values_ger_1 = {
+        'build_version': '10.0.19041.1682',
+        'clr_version': '4.0.30319.42000',
+        'command': 'PS C:\\Users\\User\\Documents>'+
```
sydp, line 116: use parenthesis instead of concatenation

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+          ' Ping-Statistik fÃ¼r 127.0.0.1:; Pakete: Gesendet = 5,'+
+          ' Empfangen = 5, Verloren = 0; (0% Verlust),; Ca.'+
+          ' Zeitangaben in Millisek.:; Minimum = 0ms, Maximum = 1ms,'+
+          ' Mittelwert = 0ms; ',
+        'compatible_versions': '1.0, 2.0, 3.0, 4.0, 5.0, 5.1.19041.1682',
+        'date_time': '2022-08-24 12:31:14',
+        'data_type': 'powershell:transcript:event',
+        'edition': 'Desktop',
+        'host_application':
+          'C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe',
```
sydp, line 133: add 2 spaces

tests/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,161 @@
...
+        'process_id': '18716',
+        'remoting_protocol_version': '2.3',
+        'runas_user': 'DE\\User',
+        'serialization_version': '1.1.0.1',
+        'username': 'DE\\User',
+        'version': '5.1.19041.1682',
+        'ws_man_stack_version': '3.0'}
+
+    self.CheckEventValues(storage_writer_ger, events_ger[1],
+      expected_event_values_ger_1)
```
sydp, line 144: add spaces to vertically align

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+
+    Returns:
+      bool: True if this is the correct parser, False otherwise.
+    """
+    if lines[:25] == 'ï»¿' + self._SEPARATOR:
+      if 'Windows PowerShell' in lines:
+        return True
+    return False
+
+  def __del__(self):
```
sydp, line 337: i would reconsider your design here -firstly please note this method is called when the object is about to be destroyed.  secondly, i'm not really sure why your unit tests would need to do this.

https://docs.python.org/3/reference/datamodel.html#object.__del__

sydp (changes requested):
Some initial thoughts, mainly on general style.

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,343 @@
...
+      if key == 'transcript_line':
+        transcript_text = transcript_text + structure + '; '
+      if key == 'metadata_line':
+        metadata_lines.append(structure)
+    # transcript has a full header
+    # that means we can try to extract all header infos
+    if len(metadata_lines) > 1:
+      # counter needed for identification of line
+      counter = 0
+      for metadata in metadata_lines:
```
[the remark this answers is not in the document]
FabFaeb: I don't think this is possible here, since the pyparsing logic needs to be overwritten sometimes as in lines 333 and 336

test_data/powershell_transcript.txt
FabFaeb: Added another example files and a few test  cases.

FabFaeb:
Hi @sydp ,
I just implemented most of the changes you suggested. 
The `__del__` is indeed no longer needed after I switched to instance variables and the since the file is now parsed with the right encoding, special characters are shown correctly now - thanks for the pointing me in the right direction!

joachimmetz:
@Onager @FabFaeb what is the status of this PR?

Onager:
@joachimmetz This is waiting for me, I started reviewing this the other day but ran into some format issues

joachimmetz:
@Onager this PR needs to be brought up to date as well.

test_data/powershell_transcript_ger.txt
```diff
@@ -0,0 +1,56 @@
+﻿**********************
+nStart der Windows PowerShell-Aufzeichnung
```
joachimmetz, line 2: is this leading `n` part of the log file?

plaso/parsers/powershell_transcript.py
joachimmetz: the current logic checks the same thing multiple times

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,356 @@
...
+
+    Raises:
+      ParseError: when the structure type is unknown.
+    """
+    # skip the very first (separator) line since it does not contain any
+    # relevant info and will break our logic
+    if self.is_first_line:
+      self.is_first_line = False
+      return
+    # the first line of a transcript will always be a metadata line, however
```
joachimmetz, line 302: what is the first line of a transcript in your perspective? is that the line after the second `**********************` ?

joachimmetz: or is that e.g. `Windows PowerShell transcript start` ?

joachimmetz: I'll assume the latter

test_data/powershell_transcript_ger.txt
```diff
@@ -0,0 +1,56 @@
...
+
+Ping-Statistik für 127.0.0.1:
+    Pakete: Gesendet = 5, Empfangen = 5, Verloren = 0
+    (0% Verlust),
+Ca. Zeitangaben in Millisek.:
+    Minimum = 0ms, Maximum = 1ms, Mittelwert = 0ms
+**********************
+Befehlsstartzeit: 20220824124237
+**********************
+PS C:\Users\User> rm .\filename
```
joachimmetz, line 56: Was this file truncated or is the footer missing accurate?

plaso/parsers/powershell_transcript.py
```diff
@@ -0,0 +1,356 @@
...
+  MAX_LINE_LENGTH = 65536
+
+  _ENCODING = 'utf-8-sig'
+
+  _SEPARATOR = '**********************'
+
+  _COLON = pyparsing.Literal(': ').suppress()
+
+  # key always start with an uppercase character
+  _METADATA_KEY = (pyparsing.Word(pyparsing.alphas.upper(), ' ' +
```
joachimmetz, line 78: looks like `' '` (only space) should be `'- '` (dash and space)  ?

joachimmetz:
@FabFaeb PTAL I've updated the code to work with HEAD and various clean up. Please check if the parser still works with your understanding of the format. I also left a couple of questions / comment.

joachimmetz:
@Onager PTAL from a code review perspective and seeing you appear to have looked at the format a bit

FabFaeb:
Hi @joachimmetz,
thanks for taking a look at this!
I just had a few minutes to check your changes and it seems like the parser/plugin does not work correctly anymore. While it still produces the expected amount of events, the events all seem to be duplicates of the last entry in the transcript. E.g. for the provided `powershell_transcript_ger.txt` example, the `...ping 127.0.0.1 -n 5...` event/entry is created two times. 

Unfortunately, I don't have too much time to look deeper into this right now. Maybe you have a quick idea about this issue? If I have some time at my hands in the coming weeks, I'll also try to debug this further and report back here.

I'll answer to your inline comments in a minute.

test_data/powershell_transcript_ger.txt
FabFaeb: The missing footer is accurate. If a session/transcript is ended unexpectedly, there will be no footer as far as I can tell.

plaso/parsers/powershell_transcript.py
FabFaeb: I'd also vote for the latter.

test_data/powershell_transcript_ger.txt
FabFaeb: I'll retest this on one of my Windows Machines and report back later.

joachimmetz: thanks

test_data/powershell_transcript_ger.txt
joachimmetz: thanks for confirming

joachimmetz:
re-added the use of copy.deepcopy of the event_data object

FabFaeb:
With your last commit, the events are created correctly again!

I found that you removed my `TODO` comment regarding the last entry in a transcript that was not exited cleanly 
https://github.com/log2timeline/plaso/blob/15de114d41a3390dcfabf042fc8993313ed0c7e7/plaso/parsers/powershell_transcript.py#L318
However, I think that this is still an open issue, right? I stumbled upon this due to the still existing comment in the test file.
https://github.com/log2timeline/plaso/blob/e8c36a99e7ce727e03ec6ab1a30978f82b21cc13/tests/parsers/text_plugins/powershell_transcript.py#L61
In short: if a transcript is exited suddenly, and no footer info is added, currently the last entry is not parsed. E.g. in the `powershell_transcript_ger.txt`, the last entry `PS C:\Users\User> rm .\filename` is not added as an event, since the finalizing separator line is not there. This is also the case in the most recent version.

As I wrote in my original comment, I think we needed to know that the currently parsed line is the last line in the file, so that we could wrap up the last event. Do you see any way to do this or do you maybe have another idea how to fix this?
Or do we have to just accept this limitation?

Besides that, now everything seems to be working as expected to me.

joachimmetz:
> I found that you removed my TODO comment regarding the last entry in a transcript that was not exited cleanly

I replaced it with https://github.com/log2timeline/plaso/pull/4168/files#diff-d263ca460ae4064ef96c2981245da460bd3a4cf0438265403bb7715a14a305e3R128 but needed to know some more context first. I might have an idea to add support for it be as processing warnings to begin with.

joachimmetz:
Note to self add footer support / incomplete file warning

PR #4191, opened 2022-08-10
Changed winevtx parser to extract event version

joachimmetz:


joachimmetz:
@marcobrotto FYI

marcobrotto:
> @marcobrotto FYI

Great!!! Thank you for the prompt reply and quick fix!

PR #4226, opened 2022-09-05
Handle negative fraction of seconds in non-dynamic time #4225

joachimmetz:


plaso/output/formatting_helper.py
```diff
@@ -138,8 +138,17 @@ def _FormatDateTime(
...
+      if fraction_of_second >= 0:
+        while fraction_of_second > 1000000:
+          fraction_of_second, _ = divmod(fraction_of_second, 10)
+
+      else:
+        while fraction_of_second < -1000000:
+          fraction_of_second, _ = divmod(fraction_of_second, 10)
+
+        number_of_seconds -= 1
+        fraction_of_second += 1000000
```
joachimmetz, line 151: Note to self what about fraction of second > -1000000 < 0

PR #4325, opened 2022-10-18
Changed Dockerfile to use single RUN statement for similar commands

fbausch:
This commit reduces the size of plaso docker images and implements best practices for Dockerfiles.

## Description:

This commit merges subsequent apt calls into one RUN command in the Dockerfile. This reduces the number of layers created and the total size of the container image:

```
REPOSITORY           TAG       IMAGE ID       CREATED             SIZE
log2timeline/plaso   test      873096d9c022   20 minutes ago      410MB
log2timeline/plaso   latest    51cc3cae181f   26 minutes ago      453MB
```

(`test` is after changing the Dockerfile.)

Furthermore, the deprecated MAINTAINER statement is replaced by the recommended LABEL statement.

The problems were identified i.a. by using `trivy config Dockerfile`. (see https://github.com/aquasecurity/trivy)

The commit does not change the logic or order of the commands in the Dockerfile.

## Checklist:
  - [ ] Reviewer assigned

### DIFF
--- a/config/docker/Dockerfile
+++ b/config/docker/Dockerfile
@@ -13,21 +13,18 @@ ARG PPA_TRACK=stable
 
 ENV DEBIAN_FRONTEND=noninteractive
 
+# Combining the apt-get commands into a single run reduces the size of the resulting image.
+# The apt-get installations below are interdependent and need to be done in sequence.
 RUN apt-get -y update && \
     apt-get -y install apt-transport-https apt-utils && \
     apt-get -y install libterm-readline-gnu-perl software-properties-common && \
-    add-apt-repository -y ppa:gift/$PPA_TRACK && \
-    \
-    apt-get -y update && \
+    add-apt-repository -y ppa:gift/$PPA_TRACK && apt-get -y update &&
     apt-get -y upgrade && \
-    \
     apt-get -y install locales plaso-tools && \
-    \
     apt-get clean && rm -rf /var/cache/apt/* /var/lib/apt/lists/*
 
 # Set terminal to UTF-8 by default
-RUN locale-gen en_US.UTF-8 && \
-    update-locale LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8
+RUN locale-gen en_US.UTF-8 && update-locale LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8
 ENV LANG en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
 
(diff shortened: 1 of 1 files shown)


joachimmetz:
Can't say I agree that this improves maintainability of the Docker file

fbausch:
> @fbausch can you point me to the rationale of the "best" practices?

Docker containers have layered filesystems and (almost) each command in a Dockerfile (RUN, COPY etc.) creates a new unchangeable layer.

If the `apt update` and the apt cache cleaning happen in separate RUN commands, the resulting docker container will be shipped with a copy of the apt cache, even though it is not visible in the docker container during runtime. But it still needs space. Cached layers can also lead to issues during container build.
The details are also described here: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
As I showed, the plaso docker image could be roughly 40MB smaller.

And in my opinion the maintainability of the Dockerfile is not affected, since it contains the same command in the same order - each command in its own line.

joachimmetz:
> And in my opinion the maintainability of the Dockerfile is not affected, since it contains the same command in the same order - each command in its own line.

No it's not, the proposed method is more prone to errors. Much easier to typo, harder to debug, bottom line I don't understand the "best" here in these practices.

Why not use `--squash` instead if the only gain is a smaller Docker image? 
https://docs.docker.com/engine/reference/commandline/image_build/

fbausch:
> No it's not, the proposed method is more prone to errors. Much easier to typo, harder to debug, bottom line I don't understand the "best" here in these practices.

The docker documentation describes it as best practice and also explains why. For example, the separate RUN commands may lead to issues caused by cached layers – which is also not easy to debug.
 
> Why not use `--squash` instead if the only gain is a smaller Docker image? https://docs.docker.com/engine/reference/commandline/image_build/

`--squash` is an experimental feature and it cannot address issues that may occur during `docker build`.

joachimmetz:
> --squash is an experimental feature and it cannot address issues that may occur during docker build.

your argument was to have a smaller image

> The docker documentation describes it as best practice and also explains why. For example, the separate RUN commands may lead to issues caused by cached layers – which is also not easy to debug.

So far the current approach has not caused any issues. Where as the proposed approached definitely has. So your argumentation is not convincing me.

I'll have a look at the LABEL statement when time permits

fbausch:
> So far the current approach has not caused any issues. Where as the proposed approached definitely has.

Please show me a prove that my proposal _definitely_ caused issues. I highly doubt that.

joachimmetz:
> Please show me a prove that my proposal definitely caused issues. I highly doubt that.

The current way you have chained the commands will allow them to fail silently.

fbausch:
I updated the merge request for not failing silently.

joachimmetz:
> I updated the merge request for not failing silently.

that is better, there are some issues with `apt-get update` erroring due to environmental factors, but that is not different from the current situation.

joachimmetz (approved):
LGTM after changes

joachimmetz:
What I said https://github.com/log2timeline/plaso/pull/4333 this solution is more error prone.

Issue #4486, opened 2022-12-14
Add support for Safari Downloads.plist


joachimmetz:
* [x] generate test data
* [x] add plist plugin and tests - https://github.com/log2timeline/plaso/pull/4491
* [x] add artifact definition - https://github.com/ForensicArtifacts/artifacts/blob/main/data/webbrowser.yaml#L1176

chb2mn:
[Downloads.zip](https://github.com/log2timeline/plaso/files/10230863/Downloads.zip)

Hey Joachim, here is a test example. lmk if you'd like something more diverse

joachimmetz:
@chb2mn thanks will take a more detailed look tomorrow

PR #4491, opened 2022-12-16
Added support for Safari Downloads.plist #4486

joachimmetz:


plaso/parsers/plist_plugins/safari_downloads.py
```diff
@@ -0,0 +1,72 @@
...
+
+  def __init__(self):
+    """Initializes event data."""
+    super(SafariDownloadEventData, self).__init__(data_type=self.DATA_TYPE)
+    self.end_time = None
+    self.full_path = None
+    self.received_bytes = None
+    self.start_time = None
+    self.total_bytes = None
+    self.url = None
```
chb2mn, line 33: Just the field I had, except I also added `removed_on_complete := event_data.DownloadEntryRemoveWhenDoneKey`

joachimmetz: https://blog.d204n6.com/2021/05/ios-macos-tracking-downloads-from.html hints that this value is used for "remove on close"

joachimmetz: I've added it, update pending shortly

PR #4581, opened 2023-03-12
Changes to Windows Registry event data values

joachimmetz:


joachimmetz:
@berggren PTAL how this affects Timesketch

PR #4637, opened 2023-04-20
Add data type counter to storage

nflexfo:

Introduces a new data type counter into storage file.

## Description:

This patch add a new `data_type_count` counter into plaso storage file.

While the `parser_count` counter does a good job to describe "how" events are generated, this new counter allows a better overview of "what" kinds of event are actually stored in the storage file (i.e. data_type).

The patches does three things:

- Add a new `DataTypeCount` counter which add a new `data_type_count` table into storage file
- The container is populated by the extraction engine
- Pinfo is modified to display the data type counter in the "event" section

This patch is a pre-requisite for an upcoming patch series focusing on **additional fields** exploitation. The later is actually only usable when events are narrowed down to a single data type. For instance, specifying `--additional_fields event_identifier,source_name` only make sense when event data type is `windows:evtx:record`. Having the data type counter in storage avoid brute-forcing all possible combinations of data_types/additional_fields.

_PS:_ It is very likely that the storage version needs to be upgraded since it might lead to inconsistency with existing storage files.

_PS-2:_ The patch has a lots of duplicate/similar code with the `parser_count`. I intentionaly did not factorize the code to ease reviewing. Let me know if you want this MR to actually include it.

## Notes:

## Checklist:
* [x] No new [new dependencies](https://plaso.readthedocs.io/en/latest/sources/developer/Adding-a-new-dependency.html) are required or l2tdevtools has been updated
* [ ] Reviewer assigned

### DIFF
--- a/plaso/cli/pinfo_tool.py
+++ b/plaso/cli/pinfo_tool.py
@@ -122,6 +122,16 @@ class PinfoTool(tools.CLITool, tool_options.StorageFileOptions):
     analysis_reports_counter = collections.Counter()
     analysis_reports_counter_error = False
 
+    data_types_counter = {}
+    if storage_reader.HasAttributeContainers('data_type_count'):
+      data_types_counter = {
+          data_type_count.name: data_type_count.number_of_events
+          for data_type_count in storage_reader.GetAttributeContainers(
+              'data_type_count')}
+
+    data_types_counter = collections.Counter(data_types_counter)
+    data_types_counter_error = False
+
     event_labels_counter = {}
     if storage_reader.HasAttributeContainers('event_label_count'):
       event_labels_counter = {
@@ -193,6 +203,9 @@ class PinfoTool(tools.CLITool, tool_options.StorageFileOptions):
 
     if not analysis_reports_counter_error:
       storage_counters['analysis_reports'] = analysis_reports_counter
+
+    if not data_types_counter_error:
+      storage_counters['data_types'] = data_types_counter
 
     if not event_labels_counter_error:
       storage_counters['event_labels'] = event_labels_counter
@@ -270,6 +283,20 @@ class PinfoTool(tools.CLITool, tool_options.StorageFileOptions):
     storage_counters = self._CalculateStorageCounters(storage_reader)
     compare_storage_counters = self._CalculateStorageCounters(
         compare_storage_reader)
+
+    # Compare number of events by data type.
+    data_types_counter = storage_counters.get('data_types', collections.Counter())
+    compare_data_types_counter = compare_storage_counters.get(
+        'data_types', collections.Counter())
+    differences = self._CompareCounter(data_types_counter, compare_data_types_counter)
+
+    if differences:
+      stores_are_identical = False
+
+      self._PrintCounterDifferences(
+          differences,
+          column_names=['Data type name', 'Number of events'],
+          title='Events generated per data type')
 
     # Compare number of events.
     parsers_counter = storage_counters.get('parsers', collections.Counter())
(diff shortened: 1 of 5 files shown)


joachimmetz:
Could you explain your use case here a bit more. I'm actually working towards storing event data more native in SQLite, which means such a report could be done with SQL instead.

nflexfo:
Hi,

Fundamentally, I try to get as much data as possible from each event. If there is any kind of attribute filled in an event data, I want it in the output. This way I can do some anomalies detection at the attribute granularity, either automatically (unsupervised) or manually (by analysts). An apparently meaningless value/attribute might as well be an IoC which I don't want to miss.

Right now, I run `log2timeline.py` to generate the Plaso storage file and then I run `psort.py` once **for each data type** present in the storage. For instance:

```
./psort.py \
    --output_time_zone UTC \
    -o dynamic \
    --data-type "windows:evtx:record" \
    --additional_fields "event_id,source_name" \
    -w "hostname-windows:evtx:record.csv" \
    hostname.plaso
```

_NOTE: The `--data-type` parameter has been added by a custom patch, more on this below_.

In the previous example, there is only two additional fields but in practice there is actually all of them. To know beforehand which additional fields exists, I maintain a `data_types.json` file which is a dictionary in the following form:

- key: data type name (e.g. `windows:evtx:record`)
- value: list of strings (all attributes of EventData, excluding the base class attributes)

The file is generated with a script very similar to `utils/export_event_data.py` with some tweak to exclude the default fields. Also, it fixes a small bug on some data types which can't be instancied directly because of mandatory constructor arguments (e.g. `cookie:google:analytics:utm*` or `windows:distributed_link_tracking:creation`). I might open an issue for these one when time permit. Also, the `export_event_data.py` script currently escapes virtualenv because of the shebang.

Now that I know which additional fields can be "asked" to `psort.py` per data type, I need to know which data types are actually present in the storage. **This is the purpose of the current MR**. Also, I do think that other people might interested with such stat.

Finally, I use a custom patch to add this `--data-type` option to `psort.py`. It is basically a **filter**. I do not use the _filter expression_ for optimization reason. Right now, event data are stored serialized and psort has to iterate over every events to perform such filtering. Since I can run `psort.py` a hundred times, it might take ages `O(nb_data_types * nb_events)` to generate the CSV files instead of `O(nb_events)`. So I basically add a `data_type TEXT` column in the `event_data` table and do a `SELECT ... FROM event_data WHERE data_type == XXX` in `GetSortedEvents()`.

Since you might be planning to modify the storage file, I would be glad to know what design do you have in mind. I agree that storing all event serialized does not allows to benefit from SQLite facilities. Also, the "data type counter" would be obsolete once the `data_type` column become a thing. Right now, this is the only column that I really need.

Let me know what you think about it.

joachimmetz:
> Fundamentally, I try to get as much data as possible from each event. If there is any kind of attribute filled in an event data, I want it in the output. 

why use dynamic output? why not JSON or rawpy instead?

nflexfo:
Well, it looks like I completly missed this possibility and had tunnel vision on CSV format... Thanks! It will actually simplify my pipeline and get rid of those patches which were time consuming to maintain.  Could you be a bit more precise about event data attributes that might be missing the json format?

As this won't be mandatory for me soon, you can close this MR. However, I still have some interests in having the data type stats in pinfo. Any ETA about the storage rewriting?

joachimmetz:
>  Could you be a bit more precise about event data attributes that might be missing the json format?

This is the most up to date explanation, https://github.com/log2timeline/plaso/blob/main/plaso/output/shared_json.py#L101 

The JSON format was never "formalized" but trying to keep it as much as the same as before. Will likely do a revamp in more distance future.

> Any ETA about the storage rewriting?

No ETA, this project is best effort. This will happen gradually.

joachimmetz:
I'll close this issue then

Issue #3806, opened 2021-07-21
Add iOS specific parsers : DataUsage.sqlite & com.apple.identityservices.idstatuscache.plist


rgayon:
Context: https://www.amnesty.org/en/latest/research/2021/07/forensic-methodology-report-how-to-catch-nso-groups-pegasus/#_ftn1

`iOS maintains records of process executions and their respective network usage in two SQLite database files called “DataUsage.sqlite” and “netusage.sqlite” which are stored on the device. `

`record of Apple IDs seen by each installed application in a plist file located at /private/var/mobile/Library/Preferences/com.apple.identityservices.idstatuscache.plist`

We already have parsers for these file formats so I suppose what we need is some test data to figure out the schema

Mulbot1:
/private/var/mobile/Library/Preferences/com.apple.identityservices.idstatuscache.plist,

rick-slin:
I'd like to work on this issue. I was  thinking of making two PRs, one per file.

joachimmetz:
Both parsers have been implemented closing issue

PR #4673, opened 2023-05-31
implemented plist plugin parser for com.apple.identityservices.idstat…

rick-slin:
implemented plist plugin parser for com.apple.identityservices.idstatuscache.plist

## Description:
implemented plist plugin parser for com.apple.identityservices.idstatuscache.plist

**Related issue (if applicable):** fixes the second part of #3806 

## Notes:

## Checklist:
* [x] Automated checks (GitHub Actions, AppVeyor) pass
* [x] No new [new dependencies](https://plaso.readthedocs.io/en/latest/sources/developer/Adding-a-new-dependency.html) are required or l2tdevtools has been updated
* [x] Reviewer assigned


joachimmetz:
@rick-slin thanks for the PR I'll try to have a look later this week

joachimmetz:
You'll need to update https://github.com/log2timeline/plaso/blob/main/data/timeliner.yaml as well

plaso/parsers/plist_plugins/ios_identityservices.py
```diff
@@ -0,0 +1,83 @@
...
+from plaso.containers import events
+from plaso.parsers import plist
+from plaso.parsers.plist_plugins import interface
+
+
+class IOSIdstatusacheEventData(events.EventData):
+  """iOS com.apple.identityservices.idstatuscache.plist event.
+
+  Attributes:
+    process_name: (str) Name of the process that looked up an identifier.
```
joachimmetz, line 16: nit: format of the docstrings is:
```
name (type): description
```

plaso/parsers/plist_plugins/ios_identityservices.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Args:
+      plist_key (object): plist key.
+      plist_value_name (str): name of the value in the plist key.
+
+    Returns:
+      dfdatetime.TimeElementsInMicroseconds: date and time or None if not
+          available.
+    """
+    timestamp = plist_key.get(plist_value_name, None)
```
joachimmetz, line 52: this will also return None if timestamp is 0 is that intentionally?

plaso/parsers/plist_plugins/ios_identityservices.py
```diff
@@ -0,0 +1,83 @@
...
+
+  # pylint: disable=arguments-differ
+  def _ParsePlist(
+      self, parser_mediator, match=None, top_level=None, **unused_kwargs):
+    """Extract Apple ids queried by process name.
+
+    Args:
+      parser_mediator (ParserMediator): mediates interactions between parsers
+          and other components, such as storage and dfVFS.
+      match (Optional[dict[str: object]]): keys extracted from PLIST_KEYS.
```
joachimmetz, line 66: top_level missing from docstring

joachimmetz (changes requested):
@rick-slin I've made some changes PTAL so that in the future you can address these before submitting the PR.

There is one question regarding the timestamp that needs to be answered before this PR can be merged.

plaso/parsers/plist_plugins/ios_identityservices.py
rick-slin: I did not think of that case. My initial though was to avoid  returning an empty dfdatetime_cocoa_time.CocoaTime object.  

I could remove _GetDateTimeValueFromPlistKey entirely and simplify the code

plaso/parsers/plist_plugins/ios_identityservices.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    for _, process_name, process_values in self._RecurseKey(top_level, depth=1):
+      if process_name == 'CacheVersion':
+        continue
+
+      for apple_identifier, apple_identifier_values in process_values.items():
+        event_data = IOSIdstatusacheEventData()
+        event_data.process_name = process_name
+        event_data.apple_identifier = apple_identifier
+        timestamp = apple_identifier_values.get('LookupDate')
```
joachimmetz, line 62: What if this returns None?

plaso/parsers/plist_plugins/ios_identityservices.py
joachimmetz: There are multiple case here timestamp is None or timestamp is 0

the first indicates there is no timestamp value, the second could potentially indicate the timestamp is not set.

plaso/parsers/plist_plugins/ios_identityservices.py
rick-slin: My apologies, I understood from your previous comment that a timestamp of None was acceptable and that the event would just be dropped by the parser_mediator.

If that is not the case, I should proactively pass the lookup events without a timestamp.

joachimmetz: so `event_data.lookup_time` can be None but timestamp cannot, otherwise it'll set lookup_time to an empty CocoaTime object (which is not None)

joachimmetz: I've made some changes, on the assumption that 0 is a valid timestamp.

joachimmetz (approved):
LGTM

PR #4687, opened 2023-06-14
printing out hex for binary in winreg hives

chb2mn:

Changing binary output in winreg from summary of bytes to actual outputs.

## Description:

Analysts sometimes have the need to read the binary data in a registry value, so this changes binary to be output as hex.

### DIFF
--- a/plaso/parsers/winreg_plugins/interface.py
+++ b/plaso/parsers/winreg_plugins/interface.py
@@ -289,8 +289,11 @@ class WindowsRegistryPlugin(plugins.BasePlugin):
                 registry_value.DataIsString()):
             value_string = '{0!s}'.format(value_object)
 
+          elif registry_value.DataIsBinaryData():
+            value_string = '0x{}'.format(value_object.hex())
+
           else:
-            # Represent remaining types like REG_BINARY and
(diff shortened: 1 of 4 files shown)


joachimmetz:
@chb2mn unfortunately this can lead to very large blobs, and this has been a deliberate decision not to print unknown binary data. Could you give me a bit more context about common cases you would like to be able to analyze. Should these be supported in another way, e.g. winreg plugin?

chb2mn:
@joachimmetz I'll DM you use case stuff, as for a solution I was thinking some kind of config option, but a plugin might be sufficient

joachimmetz:
Per side conversation getting to this as soon as https://github.com/log2timeline/plaso/issues/4700 is completed

joachimmetz:
@chb2mn with https://github.com/log2timeline/plaso/pull/4715 merged let's discuss how to export the binary data in JSON. The built in serializer uses a type definition for this: https://github.com/log2timeline/plaso/blob/main/plaso/serializer/json_serializer.py#L76

joachimmetz:
Per side conversation, closing this in favor of alternative approach

PR #4715, opened 2023-07-23
Added option to extract Windows Registry binary values

joachimmetz:
TODO:

* [x] test multi processing
* [ ] follow up have JSON output export data as base16


joachimmetz:
@chb2mn FYI

PR #4861, opened 2024-04-04
Draft: implemented text parser plugin for Apple ps.txt files.

rick-slin:
## Description:
DRAFT: Implemented a text parser plugin for Apple ps.txt files found in sysdiagnose dumps. It uses the DatelessLogHelper 

**Related issue (if applicable):** fixes4697

## Notes:

## Checklist:
* [ ] No new [new dependencies](https://plaso.readthedocs.io/en/latest/sources/developer/Adding-a-new-dependency.html) are required or l2tdevtools has been updated
* [ ] Reviewer assigned

### DIFF
--- a/plaso/parsers/text_plugins/apple_pstxt.py
+++ b/plaso/parsers/text_plugins/apple_pstxt.py
@@ -1,0 +1,289 @@
+# -*- coding: utf-8 -*-
+"""Text parser plugin for Apple ps.txt files from sysdiagnose files."""
+
+import datetime
+import pyparsing
+
+from dfdatetime import time_elements as dfdatetime_time_elements
+
+from plaso.containers import events
+from plaso.lib import dateless_helper
+from plaso.lib import errors
+from plaso.parsers import text_parser
+from plaso.parsers.text_plugins import interface
+
+
+class ApplePSTxtEventData(events.EventData):
+  """Apple ps.txt event data.
+
+  Attributes:
+    command (str): Command and arguments that launched the process.
+    control_terminal_name (str): control terminal name (two letter
+        abbreviation).
+    cpu (str): % of cpu usage.
+    flags (str): Process flags, in hexadecimal.
+    memory (str): % of memory usage.
+    nice_value (str): nice_value.
+    persona (str): persona.
+    process_identifier (str): process identifier.
+    parent_process_identifier (str): process_parent_identifier
+    resident_set_size (str): Resident Set Size
+    scheduling_priority (str): Scheduling Priority
+    start_time (dfdatetime.DateTimeValues): Start time of the process.
+    symbolic_process_state (str): Symbolic process state.
+    up_time (str): Accumulated cpu time.
+    user (str): User name.
+    user_identifier (str): User identifier
+    virtual_size (str): Virtual size in kilobytes.
+    wait_channel (str): Wait channel.
+  """
+
+  DATA_TYPE = 'apple:pstxt:entry'
+
+  def __init__(self):
+    """Initializes event data."""
+    super(ApplePSTxtEventData, self).__init__(data_type=self.DATA_TYPE)
+    self.command = None
+    self.control_terminal_name = None
+    self.cpu = None
+    self.flags = None
+    self.memory = None
+    self.nice_value = None
+    self.persona = None
+    self.process_identifier = None
+    self.parent_process_identifier = None
+    self.resident_set_size = None
+    self.scheduling_priority = None
+    self.start_time = None
+    self.symbolic_process_state = None
+    self.up_time = None
+    self.user = None
+    self.user_identifier = None
+    self.virtual_size = None
+    self.wait_channel = None
+
+
+class ApplePSTextPlugin(
+  interface.TextPlugin, dateless_helper.DateLessLogFormatHelper):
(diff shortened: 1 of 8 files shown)


tests/parsers/text_plugins/apple_pstxt.py
```diff
@@ -0,0 +1,104 @@
...
+        'user_identifier': '501',
+        'virtual_size': '407957312',
+        'wait_channel': '-'
+    }
+
+    event_data = storage_writer.GetAttributeContainerByIndex('event_data', 52)
+    self.CheckEventData(event_data, expected_event_values)
+
+    expected_event_values = {
+        'command': '/System/Library/PrivateFrameworks/'
```
joachimmetz, line 78: style nit: use `(...)` to split the string cross lines

plaso/parsers/text_plugins/apple_pstxt.py
```diff
@@ -0,0 +1,289 @@
...
+      pyparsing.one_of('AM PM').set_results_name('ampm')
+  )
+
+  _HEADER = (
+      pyparsing.Literal('USER             UID PRSNA   PID  PPID        F  %CPU '
+                        '%MEM PRI NI      VSZ    RSS WCHAN    TT  STAT STARTED'
+                        '      TIME COMMAND') + pyparsing.LineEnd())
+
+  _COMMAND = pyparsing.OneOrMore(
+    pyparsing.Word(pyparsing.printables),
```
joachimmetz, line 98: style nit: 4 space continuation indentation

plaso/parsers/text_plugins/apple_pstxt.py
```diff
@@ -0,0 +1,289 @@
...
+      parser_mediator (ParserMediator): mediates interactions between parsers
+          and other components, such as storage and dfVFS.
+      key (str): name of the parsed structure.
+      structure (pyparsing.ParseResults): tokens from a parsed log line.
+
+    Raises:
+      ParseError: if the structure cannot be parsed.
+    """
+    # Retrieve the data from the file's metadata
+    self._SetEstimatedDate(parser_mediator)
```
joachimmetz, line 142: Move this into CheckRequiredFormat e.g. https://github.com/log2timeline/plaso/blob/main/plaso/parsers/text_plugins/syslog.py#L761

plaso/parsers/text_plugins/apple_pstxt.py
```diff
@@ -0,0 +1,289 @@
...
+        hours = int(parsed_elements.get('hours'))
+        minutes = int(parsed_elements.get('minutes'))
+        ampm = parsed_elements.get('ampm')
+        if ampm == 'PM':
+          hours += 12
+        if hours == 24:
+          hours = 0
+
+        # Retrieve year, month, day from dateless helper
+        year = self._date[0]
```
joachimmetz, line 209: Based on the format I assume date here should be the most recent date. However the date-less log helper will do the opposite it will return the earliest date as its base.

plaso/parsers/text_plugins/apple_pstxt.py
```diff
@@ -0,0 +1,289 @@
...
+
+      else:
+        raise errors.ParseError('start time was not in an expected format.')
+
+      return dfdatetime_time_elements.TimeElements(
+        time_elements_tuple=time_elements_tuple)
+
+    except (TypeError, ValueError) as exception:
+      raise errors.ParseError(
+        f'Unable to parse time elements with error: {exception}')
```
joachimmetz, line 260: style nit: to keep consistent with the rest of the code base please use `{exception!s}`

joachimmetz:
@rick-slin I will give this format some more thought, it does not fully match the use-case of the date-less log helper as with syslog where the log entries are (mostly) chronological. 

Some hints (not the same) to the format based on Linux `man ps`

```
       bsdstart    START     time the command started.  If the process was started less than 24 hours ago, the output format is " HH:MM", else it is " Mmm:SS" (where Mmm is the three letters of the month).  See also lstart, start, start_time, and stime.

       bsdtime     TIME      accumulated cpu time, user + system.  The display format is usually "MMM:SS", but can be shifted to the right if the process used more than 999 minutes of cpu time.
```

Would be good to have an example of a process that has more than 999 minutes of cpu time

rick-slin:
I don't understand the significance of the entries not being chronological. I can see the usefulness of moving the handling of the three cases from the plugin to the helper.

I can try to setup an experiment for a long lived process but I don't see how that field would impact the start time column as they appear to be independent.

joachimmetz:
> I can try to setup an experiment for a long lived process but I don't see how that field would impact the start time column as they appear to be independent.

this would be more to see if there is an edge case for the format of the TIME value

> I don't understand the significance of the entries not being chronological.

This is related to the inner workings of the date-less log helper

plaso/parsers/text_plugins/apple_pstxt.py
rick-slin: I disagree on that point. Since the file is created once by the ps command and should not be modified, the earliest timestamp should be the correct one. If the file is moved or accessed, then the most recent date would not be correct. For instance, the test file on my system as an earliest date of April 13, 2022 (modified) whereas the latest date is April 12, 2024 (accessed today).

Issue #4891, opened 2024-06-27
Empty username causes TypeError: unsupported format string passed to NoneType.__format__ in debug log statement


Melantrix:
**Describe the problem:**

There is an logger.debug command within the function AddUserAccount() in the file https://github.com/log2timeline/plaso/blob/main/plaso/preprocessors/mediator.py. This logger.debug command tries to print user_account.username. However, when the username is empty (e.g. when it's a Entra ID user and only a SID is known on the local system) the preprocessing fails, because the logger receives a TypeError because it's None. 

There are a few checks in the class WindowsUserAccountsPlugin() in the file https://github.com/log2timeline/plaso/blob/main/plaso/preprocessors/windows.py which calls AddUserAccount(), but there aren't any in AddUserAccount(). 

It's fixable by commenting out the logger.debug command, however i think it would be better to add some checks/safeguards in the function.

**To Reproduce:**

The version of Plaso you used: **20240308**

The operating system you are running Plaso on (Not the operating system of the image/files you're trying to analyze):
**In docker on Ubuntu 22.04.4 LTS**

Steps to reproduce the behavior including command line and arguments and output:

`log2timeline.py --workers 30 --storage_file timeline.plaso diskimage.img`

Please provide the source data you used when you experienced the problem. For publicly available data please provide a URL or path of the source data.
_this is not possible because of an active investigation_

The method you used to install Plaso:
**I pulled the docker image and ran the command within the docker with**
`docker run -t -i --entrypoint=/bin/bash -v /home/investigation:/data log2timeline/plaso`

**Expected behavior:**
I would expect log2timeline to continue instead of hard failing.

**Debug output/tracebacks:**

You can run log2timeline tools with "-d" to generate debug output, and include anything relevant. Also see: [Producing debug logs][https://plaso.readthedocs.io/en/latest/sources/Troubleshooting.html#producing-debug-logs]

```
Processing started.
Traceback (most recent call last):
  File "/usr/bin/log2timeline.py", line 33, in <module>
    sys.exit(load_entry_point('plaso==20240308', 'console_scripts', 'log2timeline')())
  File "/usr/lib/python3/dist-packages/plaso/scripts/log2timeline.py", line 81, in Main
    tool.ExtractEventsFromSources()
  File "/usr/lib/python3/dist-packages/plaso/cli/extraction_tool.py", line 754, in ExtractEventsFromSources
    processing_status = self._ProcessSource(session, storage_writer)
[...]

PR #4899, opened 2024-08-28
Changes to address TypeError exception in edge case

certxlm:
Use the usual "format" string method in mediator.py to avoid early f-string interpolation that causes TypeError issues and make log2timeline crashes

## Description:

When username is None, using 'f-strings' raise a "TypeErrpr: unsupported format string passed to NoneType.__format__"
Long before the KeyError that is supposed to catch the issue can be raised.
Therefore, in _ParseKey, in preprocessors/windows.py, we failed to catch the error and log2timeline stops with a failure.

TypeError

**Related issue (if applicable):** fixes #<plaso issue number here>

## Notes:

## Checklist:
* [ ] No new [new dependencies](https://plaso.readthedocs.io/en/latest/sources/developer/Adding-a-new-dependency.html) are required or l2tdevtools has been updated
* [ ] Reviewer assigned

### DIFF
--- a/plaso/preprocessors/mediator.py
+++ b/plaso/preprocessors/mediator.py
@@ -113,11 +113,12 @@ class PreprocessMediator(object):
       KeyError: if the user account already exists.
     """
     if not user_account.username:
-      logger.debug(f'adding user account: {user_account.identifier:s}')
+      logger.debug('adding user account: {0!s}'.format(user_account.username))
     else:
-      logger.debug(
-          f'adding user account: {user_account.username:s} '
-          f'({user_account.identifier:s})')
+      logger.debug('adding user account: {0!s} ({1!s})'.format(
+        user_account.username,
+        user_account.identifier
+      ))
 
     if self._storage_writer:
       self._storage_writer.AddAttributeContainer(user_account)


joachimmetz:
Please follow the style guide

joachimmetz:
Can you also indicate what the data contains when the exception raises for you

plaso/preprocessors/mediator.py
```diff
@@ -113,11 +113,12 @@ def AddUserAccount(self, user_account):
       KeyError: if the user account already exists.
     """
     if not user_account.username:
-      logger.debug(f'adding user account: {user_account.identifier:s}')
+      logger.debug('adding user account: {0!s}'.format(user_account.username))
```
joachimmetz, line 116: Please use the corresponding f string notation instead

joachimmetz: username is not set why debug print it?

plaso/preprocessors/mediator.py
```diff
@@ -113,11 +113,12 @@ def AddUserAccount(self, user_account):
...
     else:
-      logger.debug(
-          f'adding user account: {user_account.username:s} '
-          f'({user_account.identifier:s})')
+      logger.debug('adding user account: {0!s} ({1!s})'.format(
```
joachimmetz, line 118: why is identifier not set or not a string in your case?

plaso/preprocessors/mediator.py
certxlm: I don't thing the issue is about printing it or not. I thing the f-string is the issue here, it **is** extrapolated whether you pass into logger or not. Hence the bug.

plaso/preprocessors/mediator.py
certxlm: In our case the username value was "None" when in WindowsUserAccountsPlugin / _ParseKey
Looking more closely at the code, I think it's probably because in this section, if you fall in the first condition:
```
if not registry_value:
  username = 'N/A'
else:
  profile_path = registry_value.GetDataAsObject()
  username = self._GetUsernameFromProfilePath(profile_path)

  user_account.user_directory = profile_path or None
  user_account.username = username or None
```

Then neither `user_account. user_directory ` nor `user_account.username`  is set.
Which then fail in  `mediator.AddUserAccount(user_account)`, but not with a KeyError

plaso/preprocessors/mediator.py
certxlm: so basically, if you want to keep the f-string, enclosing the "debug" statement with an "if" close testing if the values are **not** none would work.

plaso/preprocessors/mediator.py
joachimmetz: can you provide the edge case data why identifier is not set?

plaso/preprocessors/mediator.py
joachimmetz: can you provide the edge case data why identifier is not set? given it is derived from the Registry value name

plaso/preprocessors/mediator.py
joachimmetz: I'm asking about identifier not username that is the one where you are using `!s`

joachimmetz:
It also looks like you are making changes to code that has been previously changed to address the issue https://github.com/log2timeline/plaso/issues/4891

joachimmetz:
Closing this for now unless you can provide me with the test data (so a CI test can be built) and indication that the changed code is failing. 

This it is why it is important to report which version of plaso you used.

plaso/preprocessors/mediator.py
certxlm: This is a raw image from an ongoing investigation, I'm afraid I cannot give it. I can run whatever debug command you need though if it can help

joachimmetz: try running the pre-release and see if the issue still surfaces.

Also no need to provide the "exact" data but something representative enough of the edge case, e.g. change the username to "John Doe", if the identifier is an UUID generate a different one. As long as the example is representative of the edge case.

certxlm:
Well, we're using the version from your repository, so _plaso - log2timeline version 20240308_ on Ubuntu 22.04 LTS

For information, I've added debugging information to the preprocessors/windows.py class, you'll find the ouptut below. From a fresh install in a new pip env (minus the downgrade of the acstore that was needed as per one of the issue currently opened), which caused the same issue.

I'll try to run log2timeline from the pre-release version tomorrow, and we'll let you know if the bug is fixed or not.
That being said, I stand by my first remark. Using f-string in conjonction with logger will always lead to unexpected behavior, since the "string" is interpolated before being passed to logger. Which is not the case with format&co.

Lastly, the devices being investigated are Windows terminal server, not sure if it might impact log2timelime or not.

```
 2024-08-28 18:50:06,161 [ERROR] (MainProcess) PID:1393407 <windows> looking for user_account with identifier=S-1-5-21-2025429265-1454471165-682003330-2243
 2024-08-28 18:50:06,161 [ERROR] (MainProcess) PID:1393407 <windows> user_account: <plaso.containers.artifacts.UserAccountArtifact object at 0x7fe65fb1bbe0>
 2024-08-28 18:50:06,161 [ERROR] (MainProcess) PID:1393407 <windows> user_account.full_name = None
 2024-08-28 18:50:06,162 [ERROR] (MainProcess) PID:1393407 <windows> user_account.group_identifier = None
 2024-08-28 18:50:06,162 [ERROR] (MainProcess) PID:1393407 <windows> user_account.identifier = S-1-5-21-2025429265-1454471165-682003330-2243
 2024-08-28 18:50:06,162 [ERROR] (MainProcess) PID:1393407 <windows> user_account.user_directory = None
 2024-08-28 18:50:06,162 [ERROR] (MainProcess) PID:1393407 <windows> user_account.username = None
 2024-08-28 18:50:06,162 [ERROR] (MainProcess) PID:1393407 <windows> registry_value: None
```

certxlm:
Actually was not that long to try :-)
Running on the main branch, log2timeline does not fail anymore. We'll keep two separate installations for "problematic" images until a new package/docker is available.

Thanks for the hard work!

joachimmetz:
> That being said, I stand by my first remark. Using f-string in conjonction with logger will always lead to unexpected behavior, since the "string" is interpolated before being passed to logger. 

Yeah but f-string interpolation is not the issue here, nor is the proposed PR addressing the issue you are encountering. It is important to spend some time in understanding the problem before creating a fix.

thanks for the example data, I'll see if that can be used for a CI test case and for confirming the original changes address the issue.

PR #4904, opened 2024-10-04
IIS parser: Support few more username variations

pyllyukko:
E.g.:

* realm/username
* username@realm


joachimmetz:
@pyllyukko thanks for the changes could you add corresponding test cases for these as well?

pyllyukko:
> @pyllyukko thanks for the changes could you add corresponding test cases for these as well?

Will do.

joachimmetz (approved):
LGTM, thanks for the changes

PR #4910, opened 2024-10-10
IIS parser: Support User-Agents with square brackets

pyllyukko:
Example taken from
https://stackoverflow.com/questions/29094232/what-is-the-user-agent-string-for-facebook-app-from-android


joachimmetz (approved):
LGTM, thanks for the changes

PR #4913, opened 2024-10-12
Added Python 3.13 CI test target

joachimmetz:


joachimmetz:
Looks like argparse behavior changed in Python 3.13 https://github.com/log2timeline/plaso/actions/runs/11305340722/job/31444728087?pr=4913#step:7:4055

PR #4949, opened 2024-12-30
Feature: Map Extracted Files to Artifact Definitions in image_export.py

sa3eed3ed:
## Description:
This PR adds an optional feature to Plaso's image_export.py tool to generate a JSON file mapping extracted files to the artifact definitions that led to their extraction. This mapping provides valuable context about the extracted files.

#### Functionality:

The new `--enable_artifacts_map` flag activates this feature. When enabled, the tool creates an `artifacts_map.json` file in the output directory. This file contains a dictionary where:

Keys: Artifact definition names (e.g., `JupyterConfigFile`, `SshdConfigFile`, `WindowsEnvironmentVariableComSpec`).
Values: Lists of extracted file paths (relative to the output directory) that matched the corresponding artifact definition.

```
plaso/scripts/image_export.py --artifact_filters JupyterConfigFile,SshdConfigFile \
--write /home/user/tmp --enable_artifacts_map --logfile /home/user/tmp/log.log \
--volumes all --partitions all /home/user/artifact_disk.dd
```
This command would produce an `artifacts_map.json` file similar to:
```
{
  "SshdConfigFile": ["etc/ssh/sshd_config"],
  "JupyterConfigFile": ["home/dummyuser/.jupyter/jupyter_notebook_config.py"]
}
```
This output indicates that the files `etc/ssh/sshd_config` and `home/dummyuser/.jupyter/jupyter_notebook_config.py` were extracted because they matched the `SshdConfigFile` and `JupyterConfigFile` artifact definitions, respectively.

##### Registry Artifacts:

For artifacts that rely on Windows Registry keys or values (e.g., `WindowsEnvironmentVariableComSpec`), the tool automatically extracts the relevant registry hive files (e.g., `SYSTEM`, `SOFTWARE`, `NTUSER.DAT`). The `artifacts_map.json` will map these hive files to both:

The artifact that directly triggered the hive's extraction (e.g., `WindowsSystemRegistryFiles`).
Any artifacts that rely on data within those hives (e.g., `WindowsEnvironmentVariableComSpec`).

Example with Registry Artifacts:
If you run `image_export.py` with `--artifact_filters WindowsEnvironmentVariableComSpec`, the `artifacts_map.json` might contain:
```
{
  "WindowsSystemRegistryFiles": [
    "System Volume Information/Syscache.hve",
    "Windows/System32/config/SAM",
    "Windows/System32/config/SECURITY",
    "Windows/System32/config/SOFTWARE",
    "Windows/System32/config/SYSTEM"
  ],
  "WindowsEnvironmentVariableComSpec": [
    "System Volume Information/Syscache.hve",
    "Users/Warren/AppData/Local/Microsoft/Windows/UsrClass.dat",
    "Users/Warren/NTUSER.DAT",
    "Windows/ServiceProfiles/LocalService/NTUSER.DAT",
    "Windows/ServiceProfiles/NetworkService/NTUSER.DAT",
    "Windows/System32/config/SAM",
    "Windows/System32/config/SECURITY",
    "Windows/System32/config/SOFTWARE",
    "Windows/System32/config/SYSTEM"
  ],
  "WindowsUserRegistryFiles": [
    "Users/Warren/AppData/Local/Microsoft/Windows/UsrClass.dat",
    "Users/Warren/NTUSER.DAT",
    "Windows/ServiceProfiles/LocalService/NTUSER.DAT",
    "Windows/ServiceProfiles/NetworkService/NTUSER.DAT"
  ]
}
```

This shows that the `SYSTEM`, `SOFTWARE`, and other hive files were extracted because of both `WindowsSystemRegistryFiles` and `WindowsEnvironmentVariableComSpec`, the mapped paths will be relative to the provided output path under the `--write` argument.

##### Technical Details:

The core of this feature is the ArtifactsTrie class, which stores artifact definition paths in a Trie (prefix tree) data structure.

###### Artifacts Trie Structure

- Root Node: A special node that doesn't represent a path segment but has children for each unique path separator in the definitions.
- Path Separator Nodes: Children of the root, representing path separators (e.g., /, \).
- Other Nodes: Each node represents a path segment from an artifact definition.
- Glob Handling: Glob patterns (like * and **) are stored as literal node keys.
- Artifact Names: Nodes corresponding to the end of a valid artifact path store a list of associated artifact names in their artifacts_names attribute.
Example Trie:

```
Root
├── / (path separator)
│   ├── Users
│   │   └── **
│   │       └── Downloads
│   │           └── *.pdf (artifacts_names: ["PDFDownloads"])
│   └── Windows
│       └── System32
│           └── config
│               └── SAM (artifacts_names: ["WindowsSAMRegistry"])
└── \ (path separator)
    └── Users
        └── *\
            └── AppData
                └── Local
                    └── test.ini (artifacts_names: ["LocalAppDataFiles"])
```
###### Matching Logic

Paths are normalized to use os.sep as the separator.
The `GetMatchingArtifacts` method traverses the Trie based on input path segments, using `fnmatch.fnmatch` for glob matching. `**` is handled recursively to match zero or more directory levels.

##### Source Type Handling
When the input  to the tool is:
- Directory: A `dfvfs.FileSystem` object of type `OS` is created, with a `dfvfs.FileSystemSearcher` using the input directory as the mount point. The tool extracts files matching the FindSpec's criteria within this directory.
- File: `ExtractPathSpecs` yields the input file path directly without searching, as it's assumed that a user-provided file path should be extracted.

Added safeguard check to exit and print if input is file, this tool can handle images, block devices and hierarchy of directories from the evidence system

### DIFF
--- a/plaso/engine/artifact_filters.py
+++ b/plaso/engine/artifact_filters.py
@@ -2,15 +2,16 @@
 """Helper to create filters based on forensic artifact definitions."""
 
 import os
+
 from artifacts import definitions as artifact_types
 
 from dfvfs.helpers import file_system_searcher as dfvfs_file_system_searcher
 
 from dfwinreg import registry_searcher as dfwinreg_registry_searcher
 
+from plaso.engine import artifacts_trie
 from plaso.engine import logger
 from plaso.engine import path_helper
-from plaso.engine import artifacts_trie
 
 
 class ArtifactDefinitionsFiltersHelper(object):
@@ -30,10 +31,10 @@ class ArtifactDefinitionsFiltersHelper(object):
         generated Windows Registry find specifications.
     registry_find_specs (list[dfwinreg.FindSpec]): Windows Registry find
         specifications.
-    registry_find_specs_artifact_names (list[]str): Windows Registry artifact
+    registry_find_specs_artifact_names (list[str]): Windows Registry artifact
         names corresponding to the find specifications.
     artifacts_trie (ArtifactsTrie): Trie structure for storing artifact
-        definitionpaths.
+        definition paths.
   """
 
   _COMPATIBLE_REGISTRY_KEY_PATH_PREFIXES = frozenset([
@@ -62,12 +63,9 @@ class ArtifactDefinitionsFiltersHelper(object):
     self.artifacts_trie = artifacts_trie.ArtifactsTrie()
 
   def _BuildFindSpecsFromArtifact(
-          self,
-          definition,
-          environment_variables,
-          user_accounts,
+          self, definition, environment_variables, user_accounts,
           enable_artifacts_map=False,
-          original_registery_artifact_filter_names=None):
+          original_registry_artifact_filter_names=None):
     """Builds find specifications from an artifact definition.
 
     Args:
@@ -77,8 +75,8 @@ class ArtifactDefinitionsFiltersHelper(object):
       user_accounts (list[UserAccountArtifact]): user accounts.
       enable_artifacts_map (Optional[bool]): True if the artifacts path map
           should be generated. Defaults to False.
-      original_registery_artifact_filter_names (Optional[set[str]]): Set of
-          original registery filter names, used in case registery hive files
+      original_registry_artifact_filter_names (Optional[set[str]]): Set of
+          original registry filter names, used in case Windows Registry hive files
           are being requested as a result of a previous filter.
 
     Returns:
@@ -96,8 +94,8 @@ class ArtifactDefinitionsFiltersHelper(object):
               environment_variables,
               user_accounts,
               enable_artifacts_map=enable_artifacts_map,
-              original_registery_artifact_filter_names=(
-                  original_registery_artifact_filter_names))
+              original_registry_artifact_filter_names=(
+                  original_registry_artifact_filter_names))
           find_specs.extend(specifications)
           self.file_system_artifact_names.add(definition.name)
 
@@ -136,8 +134,8 @@ class ArtifactDefinitionsFiltersHelper(object):
               environment_variables,
               user_accounts,
               enable_artifacts_map=enable_artifacts_map,
-              original_registery_artifact_filter_names=(
-                  original_registery_artifact_filter_names))
+              original_registry_artifact_filter_names=(
+                  original_registry_artifact_filter_names))
           find_specs.extend(specifications)
 
       else:
@@ -153,7 +151,7 @@ class ArtifactDefinitionsFiltersHelper(object):
           environment_variables,
           user_accounts,
           enable_artifacts_map=False,
-          original_registery_artifact_filter_names=None):
+          original_registry_artifact_filter_names=None):
     """Builds find specifications from a artifact group name.
 
     Args:
@@ -163,8 +161,8 @@ class ArtifactDefinitionsFiltersHelper(object):
       user_accounts (list[UserAccountArtifact]): user accounts.
       enable_artifacts_map (Optional[bool]): True if the artifacts path map
           should be generated. Defaults to False.
-      original_registery_artifact_filter_names (Optional[set[str]]): Set of
-          original registery filter names, used in case registery hive files
+      original_registry_artifact_filter_names (Optional[set[str]]): Set of
+          original registry filter names, used in case registry hive files
           are being requested as a result of a previous filter.
 
     Returns:
@@ -182,8 +180,8 @@ class ArtifactDefinitionsFiltersHelper(object):
         environment_variables,
         user_accounts,
         enable_artifacts_map=enable_artifacts_map,
-        original_registery_artifact_filter_names=(
-            original_registery_artifact_filter_names))
+        original_registry_artifact_filter_names=(
+            original_registry_artifact_filter_names))
 
   def _BuildFindSpecsFromRegistrySourceKey(self, key_path):
     """Build find specifications from a Windows Registry source type.
@@ -224,7 +222,7 @@ class ArtifactDefinitionsFiltersHelper(object):
           environment_variables,
           user_accounts,
           enable_artifacts_map=False,
-          original_registery_artifact_filter_names=None):
+          original_registry_artifact_filter_names=None):
     """Builds find specifications from a file source type.
 
     Args:
@@ -236,8 +234,8 @@ class ArtifactDefinitionsFiltersHelper(object):
       user_accounts (list[UserAccountArtifact]): user accounts.
       enable_artifacts_map (Optional[bool]): True if the artifacts path map
           should be generated. Defaults to False.
-      original_registery_artifact_filter_names (Optional[set[str]]): Set of
-          original registery filter names, used in case registery hive files
+      original_registry_artifact_filter_names (Optional[set[str]]): Set of
+          original registry filter names, used in case registry hive files
           are being requested as a result of a previous filter.
 
     Returns:
@@ -266,7 +264,7 @@ class ArtifactDefinitionsFiltersHelper(object):
         if enable_artifacts_map:
           self._AddToArtifactsTrie(artifact_name,
                                    expanded_path,
-                                   original_registery_artifact_filter_names,
+                                   original_registry_artifact_filter_names,
                                    path_separator)
 
     return find_specs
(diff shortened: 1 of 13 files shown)


plaso/engine/artifact_filters.py
```diff
@@ -52,16 +58,28 @@ def __init__(self, artifacts_registry):
...
     """Builds find specifications from an artifact definition.
 
     Args:
       definition (artifacts.ArtifactDefinition): artifact definition.
       environment_variables (list[EnvironmentVariableArtifact]): environment
           variables.
       user_accounts (list[UserAccountArtifact]): user accounts.
+      enable_artifacts_map (Optional[bool]): True if the artifacts path map
+          should be generated. Defaults to False.
+      original_registery_artifact_filter_names (Optional[set[str]]): Set of
```
joachimmetz: @sa3eed3ed could change "registery" to "registry" in this PR.

sa3eed3ed: Done

plaso/engine/artifacts_trie.py
```diff
@@ -0,0 +1,297 @@
...
+    artifacts_names (list[str]): Names of artifacts associated with this node.
+    path_separator (str): Path separator used in the Trie.
+    is_root (bool): True if this is the root node.
+  """
+
+  def __init__(self, path_separator=None, is_root=False):
+    """Initializes a trie node object.
+
+    Args:
+      path_separator (str): the path separator used in paths stored in
```
joachimmetz, line 26: Why not make '/' the default ?

plaso/engine/artifacts_trie.py
```diff
@@ -0,0 +1,297 @@
...
+      matching_artifacts = set()
+      if sub_root_node.artifacts_names:
+        matching_artifacts.update(sub_root_node.artifacts_names)
+      return list(matching_artifacts)
+    path_segments = path.strip(path_separator).split(path_separator)
+    matching_artifacts = set()
+    # Update self.artifacts_paths before starting the search.
+    self.artifacts_paths = self._GetArtifactsPaths(sub_root_node)
+
+    def _search_trie(node, current_path, segments):
```
joachimmetz, line 114: Is it possible to make this a classmethod ?

plaso/engine/artifacts_trie.py
```diff
@@ -0,0 +1,299 @@
...
+        return
+
+      segment = segments[0]
+      remaining_segments = segments[1:]
+
+      # Handle glob characters in the current segment.
+      for child_segment, child_node in node.children.items():
+        if (
+            child_segment == segment or
+            # comapring the sanitized version of the path segment stored in
```
joachimmetz, line 144: comapring => comparing.

could you restructure this code

(1) move nested function calls to separate lines
(2) combine the comments into 1 leading comment

plaso/engine/artifacts_trie.py
```diff
@@ -0,0 +1,299 @@
...
+      path_separator: The path separator used in the glob pattern.
+
+    Returns:
+      True if the path matches the glob pattern, False otherwise.
+    """
+    # Normalize paths using the appropriate separators
+    glob_pattern = glob_pattern.strip(path_separator).split(path_separator)
+    path = path.strip(path_separator).split(path_separator)
+
+    i = 0
```
joachimmetz, line 278: per style guide please don't use variable names that don't relay their purpose, this is done in other places as well

plaso/engine/artifacts_trie.py
sa3eed3ed: Done

plaso/engine/artifacts_trie.py
sa3eed3ed: Done

plaso/engine/artifacts_trie.py
sa3eed3ed: Done

plaso/engine/artifacts_trie.py
```diff
@@ -0,0 +1,304 @@
...
+      elif glob.has_magic(child_segment):
+        if self._MatchesGlobPattern(
+                child_segment, segment, child_node.path_separator):
+          self._SearchTrie(child_node, self._CustomPathJoin(
+              path_separator,
+              current_path, segment), remaining_segments, path_separator,
+              matching_artifacts)
+          self._SearchTrie(
+              node,
+              self._CustomPathJoin(
```
joachimmetz, line 171: is there a specific reason to execute ` self._CustomPathJoin(path_separator, current_path, segment)` twice here? or could the result be stored in a local variable ?

joachimmetz: Note that I changed the code, but I'm not sure if there was a reason or not.

plaso/engine/artifacts_trie.py
```diff
@@ -0,0 +1,304 @@
...
+
+    Args:
+      node (TrieNode): current trie node being traversed.
+
+    Returns:
+        dict: dictionary mapping artifact names to their paths.
+    """
+    artifacts_paths = {}
+
+    def _collect_paths(node, current_path, artifacts):
```
joachimmetz, line 225: could this be moved into a class method ?

plaso/engine/artifacts_trie.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    current_path = current_path.replace(separator, os.sep)
+    joined_path = os.path.join(current_path, new_segment)
+    return joined_path.replace(os.sep, separator)
+
+  def _MatchesGlobPattern(self, glob_pattern, path, path_separator):
+    """Checks if a path matches a given glob pattern.
+
+    Args:
+      glob_pattern: The glob pattern to match against.
```
joachimmetz, line 274: please add the types to the docstrings

plaso/engine/artifacts_trie.py
```diff
@@ -0,0 +1,304 @@
...
+      artifact_path (str): The artifact path to compare against.
+      artifact_path_seperator (str): Path separator for the artifact path.
+
+    Returns:
+      bool: True if the current path matches the artifact path or its
+          sanitized version, False otherwise.
+    """
+    artifact_path_segments = self._GetNonEmptyPathSegments(
+        artifact_path, artifact_path_seperator)
+    sanitized_path_segments = path_helper.PathHelper.SanitizePathSegments(
```
joachimmetz, line 197: Why sanitize the path here? doesn't that cause matches that incorrect matches given the sanitation is lossy ?

plaso/engine/path_helper.py
```diff
@@ -408,3 +415,47 @@ def GetWindowsSystemPath(cls, path, environment_variables):
     path = cls.ExpandWindowsPath(path, environment_variables)
 
     return path, filename
+
+  @classmethod
+  def SanitizePathSegments(cls, path_segments):
```
joachimmetz, line 420: I would recommend we move this to the cli submodule and only use it for CLI and log output.

sa3eed3ed: will wait for you reply to the comment above regarding comparing path of extracted artifact to the sanitized version of the path extracted from the trie as well. If we remove this, I will move this back to the cli sub module. but moving it there now will introduce cyclic import as cli module import artifact_trie which needs to access this method

plaso/engine/artifacts_trie.py
sa3eed3ed: done

plaso/engine/artifacts_trie.py
sa3eed3ed: done

plaso/engine/artifacts_trie.py
sa3eed3ed: this should work, thanks

plaso/engine/artifacts_trie.py
joachimmetz: ack, added a comment

plaso/engine/path_helper.py
joachimmetz: I see, let's leave it in for now

tests/engine/artifact_filters.py
```diff
@@ -304,7 +340,134 @@ def testBuildFindSpecsFromFileSourcePath(self):
...
+    self.assertIn('home', path_trie_node.children)
+    self.assertIn('Users', path_trie_node.children)
+    self.assertIn('homes', path_trie_node.children)
+
+    self.assertEqual(
+        path_trie_node.children['Windows'].artifacts_names, [])
+    self.assertIn(
+        'test_data', path_trie_node.children['Windows'].children)
+
+    self.assertEqual(
```
joachimmetz, line 355: Instead of repeatedly accessing members why not use a local variable for referencing the children?

joachimmetz:
Remaining question about tests

PR #4950, opened 2025-01-19
remove unused dependency on "six"

a-detiste:
Hi !

Six was a comatibility layer used to have a billingual Python 2+3 codebase

https://wiki.debian.org/Python3-six-removal

### DIFF
--- a/dependencies.ini
+++ b/dependencies.ini
@@ -428,12 +428,6 @@ minimum_version: 2.18.0
 rpm_name: python3-requests
 version_property: __version__
 
-[six]
-dpkg_name: python3-six
-minimum_version: 1.1.0
-rpm_name: python3-six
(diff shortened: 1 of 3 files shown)


joachimmetz:
@a-detiste thanks for the changes, have you double checked that all dependencies+python versions really no longer need it on Windows, Linux and MacOS?

a-detiste:
Hi, I only use linux (and grep) but have removed six from hundred of packages https://wiki.debian.org/Python3-six-removal

If there's no `import six` left in the code what could go wrong on Windows ?

joachimmetz:
> If there's no import six left in the code what could go wrong on Windows ?

depends on the installation method, if a required dependency of a dependency, a deployment could fail to run. I'll have a closer look when time permits,

PR #5036, opened 2026-03-16
remove dependency on old standalone "mock"

a-detiste:
this codebase already uses the newer "unittest.mock" from the standard library everywhere else

* [x] No new [new dependencies](https://plaso.readthedocs.io/en/latest/sources/developer/Adding-a-new-dependency.html) are required 

--> it's actually the reverse here


joachimmetz:
Thanks for the proposed changes, given plaso uses l2tdevtools to generate dependency files, I've made comparable changes using l2tdevtools in https://github.com/log2timeline/plaso/pull/5038 

Closing this PR in favor of 5038

a-detiste:
Thanks

PR #5070, opened 2026-05-02
Added Ivanti Connect/Pulse Secure VC0 logs parser

julianghill:

Add parser for Ivanti Connect Secure VC0 runtime logs.

## Description:

Adds a parser for Ivanti Connect Secure / Pulse Secure .vc0 log files.

The parser extracts fields including:

- timestamp
- hostname
- username
- IP address
- realm
- message code
- log file type
- record identifier
- original record values
- short formatter body

Some .vc0 files only contain the file header and no log records yet. The parser treats those as valid empty log files.

The tests use small fake .vc0 records created in the test code, so no real Ivanti appliance logs or case data are included. Did test it on real data, but obvisously cant include it here.

**Related issue (if applicable):** N/A

## Notes:

## Checklist:
* [x] Test data has a Plaso compatible license. If the test data was not authored by you (the contributor), make sure to mention its orginal source in ACKNOWLEDGEMENTS.
* [x] Reviewer assigned.


joachimmetz:
Rebased with HEAD

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,334 @@
...
+    """Normalizes .vc0 record text.
+
+    Args:
+      text (str): text to clean.
+
+    Returns:
+      str: normalized text.
+    """
+    text = self._RECORD_SEPARATOR_RE.sub('\n', text)
+    text = text.replace('\x07', ' ')
```
joachimmetz: why separately  0x07 and not part of _UNWANTED_CONTROL_CHARACTERS_RE ?

Consider using NON_PRINTABLE_CHARACTER_TRANSLATION_TABLE in plaso.lib instead

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,334 @@
...
+  _MAXIMUM_BODY_VALUES = 8
+
+  _LOG_FILENAME_RE = re.compile(
+      r'^log\.(?P<log_file_type>.+?)\.vc0(?:\.old)?$')
+  _MESSAGE_CODE_RE = re.compile(r'^[A-Z]{3}\d{5}$')
+
+  _PRINTABLE_CHARACTERS = frozenset(string.printable)
+  _RECORD_SEPARATOR_RE = re.compile(
+      r'[\x17\x15\x13\x12\x05\x04\x03\x02\x01\x00]')
+  _UNWANTED_CONTROL_CHARACTERS_RE = re.compile(
```
joachimmetz: note that the C0 and C1 control character ranges are 0x00-0x20 and 0x7f-0xa0

U+fffd is the Unicode replacement character, why replace this as well?

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,334 @@
...
+
+class IvantiVC0Parser(interface.FileObjectParser):
+  """Parser for Ivanti Connect Secure .vc0 log files."""
+
+  NAME = 'ivanti_vc0'
+  DATA_FORMAT = 'Ivanti Connect Secure .vc0 log file'
+
+  _CHUNK_SIZE = 1024 * 1024
+  _HEADER_SIZE = 8192
+  _HEADER_SIGNATURE = b'\x05\x00\x00\x00\x01\x00\x00\x00'
```
joachimmetz: This is not a very unique header signature and will likely cause this parser to be invoked on other files.

tests/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,136 @@
...
+
+    parser = ivanti_vc0.IvantiVC0Parser()
+    file_object = file_entry.GetFileObject()
+    parser.Parse(parser_mediator, file_object)
+
+    return storage_writer
+
+  def testParse(self):
+    """Tests the Parse function."""
+    data = b''.join([
```
joachimmetz: why not include this as a file in test_data ?

joachimmetz:
@julianghill thanks for the proposed changes, overall they look good. I left some question/comments.

plaso/parsers/ivanti_vc0.py
julianghill: Good point. The parser now uses a file entry filter for Ivanti filenames such as log.*.vc0 and log.*.vc0.old, and still validates the VC0 header before parsing. Hope this is more inline

joachimmetz:
@julianghill can you rebase with HEAD there are merge conflicts

joachimmetz:
Rebased with HEAD

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,339 @@
...
+      ip_address (str): IP address found in the record values.
+      line_identifier (str): identifier after the timestamp in the record ID.
+      log_file_type (str): log family, such as "access" or "events".
+      message_code (str): Ivanti message code.
+      realm (str): Ivanti realm.
+      record_identifier (str): original record identifier in the form
+          hex_timestamp.hex_line_identifier.
+      record_values (str): original tab-separated values after the realm field.
+      recorded_time (dfdatetime.DateTimeValues): record timestamp.
+      source_filename (str): name of the source log file.
```
joachimmetz: Note that Plaso already track this using the event_data_source attribute container

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,339 @@
...
+            )
+            return None
+
+        event_data = IvantiVC0EventData()
+        event_data.hostname = fields[1].strip() or None
+        event_data.ip_address = (
+            self._ExtractIPAddress(fields[5]) if len(fields) > 5 else None
+        )
+        event_data.line_identifier = line_identifier
+        event_data.log_file_type = self._ExtractLogFileType(source_filename)
```
joachimmetz: Given this is dependent on the filename it can be moved to the main parsing loop and only extracted once

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,339 @@
...
+        Args:
+          file_entry (dfvfs.FileEntry): file entry.
+
+        Returns:
+          bool: True if the file entry matches.
+        """
+        if not file_entry:
+            return False
+
+        return bool(self._FILENAME_RE.match(file_entry.name.lower()))
```
joachimmetz: lower() can be replaced by adding `flags=re.IGNORECASE` to re.compile()

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,339 @@
...
+    _LOG_FILENAME_RE = re.compile(r"^log\.(?P<log_file_type>.+?)\.vc0(?:\.old)?$")
+    _MESSAGE_CODE_RE = re.compile(r"^[A-Z]{3}\d{5}$")
+
+    _NON_PRINTABLE_CHARACTER_TRANSLATION_TABLE = (
+        definitions.NON_PRINTABLE_CHARACTER_TRANSLATION_TABLE.copy()
+    )
+    _NON_PRINTABLE_CHARACTER_TRANSLATION_TABLE.pop(ord("\n"), None)
+    _NON_PRINTABLE_CHARACTER_TRANSLATION_TABLE.pop(ord("\t"), None)
+
+    _RECORD_SEPARATOR_RE = re.compile(r"[\x17\x15\x13\x12\x05\x04\x03\x02\x01\x00]")
```
joachimmetz: this can be simplified to `r"[\x00-\x05\x12\x13\15\x17]"`

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,339 @@
...
+
+        Args:
+          file_object (dfvfs.FileIO): a file-like object.
+
+        Yields:
+          str: record string.
+        """
+        file_object.seek(self._HEADER_SIZE)
+
+        decoder = codecs.getincrementaldecoder("utf-8")(errors="replace")
```
joachimmetz: why the replace? is this because incremental decoding?

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,339 @@
...
+            if not data:
+                break
+
+            text = decoder.decode(data)
+            text = self._CleanText(text)
+
+            if record_buffer:
+                text = "".join([record_buffer, text])
+
+            records = text.split("\n")
```
joachimmetz: why not use self._RECORD_SEPARATOR_RE.split() instead of translating the separators to newlines first?

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,339 @@
...
+        event_data.realm = fields[4].strip() if len(fields) > 4 else None
+        event_data.record_identifier = record_identifier
+        event_data.recorded_time = dfdatetime_posix_time.PosixTime(timestamp=timestamp)
+        event_data.source_filename = source_filename
+        event_data.username = fields[6].strip() if len(fields) > 6 else None
+        event_data.username = event_data.username or None
+
+        record_values = fields[5:]
+        event_data.record_values = "\t".join(record_values) or None
+        event_data.body = self._CreateBody(
```
joachimmetz: why is the body not `fields[7]` or `fields[7:]` ?

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,339 @@
...
+
+        Returns:
+          str: log file type or None if not available.
+        """
+        if not source_filename:
+            return None
+
+        match = self._LOG_FILENAME_RE.match(source_filename)
+        if match:
+            return match.group("log_file_type")
```
joachimmetz: missing lower()? in case the filename is in upper case

joachimmetz (commented):
Heads up, I've made some changes to address some of these already

joachimmetz:
@julianghill PTAL I made some changes. Some more example files would be beneficial, the current parser does a lot of text scraping, which makes it fragile

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,360 @@
...
+        event_data = IvantiVC0EventData()
+        event_data.hostname = fields[1] or None
+        event_data.line_number = line_number
+        event_data.log_type = log_type
+        event_data.message_code = message_code
+        event_data.recorded_time = dfdatetime_posix_time.PosixTime(timestamp=timestamp)
+        event_data.record_identifier = record_identifier
+
+        if number_of_fields > 4:
+            event_data.realm = fields[4] or None
```
joachimmetz: How certain are you about these values, given other implementations seem to refer to these in a more generic way. Or is this something specific to the "events" log type?

plaso/parsers/ivanti_vc0.py
```diff
@@ -245,14 +245,7 @@ def _ParseRecord(self, parser_mediator, log_type, record_data, record_offset):
...
         """
-        try:
-            record = record_data.decode("utf-8")
-        except UnicodeDecodeError:
-            parser_mediator.ProduceExtractionWarning(
-                f"Invalid record at offset: 0x{record_offset:08x} - unable to decode "
-                f"as UTF-8"
-            )
-            return
+        record = record_data.decode("utf-8", errors="replace")
```
joachimmetz: IMHO this is not forensically sound. You are changing data without any notice, better to err and debug what is wrong

julianghill:
Adjustment made. 
I renamed field [4] from `realm` to `log_context` to avoid assuming it is always
an Ivanti realm across all VC0 log types.

For field [5], the samples checked show it is either empty or an IP address. The
parser now keeps this conservative: it only sets ip_address when the value
validates as an IP address, and leaves it unset when the field is empty.

The fields after that vary by log type and message code, so they remain in
body instead of adding more specific fields that are not proven across all log
types.

I also added small test logs. They are based on the observed layouts, including empty field [5]/[6] seen in the real logs.
policytrace is kept header-only because the available policytrace VC0 files from incidents were header-only as well so hopefully we can get more in the future :)

joachimmetz:
@julianghill thanks for the updates, left one comment already, but will take a closer look tomorrow.

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,363 @@
...
+
+    _LOG_FILENAME_RE = re.compile(r"^.*[.](?P<log_type>.+?)[.]vc0(?:[.]old)?$")
+    _MESSAGE_CODE_RE = re.compile(r"^[A-Z]{3}\d{5}$")
+
+    _NON_PRINTABLE_CHARACTER_TRANSLATION_TABLE = (
+        definitions.NON_PRINTABLE_CHARACTER_TRANSLATION_TABLE.copy()
+    )
+    _NON_PRINTABLE_CHARACTER_TRANSLATION_TABLE.pop(ord("\t"), None)
+
+    _RECORD_SEPARATORS_RE = re.compile(b"[\\x00-\\x05\\x0a\\x12\\x13\\x15\\x17]")
```
joachimmetz: the test files only use 0x05 as a record separator. Any test files from which can be observed that other separator values are also used?

tests/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,260 @@
...
+            "message_code": "SYS12345",
+            "recorded_time": "2024-02-08T10:00:47+00:00",
+            "record_identifier": "65c4a64f.00000001",
+            "username": "jdoe",
+        }
+        event_data = storage_writer.GetAttributeContainerByIndex("event_data", 0)
+        self.CheckEventData(event_data, expected_event_values)
+
+    def testParseWithEmptyVC0File(self):
+        """Tests the Parse function with an empty .vc0 file."""
```
joachimmetz: Note that log.policytrace.vc0 is also empty, which makes this test redundant

plaso/parsers/ivanti_vc0.py
```diff
@@ -0,0 +1,363 @@
...
+
+class IvantiVC0EventData(events.EventData):
+    """Ivanti Connect Secure (.vc0) log record.
+
+    Attributes:
+      body (str): short record text for formatters.
+      hostname (str): appliance hostname.
+      ip_address (str): IP address found in the record values.
+      line_number (str): line number.
+      log_context (str): context value found before the record values.
```
joachimmetz: Isn't this value related to https://help.ivanti.com/ps/help/en_US/ICS/21.9R1/ag/auth_realms.htm ?

joachimmetz: https://hub.ivanti.com/s/article/KB43960?language=en_US

joachimmetz (approved):
LGTM

PR #5097, opened 2026-05-16
Updated test storage files for data type counter

joachimmetz:


joachimmetz:
~~Note to self, test pinfo with older storage file~~

PR #5113, opened 2026-06-14
Changes to systemd journal parser to handle corrupt and binary data

### LINKED ISSUE #5109: systemd journal parser aborts the whole file on one corrupt entry and can crash on non-UTF-8 field values
## Summary

The systemd journal parser (`plaso/parsers/systemd_journal.py`) has three robustness gaps that cause it to lose data or raise uncaught exceptions on damaged, rotated or unclean journals — the journals most relevant in forensics.

### 1. One corrupt entry aborts extraction of the entire file

In `ParseFileObject`, the loop over entry offsets does `return` when `_ParseJournalEntry` raises `ParseError`, so the first unparsable entry stops extraction of everything after it. systemd's own format guidance is the opposite — a reader "should try hard to handle this gracefully, such as skipping over the corrupted data, but allowing access to as much data around it as possible" (https://systemd.io/JOURNAL_FILE_FORMAT/).

**Evidence:** the existing `test_data/systemd/journal/system@…journalTILDE` sample (an unclean-shutdown journal) contains 46 corrupt entries in its tail. Today the parser reports 1 warning and stops at the first; with `return` → `continue` it traverses all of them (46 warnings) and would recover any valid entries following corruption (in this sample the corruption is tail-only, so the recovered count is unchanged at 2211 — but on journals with mid-file corruption the difference is real data).

### 2. Non-UTF-8 field values raise an uncaught `UnicodeDecodeError`

`_ParseJournalEntry` does `event_data.decode("utf-8")` with no error handling. Journal DATA objects are `KEY=VALUE` where the value "may include binary non-text data" (https://systemd.io/JOURNAL_FILE_FORMAT/). `UnicodeDecodeError` is a `ValueError`, not an `errors.ParseError`, so the surrounding `except errors.ParseError` never catches it and it propagates out of the parser. `journalctl` never fails here (it renders such fields as byte arrays / "blob data").

### 3. XZ and LZ4 decompression are not wrapped

`_ParseDataObject` wraps ZSTD decompression in try/except → `ParseError`, but not XZ (`lzma.decompress`) or LZ4 (`lz4_block.decompress`). Corrupt compressed data raises an uncaught `lzma.LZMAError` / `lz4.block.LZ4BlockError` instead of a `ParseError`.

## Proposed fix

- `return` → `continue` in the entry loop.
- Split the DATA object on the first `=` byte (`partition(b"=")`) and decode key/value with `errors="replace"` (event-data values must be `str`).
- Wrap XZ and LZ4 decompression in try/except → `ParseError`, matching ZSTD.

## Tests

Update `testParseDirty` to expect the now-surfaced 46 warnings (1 corrupt pointer + 45 unused objects; event count and first-warning message unchanged). Add a unit test for the binary-safe decode (text and non-UTF-8 values). The clean-journal tests are unaffected. The binary path was also validated end-to-end: a journal written by `systemd-journal-remote` with a non-UTF-8 `MESSAGE` parses to a body with U+FFFD replacement characters, matching journalctl's byte-array rendering.

## Open question

The dirty-journal warning count rises 1 → 46 because all corrupt tail entries are now reported. If that is too noisy, a run of trailing `Unsupported object type: 0` entries could be collapsed into a single "reached corrupt tail" warning. Happy to adjust.

A PR implementing the fix + tests will follow.

kev365:
## Description

Makes the systemd journal parser resilient to the damaged, rotated and unclean journals that matter most in forensics. Three related robustness fixes:

- **Continue past a corrupt entry instead of returning.** The entry loop in `ParseFileObject` did `return` on the first `ParseError`, abandoning every entry after the first unparsable one. It now does `continue`, matching systemd's reader guidance to skip corruption and recover as much surrounding data as possible.
- **Decode field data at the byte level.** `_ParseJournalEntry` did `event_data.decode("utf-8")` with no error handling; a non-UTF-8 field value (the on-disk format permits binary DATA payloads) raised an uncaught `UnicodeDecodeError` — a `ValueError`, so the surrounding `except errors.ParseError` never caught it. The split-and-decode is extracted into `_SplitFieldData`, which splits on the first `=` byte and decodes with `errors="replace"` (event-data values must be `str`).
- **Wrap XZ and LZ4 decompression** in try/except → `ParseError`, as ZSTD already was, so corrupt compressed data raises a `ParseError` instead of an uncaught library exception.

**Related issue:** fixes #5109

## Testing

- `testParseDirty` updated: the parser now reports all 46 corrupt entries in the unclean tail (1 corrupt entry pointer + 45 unused objects) instead of only the first; the event count (2211) and the first-warning message are unchanged. The 46 = 1 + 45 breakdown is asserted so the count is documented rather than magic.
- Added `testSplitFieldData` covering a text value and a non-UTF-8 value (→ U+FFFD replacement).
- Validated end-to-end against `journalctl`: a journal written by `systemd-journal-remote` with a `MESSAGE` containing non-UTF-8 bytes parses to a body with U+FFFD, matching `journalctl --output=json`'s byte-array rendering.
- Existing clean-journal tests (`testParse`, `testParseLZ4`, `testParseCompactZSTD`) unchanged and passing; Black + pylint clean.

## Open question for the reviewer

On the dirty-journal sample the extraction-warning count rises 1 → 46 because every corrupt tail entry is now surfaced rather than only the first. If that's too noisy, I'm happy to collapse a run of trailing `Unsupported object type: 0` entries into a single "reached corrupt tail" warning — let me know your preference.

## Checklist
- [x] No new dependencies are required or l2tdevtools has been updated.
- [x] Test data has a Plaso compatible license. (No new test data added.)
- [x] Reviewer assigned.
- [x] Automated checks (GitHub Actions, AppVeyor) pass.

### DIFF
--- a/plaso/parsers/systemd_journal.py
+++ b/plaso/parsers/systemd_journal.py
@@ -286,30 +286,8 @@ class SystemdJournalParser(interface.FileObjectParser, dtfabric_helper.DtFabricH
 
         return entry_object_offsets
 
-    def _DecodeFieldData(self, parser_mediator, byte_data):
-        """Decodes journal field data as UTF-8.
-
-        Produces an extraction warning when the data contains invalid UTF-8,
-        which is replaced rather than raising.
-
-        Args:
-          parser_mediator (ParserMediator): parser mediator.
-          byte_data (bytes): journal field data to decode.
-
-        Returns:
-          str: data decoded as UTF-8, with invalid bytes replaced.
-        """
-        try:
-            return byte_data.decode("utf-8")
-        except UnicodeDecodeError:
-            parser_mediator.ProduceExtractionWarning(
-                "Unable to decode journal field data as UTF-8, replaced invalid "
-                "characters"
-            )
-            return byte_data.decode("utf-8", errors="replace")
-
-    def _SplitFieldData(self, parser_mediator, field_data):
-        """Splits journal field data into a key and value.
+    def _ParseKeyValuePair(self, parser_mediator, field_data):
+        """Parses a key value pair.
 
         Args:
           parser_mediator (ParserMediator): parser mediator.
@@ -320,11 +298,27 @@ class SystemdJournalParser(interface.FileObjectParser, dtfabric_helper.DtFabricH
           tuple[str, str]: key and value decoded as UTF-8, with invalid bytes
               replaced rather than raising.
         """
-        key, _, value = field_data.partition(b"=")
-        return (
-            self._DecodeFieldData(parser_mediator, key),
-            self._DecodeFieldData(parser_mediator, value),
-        )
+        key_data, _, value_data = field_data.partition(b"=")
+
+        try:
+            key = key_data.decode("utf-8")
+        except UnicodeDecodeError:
+            parser_mediator.ProduceExtractionWarning(
+                "Unable to decode journal field key as UTF-8. Unsupported code points "
+                "are escaped."
+            )
+            key = key_data.decode("utf-8", errors="backslashreplace")
+
+        try:
+            value = value_data.decode("utf-8")
+        except UnicodeDecodeError:
+            parser_mediator.ProduceExtractionWarning(
+                f"Unable to decode journal field with key: {key:s} value as UTF-8. "
+                f"Unsupported code points are escaped."
+            )
+            value = value_data.decode("utf-8", errors="backslashreplace")
+
+        return key, value
 
     def _ParseJournalEntry(self, parser_mediator, file_object, file_offset):
         """Parses a journal entry.
@@ -363,8 +357,8 @@ class SystemdJournalParser(interface.FileObjectParser, dtfabric_helper.DtFabricH
                 )
             except (ValueError, errors.ParseError) as exception:
                 raise errors.ParseError(
-                    f"Unable to parse entry item at offset: 0x{file_offset:08x} "
-                    f"with error: {exception!s}"
+                    f"Unable to parse entry item at offset: 0x{file_offset:08x} with "
+                    f"error: {exception!s}"
                 )
 
             file_offset += entry_item_data_size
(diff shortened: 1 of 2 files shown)


plaso/parsers/systemd_journal.py
```diff
@@ -272,6 +286,24 @@ def _ParseEntryObjectOffsets(self, file_object, file_offset):
...
+          field_data (bytes): journal field data, which is "key=value" where the
+              value can contain arbitrary binary (non-UTF-8) data.
+
+        Returns:
+          tuple[str, str]: key and value decoded as UTF-8, with invalid bytes
+              replaced rather than raising.
+        """
+        key, _, value = field_data.partition(b"=")
+        return (
+            key.decode("utf-8", errors="replace"),
```
joachimmetz: This will silently replace "invalid" characters let's make sure to set an extraction warning if this situation occurs

kev365: Thanks Joachim, here is a proposed adjustment for the extraction warning:

`_SplitFieldData` now decodes each field via a new `_DecodeFieldData`, which produces an extraction warning ("Unable to decode journal field data as UTF-8, replaced invalid characters") when invalid UTF-8 is replaced, rather than replacing silently (dc23560ff). The parser mediator is threaded down (`_ParseJournalEntry` → `_SplitFieldData` → `_DecodeFieldData`). `testSplitFieldData` asserts the warning fires on a non-UTF-8 value and not on a valid one.

plaso/parsers/systemd_journal.py
```diff
@@ -272,12 +286,53 @@ def _ParseEntryObjectOffsets(self, file_object, file_offset):
...
+          str: data decoded as UTF-8, with invalid bytes replaced.
+        """
+        try:
+            return byte_data.decode("utf-8")
+        except UnicodeDecodeError:
+            parser_mediator.ProduceExtractionWarning(
+                "Unable to decode journal field data as UTF-8, replaced invalid "
+                "characters"
+            )
+            return byte_data.decode("utf-8", errors="replace")
```
joachimmetz: Unfortunately decode("utf-8", errors="replace") uses the Unicode placeholder character which makes it very hard to determine what the original data was from the extracted data. 

It might be better to omit the log line than present an altered one, or better to "escape" the unsupported byte values

joachimmetz: errors="backslashreplace" might be more appropriate

joachimmetz (approved):
LGTM

PR #5115, opened 2026-06-14
Recover from corrupt utmp records instead of aborting at the first

### LINKED ISSUE #5111: utmp parser improve handling of 64-bit and corrupt records
## Summary

Two items in the libc6 utmp parser (`plaso/parsers/utmp.py`).

### 1. A single corrupt record aborts parsing of the rest (bug)

In `ParseFileObject`, the per-record loop does `break` on the first `ParseError`, so one unparsable record (e.g. an out-of-range `ut_type`) abandons every record after it. Because utmp/wtmp/btmp records are a fixed 384-byte size, a corrupt record can be skipped and the remaining records recovered.

**Proposed fix:** on a `ParseError`, if a full record could still follow, emit a warning and seek to the next 384-byte boundary instead of stopping; trailing data shorter than a record is still treated as end-of-file. This preserves the existing trailing-data behaviour and cannot loop (offset strictly increases). The first record is still validated strictly (a non-utmp first record still raises `WrongParser`).

**Open question:** is skip-and-continue acceptable, or do you prefer the current conservative `break`?

### 2. 64-bit (400-byte) layout (note only)

`utmp.yaml` encodes the 384-byte biarch layout (correct for x86_64, where glibc forces `__WORDSIZE_TIME64_COMPAT32`). A pure-64-bit variant (8-byte `ut_session` + 16-byte `ut_tv` = 400 bytes) exists on some non-x86 64-bit arches and would not parse. Documenting only; not proposing a change unless there's demand.

## Tests

Synthetic, self-authored `utmp_corrupted` (valid / two consecutive unsupported-type / valid / truncated tail → 2 events + 2 warnings, proving the trailing record is recovered after the skipped ones and the truncated tail is treated as end-of-file). Values derived from `utmpdump` ground truth.

A PR implementing item 1 will follow.

kev365:
## Description

The libc6 utmp parser's per-record loop did `break` on the first `ParseError`, abandoning every record after a single corrupt one. Because libc6 utmp/wtmp/btmp records are a fixed 384-byte size, a corrupt record can be skipped and the rest recovered: on a `ParseError`, if a full record could still follow, emit a warning and seek to the next 384-byte boundary instead of stopping; trailing data shorter than a record is still treated as end-of-file. The first record is still validated strictly (a non-utmp first record still raises `WrongParser`), and the loop cannot spin (the offset strictly increases).

**Related issue:** fixes #5111

## Testing

Added `testParseCorruptUtmpFile` with a synthetic, self-authored `test_data/utmp_corrupted`: a valid record, two consecutive unsupported-type records, a valid record, then a truncated trailing record → 2 events + 2 warnings, proving the trailing record is recovered after the skipped ones and the truncated tail is treated as end-of-file. Expected values were derived from `utmpdump`. Existing `testParseUtmpFile` / `testParseWtmpFile` are unchanged; Black + pylint clean.

## Open question for the reviewer

Is skip-and-continue acceptable, or do you prefer the current conservative `break`? The change is isolated to the loop's exception handler and can be dropped if you'd rather keep the existing behaviour.

## Checklist
- [x] No new dependencies are required or l2tdevtools has been updated.
- [x] Test data has a Plaso compatible license. (`test_data/utmp_corrupted` is synthetic and self-authored; no external source, so no ACKNOWLEDGEMENTS entry is needed.)
- [x] Reviewer assigned.
- [x] Automated checks (GitHub Actions, AppVeyor) pass.


plaso/parsers/utmp.py
```diff
@@ -61,6 +61,9 @@ class UtmpParser(interface.FileObjectParser, dtfabric_helper.DtFabricHelper):
     _INIT_PROCESS_TYPE = 5
     _DEAD_PROCESS_TYPE = 8
 
+    # A libc6 utmp entry is a fixed-size record.
```
joachimmetz: what about the 64-bit version of the record ? https://github.com/libyal/dtformats/blob/main/documentation/Utmp%20login%20records%20format.asciidoc

kev365: Thanks Joachim. I've added support for the 64-bit (400-byte) libc6 utmp layout (e.g. aarch64, where `ut_session` and the `ut_tv` timeval are widened).

The parser selects the layout by validating the first record with each candidate: the 32-bit and 64-bit layouts share the fields before the session field, so the correct one is identified by a supported type and a valid subsecond field — the microseconds field is defined to be < 1,000,000, and a 64-bit record read with the 32-bit layout has its wider seconds field spill into it and is rejected. The record size is then derived from the selected layout, so the skip-and-recover logic no longer hard-codes 384. Verified against a sample generated on real aarch64 (`test_data/utmp_64bit`: boot, login and logout records).

joachimmetz:
Thanks for the addition, my comment was mostly on the regarding the comment in the code.

I put something together to generate test data https://github.com/dfirlabs/utmp-specimens/ 

Recovering from a corrupt utmp is tricky given 384 or 400 0-byte could be a valid record. An approach to explore is to try a small number of records and abort if the data does not make sense

Note that the utmp parser should also not interfere with other parsers in the sense of it claiming a file it is not the appropriate parser for.

Also if I'm not mistaken the format is native-endian, which Plaso does not support at the moment.

joachimmetz:
Maybe scan the the first 8 - 16 records and see:

* 0 <= ut_type <= 8
* 0 <= ut_tv.tv_usec <= 999999
* ut_id and ut_line most likely to have a value if entry.ut_type != 0

joachimmetz:
utmp from big-endian system

```
00000000  00 00 00 00 00 00 00 20  00 00 00 00 00 00 00 00  |....... ........|
00000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000150  00 00 00 00 00 00 00 00  00 00 00 00 6a 48 93 69  |............jH.i|
00000160  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000190  00 08 00 00 00 00 00 20  74 74 79 32 00 00 00 00  |....... tty2....|
000001a0  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
000001b0  00 00 00 00 00 00 00 00  74 32 00 00 00 00 00 00  |........t2......|
000001c0  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
000002e0  00 00 00 00 00 00 00 00  00 00 00 00 6a 48 93 69  |............jH.i|
000002f0  00 00 00 00 00 00 00 00  01 02 03 04 00 00 00 00  |................|
00000300  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000320  00 02 00 00 00 00 00 20  73 79 73 74 65 6d 20 62  |....... system b|
00000330  6f 6f 74 00 00 00 00 00  00 00 00 00 00 00 00 00  |oot.............|
00000340  00 00 00 00 00 00 00 00  7e 00 00 00 72 65 62 6f  |........~...rebo|
00000350  6f 74 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |ot..............|
00000360  00 00 00 00 00 00 00 00  00 00 00 00 30 2e 30 2e  |............0.0.|
00000370  30 2e 30 00 00 00 00 00  00 00 00 00 00 00 00 00  |0.0.............|
00000380  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
```

Full file https://github.com/libyal/dtformats/blob/main/test_data/utmp-s390

kev365:
Thanks Joachim, for your assist here as well as the speciman set and test data. Reworked this along the lines you suggested -- it validates the first records to pick the layout and decide whether to claim the file. So it now reads 32-bit, 64-bit and big-endian (s390x) utmp, and won't claim non-utmp files. Test data is your dtformats specimens, including utmp-s390 for the big-endian case.

joachimmetz:
Much appreciated, will try to take review the changes shortly.

joachimmetz (approved):
LGTM

PR #5117, opened 2026-06-15
Added Linux IANA time zone name preprocessor

### LINKED ISSUE #5116: Linux time zone preprocessor records an abbreviation (e.g. "CET") instead of the IANA name when /etc/localtime is a copy
## Summary

When `/etc/localtime` is a **regular file** (a copy of a zoneinfo file) rather
than a symbolic link, `LinuxTimeZonePlugin` records the time zone **abbreviation**
(for example `CET`) instead of the IANA name (for example `Europe/Zurich`).

The abbreviation is derived here:

`plaso/preprocessors/linux.py` (`LinuxTimeZonePlugin._ParseFileEntry`):

```python
else:
    # Determine the timezone based on the timezone information file.
    file_object = file_entry.GetFileObject()
    time_zone = None
    try:
        time_zone_file = tz.tzfile(file_object)
        date_time = datetime.datetime(2017, 1, 1)
        time_zone = time_zone_file.tzname(date_time)   # -> "CET"
    except ValueError:
        ...
```

The behaviour is pinned by an existing test:
`tests/preprocessors/linux.py` — `testParseFileEntryWithTZif` asserts
`time_zone.zone == "CET"`.

(The symlink path a few lines above already recovers the IANA name correctly by
splitting the link target on `zoneinfo/`.)

## Why this matters (forensic impact)

1. **Ambiguity** — `CET` maps to many IANA zones (Europe/Paris, Europe/Berlin,
   Europe/Madrid, Europe/Zurich, …). The IANA name is the precise identifier and
   carries the correct DST rules.
2. **It taints the whole timeline** — the recovered zone is used to localize
   every displayed timestamp, so an imprecise zone shifts all local times.
3. **Silent fallback to UTC** — `pytz` accepts `CET`, but most abbreviations are
   not valid zone names. For those, `SetTimeZone()` raises and the plugin emits
   "Unable to set time zone", losing the time zone entirely.
4. The abbreviation is taken as of a hard-coded `datetime(2017, 1, 1)` (winter),
   so it is also DST-sensitive.

A copied `/etc/localtime` (rather than a symlink) is common on minimal and
container images, so this is not an edge case.

## Proposed fix

Debian/Ubuntu systems write the IANA name in plaintext to **`/etc/timezone`**.
The `artifacts` project already defines `LinuxTimezoneFile` for that path, so this
is fixable in-tree:

* Add a `LinuxTimeZoneDebianPlugin` (artifact `LinuxTimezoneFile`) that reads the
  IANA name from `/etc/timezone`.
* Register it **after** `LinuxTimeZonePlugin` so the IANA name, when present,
  overrides the `/etc/localtime` abbreviation (precedence is by registration
  order — `SetTimeZone` overwrites unconditionally). The abbreviation remains the
  fallback when `/etc/timezone` is absent, so existing behaviour is preserved.

**Out of scope (needs upstream artifacts first):** RHEL/CentOS store the IANA
name as `ZONE=` in `/etc/sysconfig/clock`, but the `artifacts` project has no
definition for that file. A RHEL plugin would require a new upstream artifact
definition and a dependency bump, so it is intentionally left out here.

## Open question for maintainers

Is registration-order precedence acceptable, or would you prefer a more explicit
mechanism (e.g. only set the abbreviation if no IANA name was recovered)? Happy to
adjust.

## Relationship to #5110

The time-zone abbreviation issue was originally bundled into #5110 alongside the
LSB/os-release parsing crash. That crash was fixed by PR #5114 (merged), which
closed #5110. The two changes are independent and touch different code paths, so
this time-zone fix is tracked separately here.

## Testing

A draft change is ready with unit tests covering: IANA recovery from
`/etc/timezone`, an unsupported value (produces a warning), and precedence over a
copied `/etc/localtime` (`CET` → `Europe/Zurich`). No new binary test data is
introduced.

kev365:

Recover the IANA time zone name from `/etc/timezone` on Debian-based systems.

## Description:

When `/etc/localtime` is a copy of a zoneinfo file rather than a symbolic link,
`LinuxTimeZonePlugin` can only derive an ambiguous time zone **abbreviation**
(for example `CET`) instead of the IANA name (for example `Europe/Zurich`). The
abbreviation is imprecise, is taken as of a hard-coded winter date, and for most
abbreviations is not a usable zone name (falling back to UTC).

This change adds `LinuxTimeZoneDebianPlugin`, which reads the IANA name from the
Debian `/etc/timezone` file (via the existing `LinuxTimezoneFile` artifact). It is
registered **after** `LinuxTimeZonePlugin` so the IANA name takes precedence when
present, while the `/etc/localtime` abbreviation remains the fallback. The
`LinuxTimezoneFile` artifact definition is also vendored into the test artifacts
fixture so the plugin resolves during testing.

Unit tests cover IANA recovery from `/etc/timezone`, an unsupported value
(produces a preprocessing warning), and precedence over a copied `/etc/localtime`
(`CET` → `Europe/Zurich`). No new binary test data is introduced.

RHEL/CentOS (`/etc/sysconfig/clock`) is intentionally out of scope: the
`artifacts` project has no definition for that file, so it would require an
upstream artifact definition and a dependency bump first.

**Related issue (if applicable):** fixes #5116

## Notes:

## Checklist:
* [x] No new dependencies are required (uses the existing `LinuxTimezoneFile` artifact).
* [x] Test data has a Plaso compatible license. No new binary test data; only inline bytes, the existing `localtime.tzif`, and a self-authored artifact definition.
* [x] Reviewer assigned.
* [ ] Automated checks (GitHub Actions, AppVeyor) pass.

### DIFF
--- a/plaso/preprocessors/linux.py
+++ b/plaso/preprocessors/linux.py
@@ -217,6 +217,46 @@ class LinuxTimeZonePlugin(interface.FileEntryArtifactPreprocessorPlugin):
                 )
 
 
+class LinuxTimeZoneDebianPlugin(interface.FileArtifactPreprocessorPlugin):
+    """Linux time zone plugin for the /etc/timezone file.
+
+    The /etc/timezone file, used by Debian-based distributions, contains the
+    IANA time zone name, for example "Europe/Zurich". This is preferred over
+    the time zone abbreviation that LinuxTimeZonePlugin derives from a copied
+    /etc/localtime file, for example "CET", which is ambiguous. This plugin is
+    registered after LinuxTimeZonePlugin so that the IANA name, when present,
+    takes precedence.
+    """
+
+    ARTIFACT_DEFINITION_NAME = "LinuxTimezoneFile"
+
+    def _ParseFileData(self, mediator, file_object):
+        """Parses file content (data) for a time zone preprocessing attribute.
+
+        Args:
+          mediator (PreprocessMediator): mediates interactions between preprocess
+              plugins and other components, such as storage.
+          file_object (dfvfs.FileIO): file-like object that contains the artifact
+              value data.
+
+        Raises:
+          errors.PreProcessFail: if the preprocessing fails.
+        """
+        text_file_object = dfvfs_text_file.TextFile(file_object, encoding="utf-8")
+
+        time_zone = text_file_object.readline()
+        time_zone = time_zone.strip()
+
+        if time_zone:
+            try:
+                mediator.SetTimeZone(time_zone)
+            except ValueError:
+                mediator.ProducePreprocessingWarning(
+                    self.ARTIFACT_DEFINITION_NAME,
+                    f"Unable to set time zone: {time_zone:s} in knowledge base.",
+                )
+
+
 class LinuxUserAccountsPlugin(interface.FileArtifactPreprocessorPlugin):
     """The Linux user accounts plugin."""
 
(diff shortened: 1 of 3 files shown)


joachimmetz (approved):
LGTM, thanks for the proposed changes

Issue #5118, opened 2026-06-15
Expand systemd: journal event data with fields: identity, unit, severity/facility, source timestamp


kev365:
*This is a design proposal seeking maintainer buy-in before a PR, per the contribution process. No code is attached yet.*

## Motivation

The systemd journal is the richest structured log source on modern Linux, but
the `systemd_journal` parser currently surfaces only **5** attributes
(`body`, `hostname`, `pid`, `reporter`, `written_time`) and discards the rest.
In particular it drops the **trusted** (`_`-prefixed) fields that journald
appends itself and that the logging process cannot forge — which is exactly what
makes the journal more valuable than text syslog, where every field is
attacker-controllable.

Surfacing these brings `systemd:journal` to parity with the `syslog:line` parser
(severity/facility/pid) and beyond it (verified identity, unit, executable,
command line, audit login uid).

**Evidence (plaso's own samples):** running `journalctl --output=json` over the
four test journals (`system.journal`, `system.journal.lz4`, `user-1000.journal`,
and the dirty `…journalTILDE`) — 4,399 entries total — yields **91 distinct
fields**. The proposed fields below are the high-value subset, with their
prevalence (number of entries containing them) shown.

## Current mapping (for reference)

| attribute | journal field |
|---|---|
| `body` | `MESSAGE` |
| `hostname` | `_HOSTNAME` |
| `reporter` | `SYSLOG_IDENTIFIER` |
| `pid` | `_PID` / `SYSLOG_PID` |
| `written_time` | `__REALTIME_TIMESTAMP` |

(Minor: the `SystemdJournalEventData` docstring says `pid (int)`, but it is stored
as `str`; would correct while here.)

## Proposed Tier 1 — core forensic set

All present in the existing test data **except** `_SELINUX_CONTEXT` (see Testing).

| proposed attribute | journal field | prevalence | why it matters |
|---|---|---|---|
| `user_identifier` | `_UID` | 3269 | **trusted** uid of the logging process |
| `group_identifier` | `_GID` | 3268 | **trusted** gid |
| `process_name` | `_COMM` | 3170 | **trusted** process name |
| `executable` | `_EXE` | 2943 | **trusted** on-disk binary path |
| `command_line` | `_CMDLINE` | 2943 | full invocation |
| `systemd_unit` | `_SYSTEMD_UNIT` | 2993 | originating service |
| `boot_identifier` | `_BOOT_ID` | 4399 | correlate entries within/across boots |
[...]

PR #5139, opened 2026-07-05
Add Tier 1 fields to systemd journal event data

kev365:
## Summary

Implements the **Tier 1** field set from the design proposal in #5118, which you
approved ("Please feel free to go ahead ... Let's start with tier 1 and assess
what the volume of additional events is"). Built on the now-merged #5113
binary-safe field decode.

The `systemd_journal` parser previously surfaced only 5 attributes and discarded
the rest. This adds 14 attributes, most of them trusted (`_`-prefixed) fields
that journald adds itself and that the logging process cannot forge.

## New attributes

| attribute | journal field |
|---|---|
| `user_identifier` | `_UID` |
| `group_identifier` | `_GID` |
| `process_name` | `_COMM` |
| `executable` | `_EXE` |
| `command_line` | `_CMDLINE` |
| `systemd_unit` | `_SYSTEMD_UNIT` |
| `boot_identifier` | `_BOOT_ID` |
| `machine_identifier` | `_MACHINE_ID` |
| `transport` | `_TRANSPORT` |
| `audit_login_identifier` | `_AUDIT_LOGINUID` |
| `selinux_context` | `_SELINUX_CONTEXT` |
| `severity` | `PRIORITY` (→ syslog label) |
| `facility` | `SYSLOG_FACILITY` (→ syslog label) |
| `recorded_time` | `_SOURCE_REALTIME_TIMESTAMP` |

Also: `reporter` falls back to the trusted `_COMM` when `SYSLOG_IDENTIFIER` is
absent, and the `pid` docstring is corrected (`int` → `str`).

## Volume of additional events

The only new attribute that generates timeline events is `recorded_time`. Across
the four test journals (4,398 entries) it adds **3,177** events on top of the
4,398 `written_time` events → **7,575 total, +72.2%** — only for entries that
carry `_SOURCE_REALTIME_TIMESTAMP` (measured with log2timeline + pinfo).

## Notes / decisions

- **severity / facility as labels**, matching `syslog:line`, for cross-parser
  parity. Per `systemd.journal-fields(7)` `PRIORITY` is 0–7 and `SYSLOG_FACILITY`
  0–23, indexed directly (no RFC 3164 `>>3` / `&7`). These are *unvalidated* user
  fields — the samples contain e.g. `SYSLOG_FACILITY=DHCP4` (NetworkManager) — so
  values are bounds-checked and non-numeric ones yield `None` silently rather
  than flooding extraction warnings.
- The severity/facility label tables mirror the ones in
  `text_plugins/syslog.py`; glad to switch to importing them instead of the local
  copies if you prefer reuse.
- **recorded_time** is a distinct second timestamp (source/event time) alongside
  `written_time` (reception/collection time); for `systemd-journal-remote`
  relayed journals the two diverge.
- **selinux_context** is mapped but absent from the current test samples
  (generated on an AppArmor host), so it ships without a value assertion — happy
  to add one if a SELinux sample is contributed.

## Backward compatibility

New attributes default to `None`; acstore does not serialize `None`, so entries
that do not populate a field are byte-for-byte unchanged (no event-identity /
dedup-hash change, no storage bloat). No attribute is renamed or removed.

## Testing

Expected values are derived from `journalctl --output=json` ground truth over the
existing test journals — no new test data. Verified locally:
`tests.parsers.systemd_journal`, `tests.engine.timeliner` +
`tests.engine.yaml_timeliner_file`, pylint / black / docformatter / yamllint,
`tests.cli.psteal_tool`, and the end-to-end suite (clean — no over-claim, no
psort crash).

### DIFF
--- a/plaso/data/timeliner.yaml
+++ b/plaso/data/timeliner.yaml
@@ -1440,6 +1440,8 @@ data_type: 'systemd:journal'
 attribute_mappings:
 - name: 'written_time'
   description: 'Content Modification Time'
+- name: 'recorded_time'
+  description: 'Recorded Time'
 place_holder_event: true
 ---
 # TODO: prefix data_type with 'windows:registry:'.
--- a/plaso/parsers/systemd_journal.py
+++ b/plaso/parsers/systemd_journal.py
@@ -15,15 +15,81 @@ from plaso.lib import specification
 from plaso.parsers import interface
 from plaso.parsers import manager
 
+# Syslog severity and facility label tables, mirroring the ones in
+# plaso/parsers/text_plugins/syslog.py. In the journal the PRIORITY and
+# SYSLOG_FACILITY fields are stored as plain decimal strings (0-7 and 0-23
+# respectively per systemd.journal-fields(7)) that index directly into these
+# tables; the RFC 3164 priority-byte decomposition used for text syslog does not
+# apply here. These are unvalidated "user" fields, so their values are
+# bounds-checked before use.
+_SYSLOG_SEVERITY = [
+    "EMERG",
+    "ALERT",
+    "CRIT",
+    "ERR",
+    "WARNING",
+    "NOTICE",
+    "INFO",
+    "DEBUG",
+]
+
+_SYSLOG_FACILITY = [
+    "kernel message",
+    "user-level message",
+    "mail system",
+    "system daemons",
+    "security/authorization messages",
+    "messages generated internally by syslogd",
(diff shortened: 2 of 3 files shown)


joachimmetz (approved):
LGTM

joachimmetz:
Thanks for the proposed changes. Any indication about the average increase of number of events? Never mind I see it in https://github.com/log2timeline/plaso/issues/5118

Issue #5156, opened 2026-07-08
expand auditd (audit.log): serial, tiered fields, hex-decode, ENRICHED (extend the selinux plugin)


kev365:
*This is a design proposal seeking maintainer buy-in before a PR, per the contribution process. No code is attached yet.*

## Motivation

The Linux audit daemon (`auditd`) log is the authoritative kernel-level record of
process execution, authentication, privilege use, and file access on modern Linux —
often one of the most valuable Linux security-timeline sources. Plaso already parses
`audit.log` today, and this proposal builds on that existing plugin to surface more of
the structure the format carries.

`plaso/parsers/text_plugins/selinux.py` (registered as "SELinux audit log") parses
audit lines generically and extracts four attributes. Two areas look worth extending:

1. **The audit serial.** The grammar parses the serial integer after the timestamp but
   does not currently give it a `set_results_name`, so it isn't stored on the event
   data. Since the records of a single event (`SYSCALL`+`EXECVE`+`CWD`+`PATH`+
   `PROCTITLE`) share one `(timestamp:serial)`, surfacing the serial would let those
   records be correlated.
2. **The `key=value` payload.** The fields (`uid`/`auid`/`exe`/`comm`/`syscall`/
   `success`/`exit`/`a0..aN`/`name`/`cwd`/`ses`/`subj`, …) are currently retained
   together in the raw `message_body`. Promoting them to typed attributes — with hex
   decoding and ENRICHED handling (below) — would make them queryable while keeping the
   raw body.

## Current behavior (verified, for reference)

`selinux:line` produces:

| attribute | source |
|---|---|
| `audit_type` | `type=` |
| `last_written_time` | seconds/milliseconds in `audit(EPOCH.MSEC:…)` |
| `message_body` | the text after `):` via `restOfLine()` (raw, including any ENRICHED `0x1d` suffix) |
| `pid` | best-effort `pid=` scraped from `message_body` (stored as `str`; docstring says `int`) |

Where today's grammar (`selinux.py`) leaves room to extend:
- the serial `_INTEGER` (line ~80) has no results name, so it isn't stored;
- `message_body = restOfLine()` keeps the ENRICHED `\x1d…` suffix and all `key=value`
  structure together as one raw string;
- the best-effort `pid` extraction uses a `Word(printables)` key/value dict that
  doesn't yet handle hex values or the nested single-quoted `msg='…'` of `USER_*`
  records.

## What real captures show (evidence)

[...]

PR #5172, opened 2026-07-16
Add Tier 1 structured fields to the SELinux audit log parser

kev365:
Tier 1 of the auditd (`audit.log`) parsing enhancement from #5156 (approved:
"start with tier 1 and assess the overall impact").

`text_plugins/selinux.py` parses all of `audit.log` but currently surfaces only
`audit_type`, `last_written_time`, a raw `message_body`, and a best-effort `pid`.
This adds:

- **`audit_serial`** — the audit serial number, previously parsed then discarded;
  enables correlating a single event's records by `(timestamp:serial)`.
- **12 typed identity/outcome fields** promoted once from the parsed body:
  `audit_login_identifier` (auid), `audit_session_identifier` (ses),
  `user_identifier` (uid), `group_identifier` (gid),
  `parent_process_identifier` (ppid), `process_name` (comm), `executable` (exe),
  `security_context` (subj), `system_call` (syscall), `success`, `exit_code`, and
  a robust `pid`. Names follow the `systemd:journal` set; stored as `str`
  passthroughs. (`pid` docstring corrected `int`→`str`.)
- **ENRICHED handling** — ENRICHED is the modern cross-distro default (Fedora and
  Ubuntu 26.04); each record appends a `0x1d`-delimited interpreted suffix. The
  parser splits on `0x1d`, parses the raw body for the fields (keeping the raw
  numeric ids for offline soundness), and surfaces the resolved `SYSCALL` name
  (e.g. `execve`), which is hard to reproduce from an offline image. RAW logs are
  unaffected.

**Impact:** no new events — each `audit.log` line stays one `selinux:line` event,
so event counts and identity are unchanged and `None` attributes aren't serialized.
`audit_serial` populates on 100% of records. `tag_linux.txt`, the formatter, and
`timeliner.yaml` are intentionally unchanged (tag rules key on
`audit_type`/`message_body`, both retained; no new `*_time`).

**Testing:** unit tests over RAW (`selinux.log`) + a new ENRICHED specimen
(`audit_enriched.log`, self-generated → Apache-2.0-compatible). Verified on Ubuntu
26.04 with GIFT-PPA deps (matching CI): unit tests, `psteal` 10/10, and the full
end-to-end suite incl. the `ext4_with_binaries` over-claim guard — all green.


joachimmetz (approved):
LGTM

joachimmetz:
Thanks for the proposed changes.

PR #5176, opened 2026-07-19
Fix false EOF in EncodedTextReader.ReadLine() caused by lines_size inflation

### LINKED ISSUE #5175: lines_size inflated in EncodedTextReader.ReadLine() causes false EOF / silent event loss - 3 million log entries on test dataset dropped
**Describe the problem:**
EncodedTextReader.ReadLine() in plaso/parsers/text_parser.py corrupts its own buffer-size accounting, which causes the text-based log parser to stop at a false end-of-file on large logs that contain unparsable lines.

The reader tracks the size of the in-memory line buffer in self.lines_size, and ReadLines() only refills the buffer while self.lines_size < self.BUFFER_SIZE (65536). ReadLine() removes a line from self.lines but adds its length back to the counter instead of subtracting it:

```
def ReadLine(self):
    if not self.lines:
      self.ReadLines()

    line, _, self.lines = self.lines.partition('\n')
    self.lines_size += len(line) + 1   # <-- should be -=
    self.line_number += 1

    return line
```

ReadLine() is invoked on the plugin reject path: in plaso/parsers/text_plugins/interface.py, when _ParseString raises ParseError, the loop calls text_reader.ReadLine() to discard the offending line. Each rejected line therefore inflates lines_size by len(line)+1 instead of decreasing it — a swing of 2·(len(line)+1) away from the true value. Once enough rejected lines push lines_size above BUFFER_SIZE, ReadLines() stops refilling even though the real self.lines buffer has drained. while text_reader.lines: then goes false and parsing halts early, silently, well before the physical end of the file.

The documented invariant lines_size == len(lines) (per the class attribute docstring) is violated after the first rejected line. The successful path is unaffected because it consumes via SkipAhead(), which decrements the counter correctly.

**When running against test dataset, this caused `/var/log/apache2/access.log.1` to be skipped, causing 3,070,726 apache web server logs to be dropped.**

**To Reproduce:**
**The version of Plaso you used:**
20250918, 20260512 (also present on current main, 20260719 — the ReadLine line is unchanged there).

**The operating system you are running Plaso on:**
macOS 15 (Darwin 24.6.0), CPython 3.12.13. dfVFS 20260717, libfsext-python 20251107.  Docker and native python versions.

**Steps to reproduce the behaviour including command line and arguments and output:**
`--single-process --unattended --status_view linear --vss_stores all --partitions all --volumes all --hashers md5,sha1,sha256 --parsers win7_slow,winxp_slow,text,linux` against 20240606-defaced_ecomm-Ubuntu_22.04.3.E01 https://doi.org/10.21954/ou.rd.26038669.v1

The issue is in EncodedTextReader and reproduces deterministically without any image, using only the reader and a dfVFS fake file:

```
from dfvfs.path import fake_path_spec
from dfvfs.file_io import fake_file_io
from dfvfs.resolver import context as dfvfs_context
from plaso.parsers import text_parser

# ~10 MB of short lines: far larger than BUFFER_SIZE (64 KiB)
# and _READ_BUFFER_SIZE (1 MiB), so a correct reader must refill.
NUM_LINES = 200000
DATA = b'\n'.join([b'unparseable-recon-probe-line-' + b'x' * 20] * NUM_LINES) + b'\n'

ctx = dfvfs_context.Context()
path_spec = fake_path_spec.FakePathSpec(location='/file.txt')
file_object = fake_file_io.FakeFile(ctx, path_spec, DATA)
file_object.Open()

reader = text_parser.EncodedTextReader(file_object)
consumed = 0
reader.ReadLines()
while reader.lines:            # mirrors `while text_reader.lines:`
    reader.ReadLine()          # mirrors the ParseError reject path
    consumed += 1
    reader.ReadLines()         # plugin refills each iteration

print('consumed', consumed, 'of', NUM_LINES,
      '| final offset', reader.get_offset(), 'of', len(DATA),
      '| invariant', reader.lines_size == len(reader.lines))
```

**Output on the current release:**
```
consumed 20972 of 200000 | final offset 1048576 of 10000000 | invariant False
```

The reader stops after 20,972 of 200,000 lines, with its offset frozen at exactly 1,048,576 bytes — one _READ_BUFFER_SIZE (16 × 64 KiB) — and never refills.

Real-world trigger: log2timeline --parsers text/apache_access over a 3.3 GB access.log.1 (containing many probe/recon lines the apache_access grammar rejects) inside an Ubuntu 22.04.3 E01 image. The stock run stopped normally after 248,048 events and 207 rejected-line warnings, never reaching the file's 3,070,726-line end and emitting no pyfsext maximum-read warning. Direct dfVFS reads at offsets up to 3,311,080,000 all succeeded and the EXT file object reported the correct 3,311,081,154-byte logical size, ruling out the storage layer.

**The method you used to install Plaso:**
Docker container and also uv pip install plaso (PyPI release 20260512) into a CPython 3.12.13 virtualenv.

**Expected behavior:**
The parser reads to the true end of the file. Rejected lines should not affect refill behaviour. In the minimal reproduction, consumed should equal 200,000, final offset should equal the file size, and the lines_size == len(lines) invariant should hold after every ReadLine().

**Debug output/tracebacks:**
There is no traceback — the failure is a silent early stop, which is what makes it dangerous (events are dropped with no error). The observable signal is the frozen offset and broken invariant shown above. With the one-character fix below applied, the same reproduction prints:

```
consumed 200000 of 200000 | final offset 10000000 of 10000000 | invariant True
```

**Additional context**
Root cause and fix — change the += to -= in EncodedTextReader.ReadLine():

```
     line, _, self.lines = self.lines.partition('\n')
-    self.lines_size += len(line) + 1
+    self.lines_size -= len(line) + 1
     self.line_number += 1
```

This restores the lines_size == len(lines) invariant. lines_size is referenced only within text_parser.py (the refill gate in ReadLines and the loop in SkipAhead), so the change is fully contained. The existing test suites tests/parsers/text_parser.py (5 tests) and tests/parsers/text_plugins/ (114 tests) pass unchanged both before and after the fix — none of them assert on lines_size, which is why the regression went undetected. 

I will open a PR that includes the fix plus a regression test asserting the invariant across the read-buffer boundary and link it to this.

benjamindonnachie:
Fixes #5175

## Problem

`EncodedTextReader.ReadLine()` removes a line from `self.lines` but **adds**
its length to `self.lines_size` instead of subtracting it:

```python
line, _, self.lines = self.lines.partition('\n')
self.lines_size += len(line) + 1   # should be -=
```

`ReadLines()` only refills the buffer while `self.lines_size < self.BUFFER_SIZE`
(65536). On the plugin reject path (`text_plugins/interface.py` calls
`ReadLine()` on every `ParseError`), each rejected line inflates the counter
instead of shrinking it. Once enough rejected lines push `lines_size` past the
threshold, `ReadLines()` stops refilling even though the real buffer has
drained, `while text_reader.lines:` goes false, and parsing halts at a silent
false end-of-file — dropping events on large logs that contain unparsable
lines. The documented `lines_size == len(lines)` invariant is violated after
the first rejected line. The success path is unaffected because it consumes via
`SkipAhead()`, which decrements the counter correctly.

## Fix

Change the single `+=` to `-=`, restoring the invariant. `lines_size` is
referenced only within `text_parser.py` (the refill gate and `SkipAhead`), so
the change is fully contained.

## Tests

Adds two regression tests to `tests/parsers/text_parser.py`:

- `testReadLineUpdatesLinesSize` — asserts `lines_size == len(lines)` after
  each `ReadLine()` across the buffer.
- `testReadLineReadsEntireLargeFile` — a file larger than the read buffer,
  consumed line-by-line via `ReadLine()`, must reach true EOF.

Both fail on the current code and pass with the fix.

| suite | with fix | without fix |
| --- | --- | --- |
| `tests/parsers/text_parser.py` | 7/7 pass | 2 new tests fail |
| `tests/parsers/text_plugins/` (114 tests) | 114/114 pass | 114/114 pass |

No existing test asserts on `lines_size`, which is why the regression went
undetected.

## Impact observed

Reported in #5175: `log2timeline --parsers text/apache_access` over a 3.3 GB
`access.log.1` (3,070,726 lines, ~52k sqlmap probe lines the grammar rejects)
stopped after 248,048 events with no error. A minimal reproduction stalls at
exactly one `_READ_BUFFER_SIZE` (1,048,576 bytes) before the fix and reads to
true EOF after it.

### DIFF
--- a/plaso/parsers/text_parser.py
+++ b/plaso/parsers/text_parser.py
@@ -53,7 +53,7 @@ class EncodedTextReader:
             self.ReadLines()
 
         line, _, self.lines = self.lines.partition("\n")
-        self.lines_size += len(line) + 1
+        self.lines_size -= len(line) + 1
         self.line_number += 1
 
         return line
--- a/tests/parsers/text_parser.py
+++ b/tests/parsers/text_parser.py
@@ -88,6 +88,55 @@ class EncodedTextReaderTest(test_lib.ParserTestCase):
         text_reader.SkipAhead(10)
         self.assertEqual(text_reader.lines, self._TEST_LINES[10:])
 
+    def testReadLineUpdatesLinesSize(self):
+        """Tests that ReadLine keeps lines_size in sync with the buffer."""
+        # Newline-terminated so every line consumed by ReadLine includes its
+        # newline, which makes lines_size == len(lines) hold exactly.
+        test_data = b"\n".join([b"first line", b"second line", b"third line"])
+        test_data = b"".join([test_data, b"\n"])
+
+        resolver_context = dfvfs_context.Context()
+
+        test_path_spec = fake_path_spec.FakePathSpec(location="/file.txt")
+        file_object = fake_file_io.FakeFile(resolver_context, test_path_spec, test_data)
+        file_object.Open()
+
+        text_reader = text_parser.EncodedTextReader(file_object)
+
+        text_reader.ReadLines()
+        while text_reader.lines:
+            text_reader.ReadLine()
+            self.assertEqual(text_reader.lines_size, len(text_reader.lines))
+
+    def testReadLineReadsEntireLargeFile(self):
+        """Tests that ReadLine reads a large file to its end.
+
+        Regression test: a file larger than the read buffer, consumed
+        line-by-line via ReadLine (the plugin reject path), must be read to
+        its end. Inflating lines_size in ReadLine used to stall the refill in
+        ReadLines and produce a false end-of-file.
+        """
+        number_of_lines = 25000
+        test_data = b"\n".join([b"x" * 48] * number_of_lines) + b"\n"
+
+        resolver_context = dfvfs_context.Context()
+
+        test_path_spec = fake_path_spec.FakePathSpec(location="/file.txt")
+        file_object = fake_file_io.FakeFile(resolver_context, test_path_spec, test_data)
+        file_object.Open()
+
+        text_reader = text_parser.EncodedTextReader(file_object)
+
+        number_of_read_lines = 0
+        text_reader.ReadLines()
+        while text_reader.lines:
+            text_reader.ReadLine()
+            text_reader.ReadLines()
+            number_of_read_lines += 1
+
+        self.assertEqual(number_of_read_lines, number_of_lines)
+        self.assertEqual(text_reader.get_offset(), len(test_data))
+
 
 class TextLogParserTest(test_lib.ParserTestCase):
     """Tests for the text log parser."""


benjamindonnachie:
Apologies for the churn — I force-pushed `914ba9d` to fix the `black` formatting on the two new tests (the cause of the earlier `lint` failure). The one-character source fix is unchanged, and `black --check` now passes locally. This will need a fresh workflow approval — thanks!

joachimmetz:
Thanks for the proposed changes. Interesting find, taking a look later.

benjamindonnachie:
To keep the relevant resources together for review:

- **Root cause + minimal reproduction:** #5175
- **Real-world trigger:** a ~3.3 GB Apache `access.log.1` with heavy sqlmap probe traffic, inside a synthetic breach image I built for research. I noticed discrepancies in the parsed output that traced back to the `lines_size` accounting.
- **Dataset:** the image is public under CC-BY-NC-SA with ground truth I maintain — https://doi.org/10.21954/ou.rd.26038669.v1. Happy to share specifics if a real-world-scale text log with many unparsable lines is ever useful for regression or validation; no obligation.

joachimmetz (approved):
LGTM

joachimmetz (approved):
LGTM

Issue #5155, opened 2026-07-08
Document terminology for different event states/traits


joachimmetz:
https://github.com/log2timeline/plaso/issues/5140 introduced automatic tagging of events that are "recovered" or "corrupted" (or both). 

Different states/traits that are beneficial to capture and surface to analysts:

* corrupted; the parser encountered a format (edge) case where parsing failed. Note that this could be also related to the parsing configuration such as a code page. For now this includes certain edge cases, such as UCS-2 string conversion, where technically it is not corruption, given that due to technical reasons it is the parser that cannot handle the edge case
* recovered; the parser undertook steps to recover the event data outside "regular" (allocated) structures of the data format

Consider adding:

* dirty; the data format indicates its structures are "dirty", e.g. ESE databases
* recovered_from_journal or recovered_from_log: the parser undertook steps to recover the event data outside "regular" (allocated) structures of the data format using an accompanying journal or transaction log. What about SQLite WAL, should this be recovered_from_wal?
* incomplete; the parser encountered a format (edge) case it could not complete parse the event data, e.g. payload_type != xml in windows_push_notification parser. What about unsupported record types as in ios_app_privacy? or log action in xchatlog.py?
* stored_external/stored_remote; To indicate files with content stored remotely (network or Cloud)?

PR #5180, opened 2026-07-21
Add structured execution, file and user fields to the auditd parser

kev365:
Tier 2 of the auditd (`audit.log`) parsing enhancement from #5156, following the
merged Tier 1 (#5172). Extends `text_plugins/selinux.py` (the generic `audit.log`
parser) with the higher-value, record-type-specific fields — the executed command
line, the file touched, and the who/where/outcome of authentication events.

### Added attributes (all `str`, promoted from the parsed body)

- **Execution** — `arguments` (EXECVE `argc`/`a0..aN`, hex-decoded, space-joined),
  `proctitle` (hex-decoded, NUL separators rendered as spaces), `architecture`
  (`arch`; the resolved `ARCH` name when ENRICHED), `audit_rule_key` (`key`).
- **File / path** — `file_path` (`name`, hex-decoded when bare), `working_directory`
  (`cwd`), `file_mode` (`mode`), `owner_user_identifier` (`ouid`),
  `owner_group_identifier` (`ogid`), `name_type` (`nametype`).
- **User / auth** — `account` (`acct`), `result` (`res`), `operation` (`op`),
  `terminal`, `remote_address` (`addr`), `remote_hostname` (`hostname`) — read from
  the nested single-quoted `msg='…'` of USER_* records.

### A few decisions

- **Hex decode is byte-preserving and per-field.** Only EXECVE `a0..aN`,
  `proctitle`, and PATH `name` are hex-decoded, discriminated quoted-vs-bare from the
  raw text (so a quoted literal that happens to be all-hex is never mis-decoded);
  `arch`, SYSCALL pointer args, `mode`, `inode` are untouched. Decode uses
  `errors="backslashreplace"` with a `ProduceExtractionWarning`, and an odd-length
  (not validly hex) token is preserved as a literal rather than raising.
- **Nested `msg='…'`** — the grammar now captures the single-quoted USER_* blob as
  one token and re-parses it for the auth fields; auditd sentinels (`?`, `(null)`,
  `(none)`) map to `None`.
- **ENRICHED** — the resolved `ARCH` name is surfaced the same way Tier 1 surfaces
  the resolved `SYSCALL` name; raw numeric ids are kept for offline soundness.

### Backward compatibility

- **No new events** — each `audit.log` line stays one `selinux:line` event; only
  attributes are added, so counts and event identity are unchanged and `None`
  attributes are not serialized.
- No attribute removed or renamed; `message_body` is retained (it still carries the
  fields not yet promoted). No new `*_time`, and the message template,
  `tag_linux.txt`, and `timeliner.yaml` are unchanged — the new attributes are for
  filtering and structured output, matching the `systemd:journal` field set.

### Testing

Unit tests over a new RAW slice (`test_data/audit.log`, Ubuntu/AppArmor) and the
existing ENRICHED `audit_enriched.log`, with expected values derived from the
`ausearch -i` oracle. Validated on Ubuntu with GIFT-PPA deps (matching CI): unit
tests, `psteal` (0 extraction warnings), and the full end-to-end suite including the
`ext4_with_binaries` over-claim guard — all green (the parser produces no events
from the binary image).


plaso/parsers/text_plugins/selinux.py
```diff
@@ -28,48 +30,86 @@ class SELinuxLogEventData(events.EventData):
...
           user identifier that is retained across su and sudo.
+      audit_rule_key (str): audit rule key (key) identifying the rule that
+          triggered the record.
       audit_serial (int): audit serial number, used to correlate the records
           that belong to a single audited event.
       audit_session_identifier (str): audit session identifier (ses).
       audit_type (str): audit type.
       executable (str): path of the executable (exe).
       exit_code (str): exit status of the system call (exit).
+      file_mode (str): file mode (mode) of the path in a PATH record.
```
joachimmetz: @kev365 why not make this an integer?

plaso/parsers/text_plugins/selinux.py
```diff
@@ -28,48 +30,86 @@ class SELinuxLogEventData(events.EventData):
...
       message_body (str): message body.
+      name_type (str): type of the path reference (nametype), such as NORMAL,
+          PARENT, CREATE or DELETE.
+      operation (str): operation (op) performed by a user or auth event.
+      owner_group_identifier (str): group identifier that owns the file (ogid).
+      owner_user_identifier (str): user identifier that owns the file (ouid).
       parent_process_identifier (str): parent process identifier (ppid).
       pid (str): process identifier (PID) that created the SELinux log line.
       process_name (str): name of the process (comm).
+      proctitle (str): process title (proctitle), hex-decoded, with the NUL
```
joachimmetz: Per style guide, please do not use abbreviations, maybe change this to process_title or equivalent.

plaso/parsers/text_plugins/selinux.py
```diff
@@ -28,48 +30,86 @@ class SELinuxLogEventData(events.EventData):
...
+          argument separators rendered as spaces.
+      remote_address (str): source address (addr) of a remote auth event.
+      remote_hostname (str): source hostname (hostname) of a remote auth event.
+      result (str): result (res) of a user or auth event, such as success or
```
joachimmetz: Would recommend making this more specific, maybe authentication_result or authentication_status?

plaso/parsers/text_plugins/selinux.py
```diff
@@ -28,48 +30,86 @@ class SELinuxLogEventData(events.EventData):
...
+          failed.
       security_context (str): security context (subj) of the process, such as a
           SELinux or AppArmor label.
       success (str): whether the system call succeeded (success).
       system_call (str): system call (syscall).
+      terminal (str): controlling terminal (terminal) of a user event.
       user_identifier (str): user identifier (uid) of the process.
+      working_directory (str): process working directory (cwd) at exec time.
```
joachimmetz: exec time => execution time ?

plaso/parsers/text_plugins/selinux.py
```diff
@@ -28,48 +30,86 @@ class SELinuxLogEventData(events.EventData):
     """SELinux log event data.
 
     Attributes:
+      account (str): account (acct) targeted by a user or authentication event.
```
joachimmetz: Maybe add a bit of context, how is this different from user_identifier and what this value contains, e.g. a username, identifier, other?

plaso/parsers/text_plugins/selinux.py
```diff
@@ -114,6 +167,103 @@ class SELinuxTextPlugin(interface.TextPlugin):
...
+
+        Returns:
+          str: decoded value; bytes that are not valid UTF-8 are kept as escaped
+              byte values. An odd-length (hence not validly hex-encoded) value is
+              returned unchanged.
+        """
+        if len(hex_value) % 2 != 0:
+            return hex_value
+
+        decoded_bytes = bytes.fromhex(hex_value)
```
joachimmetz: what if the value contains non-base16 values? Please catch the corresponding exception and create an extraction warning.

plaso/parsers/text_plugins/selinux.py
```diff
@@ -114,6 +167,103 @@ class SELinuxTextPlugin(interface.TextPlugin):
...
+        try:
+            return decoded_bytes.decode("utf-8")
+        except UnicodeDecodeError:
+            parser_mediator.ProduceExtractionWarning(
+                f"unable to decode hex-encoded audit value: {hex_value:s}"
+            )
+            return decoded_bytes.decode("utf-8", errors="backslashreplace")
```
joachimmetz: Please return a bool to indicate the event data was "corrupted". For context see:

* https://github.com/log2timeline/plaso/issues/5155
* https://github.com/log2timeline/plaso/commit/e679c92f29cb8e2332e7bac32a3198453805a153#diff-8e12defd1f528f3774342b63f0fe8b0e0eef61413792e1bef34379b77ba9d751R854 

(side note working on better documentation for this, if questions let me know)

plaso/parsers/text_plugins/selinux.py
```diff
@@ -114,6 +167,103 @@ class SELinuxTextPlugin(interface.TextPlugin):
...
+        Args:
+          parser_mediator (ParserMediator): mediates interactions between parsers
+              and other components, such as storage and dfVFS.
+          body (str): raw message body.
+          field_name (str): audit field name, such as "proctitle" or "name".
+
+        Returns:
+          str: decoded value, or None if the field is not present.
+        """
+        match = re.search(rf"\b{field_name:s}={self._HEX_CAPABLE_VALUE:s}", body)
```
joachimmetz: Instead of using a regular expression to search for the value, why not extract all the values temporarily in a dict?

joachimmetz:
@kev365 thanks for the proposed change, I left some comments and questions. Let me know if you have questions, my availability is intermittent

kev365:
Thanks for the review — these are all addressed and pushed, with a couple of naming questions below. Your `res` question sent me down a bit of a rabbit hole and turned up a few things, so this is a longer reply than I'd like.

**On `res` naming.** The [linux-audit field dictionary](https://github.com/linux-audit/audit-documentation/blob/master/specs/fields/field-dictionary.csv) describes `res` as "result of the audited operation(success/fail)" and `op` as "the operation being performed that is audited" — both generic rather than auth-specific. That matches what's in the logs: `res=` shows up on SERVICE_START/SERVICE_STOP, CONFIG_CHANGE, DAEMON_START, SYSTEM_BOOT and CRYPTO_* alongside the USER_* records, and `op=` is most common on BPF (`LOAD`/`UNLOAD`).

Would `operation_result` work? It pairs with the `operation` (`op`) attribute and still reads right on the non-auth records. Happy to use `authentication_result` if you prefer.

A related question on its type: `res` comes in two forms — `success`/`failed` on most record types, and `1`/`0` on CONFIG_CHANGE and LOGIN. Both are in the test data: `audit_enriched.log` has `type=USER_AUTH … res=failed` (serial 441) and `type=LOGIN … res=1` (serial 447). I've kept it as the literal string so nothing is lost, but since both forms are really booleans, would you rather it were normalized to a `bool` per the style guide? Happy either way.

**A few bugs the same digging turned up**, all fixed here:

- I was only reading `op`/`res`/`acct` out of the nested `msg='…'`, so I missed the top-level ones (BPF, CONFIG_CHANGE, NETFILTER_CFG, LOGIN, DAEMON_START — a few hundred records in my test corpus). The reverse was true for `exe` and `comm`: I only read those at the top level and so missed them inside `msg='…'` on USER_*/SERVICE_*/CRED_*/CRYPTO_*. Those last two are tier 1 attributes, so this PR does touch them — the both-locations lookup is a single change and splitting it would have meant leaving a known gap in place and writing half the logic twice. Happy to pull it into its own PR if you'd rather.
- `acct` could store the `?` sentinel instead of `None` (USER_ERR records); it now goes through the same sentinel handling as `terminal`/`addr`/`hostname`.
- The warning call itself was wrong — I had `ProduceExtractionWarning`, which no longer exists after #5161, so it would have raised an `AttributeError` on the first value it could not decode. Now `ProduceWarning`. Worth noting the codecov gap on this PR was pointing straight at it: those lines were the uncovered ones, so nothing ever ran them. There is a small test fixture for the malformed cases now (odd number of hex digits, invalid UTF-8, bad mode, bad argc), which also covers the corrupted flag.

Also from the dictionary: `acct`, `cwd` and `key` are "encoded" fields, so they can arrive hex-encoded and I wasn't decoding them. Now they go through the same decode path — as do `comm` and `exe`, via the change above.

**Two naming things I noticed but didn't want to change unilaterally:**

- `account` (`acct`) — the style guide prefers `username`, and the dictionary calls `acct` "a user's account name". What gave me pause: `acct` is the *target* account of the event, while `user_identifier` (`uid`) is the acting process's user — two different principals on the same record. Do you want me to rename `account` to `username`, or keep it as `account`?
- `arguments` (EXECVE argv) — `command_line` looks closer to existing convention (`systemd:journal` uses `command_line`, santa uses `process_arguments`). Do you want me to rename `arguments` to `command_line`?

The rest is implemented as suggested — and thanks for the corrupted flag pointer to #5155

joachimmetz:
> Would operation_result work? It pairs with the operation (op) attribute and still reads right on the non-auth records. Happy to use authentication_result if you prefer.

operation_result is fine, authentication_result was a only a suggestion.

> I've kept it as the literal string so nothing is lost, but since both forms are really booleans, would you rather it were normalized to a bool per the style guide?

Yes, please normalize them into a bool and mention in the event data that log format represents them as success/failed or (I assume) 0/1 receptively

> Happy to pull it into its own PR if you'd rather.

Separate PR is fine

> acct could store the ? sentinel instead of None (USER_ERR records); it now goes through the same sentinel handling as terminal/addr/hostname.

Not sure I understand what you are trying to tell me. Why would you want to present the user with the sentinel value?

> I had ProduceExtractionWarning, which no longer exists

Good catch

> Two naming things I noticed but didn't want to change unilaterally:

Unfortunately unambiguous naming concepts is not straightforward. A longer term idea is to use https://github.com/log2timeline/dfkinds and formalize (some form of) ontology.

> account (acct) — the style guide prefers username, and the dictionary calls acct "a user's account name".

Keeping it account is fine. The idea is to stick with the style guide as much as possible, but there will be (edge) cases that fall outside it. My review comment was mostly around adding additional context to the documentation string:

```
Maybe add a bit of context, how is this different from user_identifier and what this value contains, e.g. a username, identifier, other?
```

> arguments (EXECVE argv) — command_line looks closer to existing convention (systemd:journal uses command_line, santa uses process_arguments). Do you want me to rename arguments to command_line?

Good question, conceptually they are comparable (not equivalent)

* command_line / command_line_arguments; command (and arguments) provided via a CLI, often what a human types or script enters into a terminal
* process_arguments; arguments of a running process

IMHO process_arguments is likely more appropriate in this context

kev365:
Thanks, that all helps. Updated:

- `operation_result` is a `bool` now. A couple of refinements came out of libauparse: `res` is `AUPARSE_TYPE_SUCCESS` in `typetab.h`, and `interpret.c` maps 1 to "yes" and 0 to "no", with anything else falling through to "unset". So `success`/`yes`/`1` become True, `failed`/`no`/`0` become False, and `4294967295` (the unset marker) becomes None without a warning. An unrecognized value gets an extraction warning and marks the event data corrupted. Nothing fell outside that mapping across ~800 records of captured audit logs. The attribute documentation mentions both representations.
- `arguments` is now `process_arguments`, agreed on the distinction you drew.
- `account` keeps its name, and the documentation now says it is the name of the account the event acts on, such as the account being authenticated, as opposed to `user_identifier` being the numeric user identifier of the process that caused the event.
- Tier 1 (`exe`/`comm` from the nested `msg`) is out of this PR again, so it only touches tier 2 attributes. I will send that as its own PR, together with a few related tier 1 items the same digging turned up.

On the sentinel question, sorry, I wrote that badly. I was not suggesting we present the sentinel to the user; it was a bug on my side. `acct=?` was being stored as the literal string "?" rather than None, and it now goes through the same handling as the other sentinel values.

The docs job is red on an unrelated linkcheck: sphinx could not reach `https://launchpad.net/~gift` from `sources/developer/Development-Dependencies`. Everything else is green.

One other thing I ran into while checking the field dictionary against real logs: an audit rule can have more than one key, and auditd stores them in a single `key=` field separated by AUDIT_KEY_SEPARATOR (0x01), hex-encoding the field as a result. I was storing that as one string with a 0x01 byte in it. `auditctl -w /tmp/keytest -p wa -k alpha -k beta` produces `key=616C7068610162657461`, which is "alpha" 0x01 "beta". So `audit_rule_key` is now `audit_rule_keys` as a `list[str]`, split on the separator, and there is a real multi-key record in the test data for it. Happy to keep it a single string if you would rather.

joachimmetz:
Regarding audit_rule_keys list of strings is fine, this is a type supported by the attribute containers.

> The docs job is red on an unrelated linkcheck: sphinx could not reach https://launchpad.net/~gift from sources/developer/Development-Dependencies. Everything else is green.

You can ignore it, the docs tests unfortunately are sometimes flaky due to connectivity issues.

joachimmetz (approved):
LGTM

PR #5182, opened 2026-07-28
Improve auditd field handling in the SELinux parser

kev365:
Follow-up to #5180, as discussed there: the tier 1 auditd attribute changes that were
kept out of that PR, plus a few related field-handling corrections found while
comparing the parser against the linux-audit field dictionary and libauparse.

### `exe` and `comm` are also stored in a nested `msg`

`text_plugins/selinux.py` read `exe` and `comm` only from the top level of the message
body, but USER_*, SERVICE_*, CRED_* and CRYPTO_* records store them inside the nested
single-quoted `msg='…'` instead, so `executable` and `process_name` were not extracted
for those record types at all. Against two captured audit logs of ~400 records each,
`executable` now populates on 139 of 139 and 81 of 81 of those records respectively,
where it previously populated on none of them, and `process_name` likewise populates on
the service records that carry it.

#5180 added the merged value lookup that makes this a one-line change per attribute;
this PR switches `exe` and `comm` over to it. No record type stores either field in
both places, so there is no ambiguity.

### `exe`, `comm` are "encoded" fields

Per the linux-audit
[field dictionary](https://github.com/linux-audit/audit-documentation/blob/master/specs/fields/field-dictionary.csv)
both are `encoded`, meaning auditd hex-encodes the value when it contains characters
that would need escaping (`AUPARSE_TYPE_ESCAPED` in libauparse's `typetab.h`). They now
go through the same byte-preserving decode as the other encoded fields.

### Field types aligned with libauparse

- **`success`** to `bool` — the same normalization applied to `operation_result` in
  #5180. libauparse maps 1 to "yes" and 0 to "no" for `AUPARSE_TYPE_SUCCESS`, and the
  log format writes `yes`/`no` here.
- **`exit_code`** to `int` — `exit` is `AUPARSE_TYPE_EXIT`. The raw numeric value is
  kept rather than translated, since libauparse renders negative values as errno names
  (`-13` as `EACCES`), which is host and architecture specific and better left to
  analysis.
- **`audit_login_identifier`** and **`audit_session_identifier`** map `4294967295` to
  `None`. That value is `(uint32_t) -1`, which libauparse renders as "unset", so it is
  an absent value rather than a real identifier.

### Backward compatibility

No new events and no attribute removed or renamed. `success` and `exit_code` change
type, and the three attributes above populate on more record types than before. The
verification grammar is unchanged, so the plugin claims exactly the same files.

### Testing

Unit tests over the existing RAW and ENRICHED fixtures, with expected values derived
from the `ausearch -i` oracle. Validated on Ubuntu with GIFT-PPA dependencies
(matching CI): unit tests, `psteal`, and the full end-to-end suite including the
`ext4_with_binaries` over-claim guard.

### DIFF
--- a/plaso/parsers/text_plugins/selinux.py
+++ b/plaso/parsers/text_plugins/selinux.py
@@ -36,15 +36,17 @@ class SELinuxLogEventData(events.EventData):
       architecture (str): CPU architecture (arch); the resolved name (e.g.
           "x86_64") when the record is ENRICHED, otherwise the raw value.
       audit_login_identifier (str): audit login identifier (auid), the login
-          user identifier that is retained across su and sudo.
+          user identifier that is retained across su and sudo, or None if unset.
       audit_rule_keys (list[str]): keys (key) of the audit rule that triggered
           the record, where a rule can have multiple keys.
       audit_serial (int): audit serial number, used to correlate the records
           that belong to a single audited event.
-      audit_session_identifier (str): audit session identifier (ses).
+      audit_session_identifier (str): audit session identifier (ses), or None
+          if unset.
       audit_type (str): audit type.
       executable (str): path of the executable (exe).
-      exit_code (str): exit status of the system call (exit).
+      exit_code (int): exit status of the system call (exit), where a negative
+          value represents an errno value.
       file_mode (int): file mode (mode) of the file, which includes the file type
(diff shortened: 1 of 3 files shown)


joachimmetz (approved):
LGTM

PR #5186, opened 2026-07-30
Correction for output mediator passed to WriteHeader #5184

### LINKED ISSUE #5184: psort -o kml fails with AttributeError in WriteHeader
Here is a random bug that came up in some testing.

## Summary

`psort.py -o kml` fails with an `AttributeError`. `ExportEvents` passes the
`plaso.output.mediator` module to `WriteHeader` instead of the `OutputMediator`
instance it created.

`plaso/multi_process/output_engine.py`:

```python
from plaso.output import mediator as output_mediator   # line 13
...
    self._output_mediator = self._CreateOutputMediator(
        storage_reader, processing_configuration)
    output_module.WriteHeader(output_mediator)          # line 560
```

`ExportEvents` has no local or parameter named `output_mediator`, so the bare name
resolves to the module-level import.

## Reproduction

```
$ psort.py -o kml -w out.kml test_data/psort_test.plaso
  File "plaso/multi_process/output_engine.py", line 560, in ExportEvents
    output_module.WriteHeader(output_mediator)
  File "plaso/output/kml.py", line 74, in WriteHeader
    f'<?xml version="1.0" encoding="{output_mediator.encoding:s}"?>'
AttributeError: module 'plaso.output.mediator' has no attribute 'encoding'
```

`-o dynamic` against the same storage file succeeds.

## Scope

`kml` is the only output module that dereferences the `WriteHeader` argument
(`plaso/output/kml.py:74`); the others ignore it. `psteal.py` uses the same engine
(`plaso/cli/psteal_tool.py:407-409`).

The end-to-end configurations cover `dynamic`, `json`, `json_line`, `l2tcsv`,
`l2ttln`, `null`, `opensearch`, `opensearch_ts`, `rawpy`, `tln` and `xlsx`, but not
`kml`.

## Fix

```python
output_module.WriteHeader(self._output_mediator)
```

Happy to send a PR.

kev365:
Fixes #5184.

`ExportEvents` passed the `plaso.output.mediator` module to `WriteHeader` instead of
the `OutputMediator` it had just created, so `psort.py -o kml` failed with
`AttributeError: module 'plaso.output.mediator' has no attribute 'encoding'`.

Also adds the end-to-end test for the `kml` output format, modelled on
`output_rawpy`.

A note on the reference output: `KMLOutputModule.WriteFieldValues` returns early for
events without latitude and longitude, and `test_data/psort_test.plaso` has no
geolocation data, so `test_data/end_to_end/kml.log` is the header and footer only:

```
<?xml version="1.0" encoding="utf-8"?><kml xmlns="http://www.opengis.net/kml/2.2"><Document></Document></kml>
```

That covers the defect, since it is in `WriteHeader`. It does not cover the placemark
path in `WriteFieldValues`, and there is currently no clean way to: `android_airtag`
appears to be the only parser that populates `latitude` and `longitude`, and the
AirGuard `beacon` table it reads stores latitude in the column named `longitude` and
longitude in the column named `latitude`
([reference](https://thebinaryhick.blog/2022/01/08/androids-airtags-oof/)). The plugin
does not compensate for that, so a placemark reference generated from
`test_data/android/attd_db` would record reversed coordinates. Noted for the record,
not addressed here.

Verified with the full end-to-end suite on Ubuntu 26.04, exit 0, including the new
`output_kml` case.

### DIFF
--- a/config/end-to-end.ini
+++ b/config/end-to-end.ini
@@ -262,6 +262,13 @@ output_format=json_line
 reference_output_file=test_data/end_to_end/json_line.log
 source=test_data/psort_test.plaso
 
+[output_kml]
+case=analyze_and_output
+output_file=kml.log
+output_format=kml
+reference_output_file=test_data/end_to_end/kml.log
+source=test_data/psort_test.plaso
+
 [output_l2tcsv]
 case=analyze_and_output
 output_file=l2tcsv.log
--- a/plaso/multi_process/output_engine.py
+++ b/plaso/multi_process/output_engine.py
@@ -557,7 +557,7 @@ class OutputAndFormattingMultiProcessEngine(engine.MultiProcessEngine):
         self._output_mediator = self._CreateOutputMediator(
             storage_reader, processing_configuration
         )
-        output_module.WriteHeader(output_mediator)
+        output_module.WriteHeader(self._output_mediator)
 
         self._StartStatusUpdateThread()
 
--- a/test_data/end_to_end/kml.log
+++ b/test_data/end_to_end/kml.log
@@ -1,0 +1,1 @@
+<?xml version="1.0" encoding="utf-8"?><kml xmlns="http://www.opengis.net/kml/2.2"><Document></Document></kml>


joachimmetz (approved):
LGTM

joachimmetz:

Issue #3714, opened 2021-06-14
Remove SQLite append event tag backwards compatibility


joachimmetz:
After implementing storing event tags only once per event https://github.com/log2timeline/plaso/issues/3601

* remove SQLite append event tag backwards compatibility
* remove EventTagIndex
  * currently used in output and analysis engine, replace by reader.GetEventTagByEventIdentifer method, since event tags map to a single event

PR #5187, opened 2026-07-31
Correction to resolve the event tag once per exported event #5185

### LINKED ISSUE #5185: Event tag is resolved twice for every exported event
## Summary

The output and formatting engine resolves the event tag twice for every exported
event: once in `_ExportEvents` for the event filter
(`plaso/multi_process/output_engine.py:256`) and again in `_FlushExportBuffer` for the
output module (`:371`).

`event_data` and `event_data_stream` are already fetched once and carried through
`PsortEventHeap` to the flush:

| container | fetched | carried on the heap | re-fetched at flush |
| --- | --- | --- | --- |
| `event_data` | `:243` | yes (`:196` -> `:359`) | no |
| `event_data_stream` | `:249` | yes (`:196` -> `:359`) | no |
| `event_tag` | `:256` | no | yes (`:371`) |

The event tag is the only one not carried.

## Measured

Instrumented over 134,499 events extracted from a real `/var/log`:

```
events written:    133,167
tag lookups:       267,666
lookups per event: 2.01
```

The count is exact: 133,167 written x 2, plus 1,332 deduplicated x 1, equals 267,666.
Deduplicated events cost one lookup because the deduplication `continue` at `:362`
fires before the second fetch.

The store is opened read-only and nothing writes between the two calls, so the second
call returns the same container as the first. The first result is only consumed when
an event filter is set (`:261-266`).

`psteal.py` uses the same engine (`plaso/cli/psteal_tool.py:407-409`) and is affected
identically.

## Cost

`GetEventTagByEventIdentifer` (`plaso/storage/reader.py:110-134`) builds a filter
expression per call, which on the SQLite path is parsed with `ast.parse` and converted
by `PythonAST2SQL` (`plaso/storage/sqlite/sqlite_file.py:410-413`).

acstore skips the query when there are no containers of that type
(`acstore/sqlite_store.py:445`), but that check happens after the expression is built,
so exporting a storage file that has never been tagged still pays the full Python cost
per lookup for a query that never executes. Measured at 41.8 µs per lookup on an
untagged file.

## Origin

Both call sites were introduced together in 65fb69861, "Removed event tag index #3714
(#4154)", which replaced `EventTagIndex.GetEventTagByIdentifier` with
`reader.GetEventTagByEventIdentifer` at each site, as #3714 describes.

Under the previous `EventTagIndex` a lookup was a dict hit followed by a
fetch-by-index that went through the attribute container cache, and returned `None`
without touching the database when the file had no tags. Calling it twice was
therefore close to free. The duplication predates the cost rather than being
introduced by it.

If retaining both resolution points was deliberate, or if there was a different
follow-up intended for #3714, I am happy to defer to that.

## Proposed fix

Resolve the event tag once and carry it through `PsortEventHeap` to the flush, the
same way `event_data` and `event_data_stream` are already carried. Where it is
resolved depends on whether the filter needs it:

- with an event filter, resolve in the read loop, since the filter can match on labels;
- without one, resolve at the flush, after deduplication, so duplicate events resolve
  no tag at all.

Timed by exporting the same 134,499 event corpus twice in immediate succession on the
same machine, once with current `main` and once with the change, since separate runs
drift by a second or so (the unmodified baseline measured 45.59 s and 44.43 s on two
occasions):

| run | lookups/event | wall clock |
| --- | --- | --- |
| current | 2.01 | 44.43 s |
| resolved once | 1.00 | 39.97 s |

Roughly 10 % of export. Both runs used an output module that discards its input, so
the difference is in the storage and engine layers rather than in formatting.

## Tests

`psort.py -o dynamic` output is byte-identical before and after against
`test_data/psort_test.plaso`.

All 7 tests in `tests/multi_process/output_engine.py` pass unmodified: giving
`PushEvent` an `event_tag=None` default keeps existing callers working, and the heap
tests assert heap length rather than the shape of the popped tuple.

Happy to send a PR.

kev365:
Fixes #5185.

Resolves the event tag once per exported event instead of twice.

## Where to look

The two call sites on `main`, both `storage_reader.GetEventTagByEventIdentifer`:

- `plaso/multi_process/output_engine.py:256`, in `_ExportEvents`, used for the event
  filter and only consumed when one is set;
- `plaso/multi_process/output_engine.py:371`, in `_FlushExportBuffer`, used for the
  output module.

Both were introduced together in 65fb69861, "Removed event tag index #3714 (#4154)",
which replaced `EventTagIndex.GetEventTagByIdentifier` with
`reader.GetEventTagByEventIdentifer` at each site. In the removed
`plaso/storage/event_tag_index.py` a lookup was a dict hit followed by
`GetAttributeContainerByIdentifier`, and returned `None` without touching the database
when the storage file had no event tags, so calling it twice was close to free.

## Change

The event tag is now resolved in one of two places, depending on whether the event
filter needs it:

- with an event filter, in the read loop, since a filter can match on labels, and
  carried through `PsortEventHeap` to the flush the same way `event_data` and
  `event_data_stream` already are;
- without one, when the export buffer is flushed, after duplicate events have been
  discarded, so deduplicated events no longer resolve a tag at all.

Events released from the time slice buffer are resolved at the point they are
exported, since they do not pass through the read loop.

Output is unchanged in both cases. The second point is a behaviour change in that a
deduplicated event no longer causes a lookup, but such events are never written, so
nothing reaches an output module that did not before.

## Effect

Measured against a storage file of 1,296,316 events extracted from a disk image, of
which 1,293,801 are exported and 2,515 are discarded as duplicates. Both versions were
run alternately on the same idle machine, twice each, so that run to run variation is
visible rather than assumed:

| run | lookups per exported event | wall clock |
| --- | --- | --- |
| before | 2.00 | 328.9 s, 355.0 s |
| after | 1.00 | 285.8 s, 295.1 s |

That is 342.0 s against 290.4 s on average, or about 15 percent of export, for 1.30
million lookups removed.

The proportion depends on what is being exported. The measurement in the issue, over
134,499 events extracted from `/var/log` where the event data is cheaper to
deserialise and the lookup is therefore a larger share of the total, came out closer
to 10 percent. The number of lookups removed per exported event is one in either case.

Note that these storage files contain no event tags, so each removed lookup is the
inexpensive path: the filter expression is still built and parsed, but
`_GetAttributeContainersWithFilter` returns without querying when there are no
containers of that type. On a storage file that has been tagged the removed lookup
also avoids a query and the deserialisation of a container.

## Tests

`tests/multi_process/output_engine.py` passes unmodified. Giving `PushEvent` an
`event_tag=None` default keeps existing callers working.

`psort.py -o dynamic` output is byte identical before and after against
`test_data/psort_test.plaso`, with no filter, with a filter on `timestamp_desc`, and
with a filter on `tag`, the last of which exercises resolving the tag before the
filter is applied.

Full end-to-end suite on Ubuntu 26.04, exit 0.


kev365:
Haven't dug into it much, but the failing check here (`testProcessSource`, `5 != 15`) looks unrelated to this change — it passes with dfdatetime 20260411 and fails with 20260730.

A quick look points at `ConvertDateTimeValuesToJSON` now going via `CopyToSerializableDict` (#309) — dfVFS's `TSKTime` doesn't implement it, so a TSK `modification_time` comes back `null` after the task store round trip. I could well be missing something, that's as far as I took it.

joachimmetz:
I opt to change the path spec type form TSK to EXT for now. Will take a closer look at the TSK timestamp when time permits https://github.com/log2timeline/plaso/issues/5189

joachimmetz:
Changes in https://github.com/log2timeline/plaso/pull/5190 to address the pre-release compatibility issue

plaso/multi_process/output_engine.py
```diff
@@ -117,6 +132,7 @@ def __init__(self):
         self._number_of_consumed_events = 0
         self._output_mediator = None
         self._processing_configuration = None
+        self._resolve_event_tag_at_flush = True
```
joachimmetz: Instead of tracking this state at the object level, I opt to pass this as an argument to _FlushExportBuffer e.g. resolve_event_tag = True/False

joachimmetz:
Thanks for the proposed changes, one suggestion

plaso/multi_process/output_engine.py
kev365: Done.

One choice worth flagging: `resolve_event_tag` defaults to True on both `_ExportEvent` and `_FlushExportBuffer`. `_ExportEvents` always passes it explicitly so the default only affects other callers, but True costs at most a redundant lookup where False would leave the event tag unresolved. Happy to flip it if you prefer False.

`psort -o dynamic` output is byte identical before and after, without a filter and with filters on `timestamp_desc` and `tag`.

joachimmetz: Thanks,. It does the safe thing by default, that is fine.

joachimmetz (approved):
LGTM

PR #5192, opened 2026-08-02
Correction to storage profiler write sample data size #5191

### LINKED ISSUE #5191: Storage profiler write samples report the compressed size in the data size column
## Summary

`--profilers storage` writes a `Data size` and a `Compressed data size` column per
sample. On the write path both columns carry the compressed length, so the
uncompressed size is not recoverable from a profiling run.

`plaso/storage/sqlite/sqlite_file.py:303-316`:

```python
if self.compression_format == definitions.COMPRESSION_FORMAT_ZLIB:
    compressed_data = zlib.compress(serialized_data)
    serialized_data = sqlite3.Binary(compressed_data)   # rebound to the compressed blob
else:
    compressed_data = ""

if self._storage_profiler:
    self._storage_profiler.Sample(
        "write_new",
        "write",
        container.CONTAINER_TYPE,
        len(serialized_data),      # compressed length
        len(compressed_data),      # same compressed length
    )
```

`serialized_data` is rebound at `:305` before the sample at `:314` reads it for the
`data_size` argument. `StorageProfiler.Sample`
(`plaso/engine/profilers.py:199-224`) documents the two as `data_size` and
`compressed_data_size` and writes them to separate columns.

The read path keeps them distinct: `_CreateAttributeContainerFromRow`
(`plaso/storage/sqlite/sqlite_file.py:88-102`) decompresses into `serialized_data`
and leaves `compressed_data` holding the stored blob, so its samples pair correctly.

## Observed

Three `event_data` containers written with a storage profiler attached:

| | Data size | Compressed data size | Ratio |
| --- | --- | --- | --- |
| current | 107 | 107 | 1.00 |
| expected | 1266 | 107 | 0.08 |

On a real run — 836,047 `event_data` containers extracted from a 5.7 GB image — the
read side reported 668.8 MB -> 342.0 MB (51 %) for the same data that the write side
reported as 344.8 MB -> 344.8 MB (1.00).

The value written to the `_data` column is the compressed blob in both cases, so
storage file contents are unaffected. This is limited to the profiler samples.

One related detail in the same block: `compressed_data` is initialised to `""` in the
non-compressed branch, where the read path at `:92` uses `b""`. `len()` is 0 either
way.

## Proposed fix

Keep `serialized_data` bound to the serialized container and bind the value written to
the `_data` column to a separate name, so the sample reads the uncompressed length:

```python
if self.compression_format == definitions.COMPRESSION_FORMAT_ZLIB:
    compressed_data = zlib.compress(serialized_data)
    column_value = sqlite3.Binary(compressed_data)
else:
    compressed_data = b""
    column_value = serialized_data
```

## Tests

`tests/storage/sqlite/sqlite_file.py` has no case that attaches a storage profiler, so
nothing exercised this path. A new
`testWriteNewAttributeContainerWithStorageProfiler` writes an `event_data` container
through a real `StorageProfiler`, reads the resulting `storage-*.csv.gz`, and asserts
the data size exceeds the compressed data size. It fails on current `main` with
`89 not greater than 89`.

Going forward with the PR.

kev365:
Fixes #5191.

`_WriteNewAttributeContainer` rebound `serialized_data` to the compressed blob before
sampling it, so the storage profiler's `write_new` samples carried the compressed
length in both the `Data size` and the `Compressed data size` column. The value bound
for the `_data` column is now a separate name, leaving `serialized_data` holding the
serialized container when the sample is taken.

| | Data size | Compressed data size |
| --- | --- | --- |
| before | 107 | 107 |
| after | 1266 | 107 |

`compressed_data` in the non-compressed branch is also `b""` now, matching
`_CreateAttributeContainerFromRow`.

Storage file contents are unchanged — the compressed blob is still what is written to
the `_data` column, and the read path was already correct.

`tests/storage/sqlite/sqlite_file.py` had no case that attached a storage profiler.
`testWriteNewAttributeContainerWithStorageProfiler` writes an `event_data` container
through a real `StorageProfiler`, reads back the `storage-*.csv.gz` sample file, and
asserts the data size exceeds the compressed data size. It fails on the parent commit
with `AssertionError: 89 not greater than 89`.

Verified: 57 tests in `tests/storage`, 44 in the storage and profiler modules, and the
30 CLI tool tests in `tests/cli/{psteal,psort,log2timeline}_tool.py`. `pylint` and
`black` clean on both files. The one CLI failure,
`log2timeline_tool.testExtractEventsFromSourcesOnBDEImage`, reproduces identically on
the unmodified parent commit and passes when run on its own — a local environment
issue, not related to this change.

## Backward compatibility

None affected. The change is confined to which value the profiler samples; the stored
data, the storage format, and the profiler's CSV schema are all unchanged.

### DIFF
--- a/plaso/storage/sqlite/sqlite_file.py
+++ b/plaso/storage/sqlite/sqlite_file.py
@@ -302,9 +302,10 @@ class SQLiteStorageFile(sqlite_store.SQLiteAttributeContainerStore):
 
             if self.compression_format == definitions.COMPRESSION_FORMAT_ZLIB:
                 compressed_data = zlib.compress(serialized_data)
-                serialized_data = sqlite3.Binary(compressed_data)
+                column_value = sqlite3.Binary(compressed_data)
             else:
-                compressed_data = ""
+                compressed_data = b""
+                column_value = serialized_data
 
             if self._storage_profiler:
                 self._storage_profiler.Sample(
@@ -316,7 +317,7 @@ class SQLiteStorageFile(sqlite_store.SQLiteAttributeContainerStore):
                 )
 
             column_names = ["_data"]
-            values = [serialized_data]
+            values = [column_value]
 
             self._CacheAttributeContainerForWrite(
                 container.CONTAINER_TYPE, column_names, values
--- a/tests/storage/sqlite/sqlite_file.py
+++ b/tests/storage/sqlite/sqlite_file.py
@@ -1,10 +1,14 @@
 #!/usr/bin/env python3
 """Tests for the SQLite-based storage."""
 
+import csv
+import gzip
 import os
 import unittest
 
 from plaso.containers import events
+from plaso.engine import configurations
+from plaso.engine import profilers
 from plaso.lib import definitions
 from plaso.storage.sqlite import sqlite_file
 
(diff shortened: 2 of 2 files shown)


joachimmetz (approved):
LGTM

Issue #5183, opened 2026-07-29
Extend syslog SSH record handling to current OpenSSH releases


kev365:
*This is a design proposal seeking maintainer buy-in before a PR, per the contribution process. No
code is attached yet.*

## Motivation

SSH login records are usually the first place an examiner looks on a Linux host: who connected, from
where, with which credential, and which attempts did not succeed. Plaso already has event data for
this (`syslog:ssh:login`, `syslog:ssh:failed_connection`, `syslog:ssh:opened_connection`) and a
grammar behind it.

OpenSSH has changed the way it writes these records since that support was added, so I would like to
extend the existing handling to cover the current form.

## What I tested against

Two machines I set up for this, chosen because they sit either side of the OpenSSH change described
below. Both are stock installs; the logs are unmodified.

| | Operating system | OpenSSH | Log file |
|---|---|---|---|
| Machine A | Ubuntu 26.04 | 10.2p1 (current) | `/var/log/auth.log` |
| Machine B | Ubuntu 24.04 | 9.6p1 (before the change) | `/var/log/auth.log` |

Running the syslog plugin over both, using plaso 20260720:

| | SSH records in the file | producing `syslog:ssh:*` |
|---|---|---|
| Machine A | 1,235 | 0 |
| Machine B | 105 | 1 |

Every line was recognised as a syslog line, so nothing is malformed. Records the SSH rules do not
match take the `syslog:line` path instead.

The rotated files on machine A (`auth.log.1`, and `auth.log.2.gz`) behave the same way, with a
further 427 records written by the per-session program, so this is not something that only affects
the current file.

### 1. The program name that writes the record

OpenSSH 9.8 split the server into two programs, and logins are now written by the second one:

> *"the server has been split into a listener binary, sshd(8), and a per-session binary
> 'sshd-session'"*
> — [OpenSSH 9.8 release notes](https://www.openssh.com/releasenotes.html)

The plugin selects the SSH handling when the reporter is exactly `sshd`
(`plaso/parsers/text_plugins/syslog.py:551`, and `:834` for the traditional format). On machine A,
1,216 of the 1,235 SSH records are written by **`sshd-session`**, so they take the general path.

The original name is still used — `sshd` writes the listener's own messages, such as
`Server listening on 0.0.0.0 port 22.` — so **`sshd-session`** would need to be recognised in
addition to `sshd`, not instead of it. A Fedora 44 machine logs `sshd-session` as well, so this is
not limited to one distribution family.

[...]

PR #5193, opened 2026-08-02
Extend syslog SSH record handling to current OpenSSH releases #5183

kev365:
Implements #5183.

OpenSSH 9.8 split the server into a listener binary, `sshd`, and a per-session binary,
`sshd-session`, which writes the authentication messages. This recognises both reporter names and
extends three of the sshd structures to the message forms current releases write.

### The per-session reporter name

`_SSHD_REPORTERS` holds both names and the dispatch in both plugins uses it. `sshd` is kept because
the listener still writes its own messages, such as `Server listening on 0.0.0.0 port 22.`

### Key fingerprints

The existing structure matches a key type followed by a colon separated hexadecimal digest, which is
the MD5 form. `sshd_config(5)` gives the default for `FingerprintHash` as `sha256`, so the value now
written is the hash name followed by base64, which can contain `+` and `/`. Both forms are matched.

Note the hexadecimal form is tried first: the hash name pattern would otherwise match `00:aa` of
`00:aa:bb:cc` and leave the remainder unparsed.

### The ssh2 suffix on unsuccessful authentication

`_SSHD_LOGIN` already expects `ssh2` after the port number. `_SSHD_FAILED_CONNECTION` ended at the
port, so it now accepts the same optional suffix. `test_data/syslog/syslog_ssh.log:6` is a record in
that form and has been in the repository since the file was added.

### User names

`_SSH_USERNAME` matched letters and digits. It is now determined by the text preceding ` from `,
since sshd logs the name as provided by the client, which is not limited to the characters
`useradd` accepts. Service and directory accounts commonly contain a hyphen, underscore or dot.

### Measured effect

Two `/var/log/auth.log` files from hosts either side of the OpenSSH 9.8 change, run through
`log2timeline.py` and `pinfo.py`:

| | before | after |
|---|---|---|
| Ubuntu 26.04, OpenSSH 10.2p1 | `syslog:line` 3,406 · `syslog:ssh:*` 0 | `syslog:line` 3,160 · `login` 240 · `failed_connection` 6 |
| Ubuntu 24.04, OpenSSH 9.6p1 | `syslog:line` 242 · `login` 1 | `syslog:line` 221 · `login` 17 · `failed_connection` 5 |

**Total event count is unchanged** in both cases, 3,409 and 246. No new timeline events are
produced; records that previously took the `syslog:line` path now take the SSH one.

Four tagging rules in `plaso/data/tag_linux.txt` that look for a reporter of `sshd` also apply again
on hosts running the current release.

### Backward compatibility

No attribute is removed or renamed and no new data type is added; the three existing SSH data types
are reused, so their formatter and timeliner entries are unchanged. The colon separated hexadecimal
fingerprint form and the unsuffixed unsuccessful form both still parse, and
`tests/parsers/text_plugins/syslog.py` asserts on records of each.

The verification grammar is untouched, so the plugins claim exactly the same files.

### Test data

`test_data/syslog/syslog_sshd_session.log` — eight records from a Fedora 44 host running OpenSSH
10.2p1, covering the per-session reporter, a base64 fingerprint, user names containing a hyphen, a
dot and an underscore, unsuccessful authentication with the `ssh2` suffix, and two messages the
structures do not define, which are retained as `syslog:line`.

`testProcessSshd` now also asserts on the two records of `syslog_ssh.log` described above, which
were counted but not previously inspected.

### Points left open in #5183

These are the questions from the issue. I have taken the smallest option in each case and am happy
to change any of them.

- **The fingerprint attribute** is kept as a single value, matching how it is stored today, rather
  than splitting out the key type and hash name.
- **Unsuccessful authentication for accounts that do not exist**, written as `Failed password for
  invalid user …`, is not covered here; it fits with the group of messages I offered to raise
  separately.
- **The reporter name** is handled by an alternation rather than by normalising the name first,
  since normalising would change how every reporter is dispatched.
- **No extraction warning** is produced when an SSH message does not match. On the hosts I measured,
  around 40% of the messages are ones the structures do not define at all, so a warning for each
  would be a large volume for something that is not an error.

## Checklist:
* [x] Reviewer assigned.

### DIFF
--- a/plaso/parsers/text_plugins/syslog.py
+++ b/plaso/parsers/text_plugins/syslog.py
@@ -155,15 +155,35 @@ class BaseSyslogTextPlugin(interface.TextPlugin):
 
     _CRON_MESSAGE = pyparsing.Group(_CRON_TASK_RUN).set_results_name("task_run")
 
+    # OpenSSH 9.8 split the server into a listener binary, sshd, and a per-session
+    # binary, sshd-session, which writes the authentication messages.
+    _SSHD_REPORTERS = frozenset(["sshd", "sshd-session"])
+
     _SSHD_AUTHENTICATION_METHOD = pyparsing.Keyword("password") | pyparsing.Keyword(
         "publickey"
     )
 
-    _SSHD_FINGER_PRINT = pyparsing.Combine(
-        pyparsing.Literal("RSA ") + pyparsing.Word(":" + pyparsing.hexnums)
+    # A key fingerprint consists of the key type followed by the digest, where the
+    # digest is either the hash name and a base64 value, such as
+    # "ED25519 SHA256:5xyQ+PG1Z3CIiShclJ2iNya5TOdKDgE/HrOXr21IdOo", or the older
+    # colon separated hexadecimal form, such as "RSA 00:aa:bb:cc". The default of
+    # the sshd_config FingerprintHash option is sha256.
+    # Note that the hexadecimal form is matched first, since the hash name and
+    # base64 pattern would otherwise match "00:aa" of "00:aa:bb:cc" and leave the
+    # remainder of the value unparsed.
+    _SSHD_FINGER_PRINT = pyparsing.Regex(
+        r"[A-Za-z0-9-]+ "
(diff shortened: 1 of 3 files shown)


joachimmetz (approved):
LGTM

joachimmetz:
Thanks for the proposed changes.

PR #5198, opened 2026-08-20
Fix crashes in pinfo related to warning 

william-billaud:

Fix multiple little issues related to warning that create crashes when pinfo is used on storage files with some warning raised during processing.

## Description:

* Allows comparison between WarningCount instance to prevent crash with text/markdown output. Text/mardown output use a Counter with WarningCount inside, but no comparison operator were defined between WarningCount, thus making the code crash when using the Counter.most_common function that need to sort Counter object. In addition similar issue will occur if some other Counter methore are used, e.g Counter.total (__add__ is undefined)
* Fix `unable to process path specification with error: local variable 'required_format' referenced before assignment` in plist plugin. (it's more that I had the issue under my eyes will debugging so I fix it) : assign variable that was only defined in the try part of a try catch.
* Fix json output formation : strip '\n' add the end of the path_spec, and escape '\n' when rendering json.
* Define a parser chain in multiprocess extraction engine. As the parse chain was None, this cause crash when using pinfo (unsupported operator between None and string +, and fail to format the value as a string. Cf. [pinfo_tool.py#L1520](https://github.com/log2timeline/plaso/blob/09266eff829982ecdecda743df90eda59bd43626/plaso/cli/pinfo_tool.py#L1520)). It's looks like than in single process mode, no warning is generated.

## Notes:

Also bumped the minimal setuptools supported version to [77,](https://setuptools.pypa.io/en/stable/history.html#v77-0-0) which is the one that introduce support to license as string, as defined in the project.tom. Using a version below cause an error (and by error, I means impossible to run any test, a setup tool crash).

Regarding test I'm not sure how it would be the more maintainable.
1st option is to generate a plaso storage file with some warning and add some test based on this file. But this may cause issue when storage format change
2nd option is to dynamically generate a plaso storage file with some warnings, but the test may start to fail it these warning are fixed. Furthermore there is no end2end to for psort, but it may be in the scope of this PR to add support for pinfo e2e test and prevent future similar errors.

Do you have any preference?

## Checklist:
* [x] Reviewer assigned.
* [ ] Automated checks (GitHub Actions, AppVeyor) pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools>=77.0.3", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=77.0.3", "wheel"]
+requires = ["setuptools", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [project]
--- a/plaso/cli/pinfo_tool.py
+++ b/plaso/cli/pinfo_tool.py
@@ -1520,7 +1520,7 @@ class PinfoTool(tools.CLITool, tool_options.StorageFileOptions):
         for key, count_container in sorted(warnings_by_path_spec.items()):
             if write_comma:
                 self._output_writer.Write(", ")
-
+            key = key.replace('\n', '\\n')
             self._output_writer.Write(
                 f'"{key:s}": {count_container.number_of_events:d}'
             )
--- a/plaso/cli/pinfo_tool.py
+++ b/plaso/cli/pinfo_tool.py
@@ -1520,7 +1520,7 @@ class PinfoTool(tools.CLITool, tool_options.StorageFileOptions):
         for key, count_container in sorted(warnings_by_path_spec.items()):
             if write_comma:
                 self._output_writer.Write(", ")
-            key = key.replace('\n', '\\n')
+            key = key.replace("\n", "\\n")
             self._output_writer.Write(
                 f'"{key:s}": {count_container.number_of_events:d}'
             )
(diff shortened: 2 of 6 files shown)


joachimmetz:
Thanks for the proposed changes but there are various non-related alterations in here as well.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools>=77.0.3", "wheel"]
```
joachimmetz: why specifically  77.0.3 and not just  77 ?

joachimmetz:
Moving the setuptools related changes to a separate PR https://github.com/log2timeline/plaso/pull/5202

joachimmetz:
Alternate way of solving the uncaught exceptions without directly changing the attribute containers https://github.com/log2timeline/plaso/pull/5201

joachimmetz:
> Regarding test I'm not sure how it would be the more maintainable.
1st option is to generate a plaso storage file with some warning and add some test based on this file. But this may cause issue when storage format change

That could work, as long as https://github.com/log2timeline/plaso/blob/main/config/tests/generate_test_files.sh generates the test case, future storage files should introduce the same test scenario.

> Do you have any preference?

both options are fine, option 1 is easier for end-to-end tests, option 2 could be an unit test that generates a storage file with a warning

pyproject.toml
william-billaud: typo, 77 is good. 

I have reverted the change

william-billaud:
Thanks for review, 
I will take a look on the suggested alternate way add tests when time permit.
Do you prefer an issue for each bullet point in description?

joachimmetz:
> I will take a look on the suggested alternate way add tests when time permit.

Take your time, I'll likely merge the alternate approach, so the attribute container remain very basic objects.

> Do you prefer an issue for each bullet point in description?

Group the ones that logically make sense; keep other separate. The tricky bit is mostly getting the necessary context, especially when context switching a lot.

joachimmetz:
I'll close this for now, but let me know if it needs to be reopened at a later date

PR #5206, opened 2026-09-09
Added backslashreplace error handler to text file output modules #5205

### LINKED ISSUE #5205: psort raises UnicodeEncodeError on an unpaired surrogate - related to #5199
**Describe the problem:**

`psort.py -o l2tcsv` aborts the entire export the first time any event field contains an unpaired UTF-16 surrogate. `TextFileOutputModule.Open` opens the output file with strict UTF-8 and no `errors=` handler (`plaso/output/text_file.py:117`), so a single unrepresentable code point raises `UnicodeEncodeError` and the timeline is lost.

Two aspects make this worse than a clean failure:

- **The output is silently truncated, not absent.** On a real collection psort wrote 235,612 rows and *then* raised. The exit code is 1, but anyone not checking it gets a plausible-looking CSV that is missing every event after the first affected value.
- **Switching output format is not a workaround.** `TextFileOutputModule` is the shared base class for `l2tcsv`, `tln`, `dynamic` and `kml`, and `text_file.py:117` is the only `open()` call in `plaso/output/` - and the only one without an error handler.

The surrogate is emitted by the `winreg` parser via the fallback `winreg/winreg_default` plugin. In the wild it comes from the Windows MDM provisioning node cache, where a DER-encoded root CA certificate is stored as raw binary in a value typed `REG_SZ`:

```
HKLM\SOFTWARE\Microsoft\Provisioning\NodeCache\CSP\Device\MS DM Server\Nodes\<n>
    NodeUri               REG_SZ     ./Device/Vendor/MSFT/RootCATrustedCertificates/Root/<thumbprint>
    ExpectedValue         REG_SZ     <-- DER certificate as raw binary, begins 30 82 ...
    ExpectedValueType     REG_DWORD  0x2011
    ExpectedValueDataType REG_DWORD  0x6
```

The chain is:

1. Windows writes binary DER into a value declared `REG_SZ`.
2. `pyregf.get_data_as_string()` decodes `REG_SZ` as UTF-16LE permissively, so binary containing a high surrogate code unit with no following low surrogate produces a `str` holding a lone surrogate.
3. `winreg_default` copies it into the event `values` attribute.
4. `l2tcsv` writes to a strict UTF-8 file and raises.

Steps 1-3 appear defensible: the value genuinely *is* a lone surrogate, and permissive decoding is what allows the artefact to round-trip. Step 4 is where the timeline is discarded.

This is the same defect as [#5199](https://github.com/log2timeline/plaso/issues/5199), which was closed for lack of a reproducible sample and no known emitting parser - both are provided below (a self-contained reproducer that needs no attached sample, plus the emitting parser identified).

**To Reproduce:**

**Plaso version:** plaso - log2timeline version 20260720.

**Host OS:** Ubuntu 26.04 LTS, Python 3.14.4, pyregf 20260526 (as shipped in the `log2timeline/plaso:latest` container image).

**Steps to reproduce:**

First, create the source hive. In an **elevated PowerShell on any Windows machine**, this writes a `REG_SZ` value whose data contains `b3 da` - U+DAB3, an unpaired UTF-16 high surrogate - and saves an 8 KB hive:

```powershell
$s = "PLASO-5199-" + [char]0xDAB3 + "-END"
New-Item -Path HKLM:\SOFTWARE\PlasoRepro5199 -Force | Out-Null
New-ItemProperty -Path HKLM:\SOFTWARE\PlasoRepro5199 -Name ExpectedValue -PropertyType String -Value $s -Force | Out-Null
reg save HKLM\SOFTWARE\PlasoRepro5199 C:\SOFTWARE_repro /y
```

To clean up afterwards: `Remove-Item HKLM:\SOFTWARE\PlasoRepro5199 -Recurse -Force`

Then run:

```
$ log2timeline.py --status_view none --parsers winreg --storage_file r.plaso SOFTWARE_repro
[INFO] (MainProcess) PID:9 <artifact_definitions> Determined artifact definitions path: /usr/share/artifacts
Checking availability and versions of dependencies.
[OK]

Source path      : SOFTWARE_repro
Source type      : single file
Processing time  : 00:00:00

Processing started.
Processing completed.
```

`log2timeline.py` exits 0. Then:

```
$ psort.py --status_view none -o l2tcsv -w r.csv r.plaso
Processing started.
Processing completed.
Traceback (most recent call last):
...
UnicodeEncodeError: 'utf-8' codec can't encode character '\udab3' in position 115: surrogates not allowed
$ echo $?
1
```

Full traceback under **Debug output/tracebacks** below.

The output module can also be made to fail on its own, with no registry or storage file involved:

```
$ python3 -c "from plaso.output import l2t_csv; m=l2t_csv.L2TCSVOutputModule(); m.Open(path='/tmp/o.csv'); m.WriteLine('x\udab3')"
UnicodeEncodeError: 'utf-8' codec can't encode character '\udab3' in position 1: surrogates not allowed
```

Same frame (`text_file.py:135`). This is probably the shape a regression test wants.

One note for anyone writing a *parser-side* test: a `dfwinreg.fake` registry cannot reproduce this. Python's strict `utf-16-le` codec rejects unpaired surrogates, so `fake.py`'s `self._data.decode('utf-16-le')` raises `WinRegistryValueError`. Only libregf passes lone surrogates through to Python, so the parser path needs a real REGF file.

To reproduce against a full, real `SOFTWARE` hive rather than the subkey save above, import the trigger as a `.reg` on a lab VM, then acquire the hive. Generate `plaso_5199_lab.reg` with the script below. It writes the value in `hex(1)` form so the exact UTF-16LE bytes, including the unpaired surrogate, reach the registry, and saves the file as UTF-16LE with a BOM (regedit's required encoding):

```python
import sys

KEY = (r"HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Provisioning\NodeCache\CSP"
       r"\Device\MS DM Server\Nodes\99999")

SURROGATE = b"\xb3\xda"          # U+DAB3, unpaired high surrogate, LE


def hex1(data):
    """Formats bytes as a .reg hex(1) (REG_SZ) value, wrapped like regedit."""
    parts = [f"{byte:02x}" for byte in data]
    lines = []
    chunk = []
    for part in parts:
        chunk.append(part)
        if len(chunk) == 24:
            lines.append(",".join(chunk))
            chunk = []
    if chunk:
        lines.append(",".join(chunk))
    # Continuation lines need a trailing comma before the backslash; without it
    # reg import concatenates the tokens and silently discards the value.
    return ",\\\r\n  ".join(lines)


def main(path):
    # Shaped like the real value: DER SEQUENCE header, then filler, then the
    # unpaired surrogate. Contents are synthetic.
    expected = (b"\x30\x82\x01\x0a"
                + "SYNTHETIC-REPRO-PLASO-5199-".encode("utf-16-le")
                + SURROGATE
                + "-END".encode("utf-16-le")
                + b"\x00\x00")

    node_uri = ("./Device/Vendor/MSFT/RootCATrustedCertificates/Root/"
                "0000000000000000000000000000000000000000"
                "\x00").encode("utf-16-le")

    text = (
        "Windows Registry Editor Version 5.00\r\n"
        "\r\n"
        f"[{KEY}]\r\n"
        f'"NodeUri"=hex(1):{hex1(node_uri)}\r\n'
        f'"ExpectedValue"=hex(1):{hex1(expected)}\r\n'
        '"ExpectedValueType"=dword:00002011\r\n'
        '"ExpectedValueDataType"=dword:00000006\r\n'
    )

    # regedit expects UTF-16LE with a BOM.
    with open(path, "w", encoding="utf-16-le", newline="") as file_object:
        file_object.write("\ufeff" + text)

    print(f"wrote {path}")
    print(f"ExpectedValue byte length : {len(expected)}")
    print(f"surrogate bytes at offset : {expected.find(SURROGATE)}")


if __name__ == "__main__":
    main(sys.argv[1] if len(sys.argv) > 1 else "plaso_5199_lab.reg")
```

`ExpectedValue` is a DER prefix `30 82 01 0a`, then `SYNTHETIC-REPRO-PLASO-5199-` as UTF-16LE, then `b3 da` (U+DAB3), then `-END`, then a NUL - mirroring the field artefact (a DER certificate stored in a value typed `REG_SZ`). Then, elevated on a lab VM:

```
python3 make_lab_reg.py plaso_5199_lab.reg
reg import plaso_5199_lab.reg
```

Acquire the hive (`C:\Windows\System32\config\SOFTWARE`)

```
log2timeline.py --status_view none --parsers winreg --storage_file r.plaso SOFTWARE
psort.py --status_view none -o l2tcsv -w r.csv r.plaso
```

Measured on a collector-acquired hive: log2timeline exits 0; psort exits 1 with `\udab3` at position 309, having written 235,612 rows to a truncated CSV before it died (the row count is specific to the acquired hive; the column position depends on the value, so both differ from the subkey-save run above). With `errors="backslashreplace"`: psort exits 0 and writes the complete timeline, the value present as `ExpectedValue: [REG_SZ] ...SYNTHETIC-REPRO-PLASO-5199-\udab3-END`.

An acquired `SOFTWARE` hive carries the machine's Windows product key and DigitalProductId blobs, so do not attach one to the issue.

**Source data:** No sample file is required - the PowerShell snippet above generates the source hive locally. It is entirely synthetic (the value is ASCII filler around one surrogate) and contains no real data. The original observation was on a Windows 10 `C:\Windows\System32\config\SOFTWARE` hive from a client triage collection, which cannot be shared.

**How Plaso was installed:** the official `log2timeline/plaso:latest` Docker image (image ID `a8d42e64dcd3`).

**Expected behavior:**

`psort.py` should write the complete timeline and exit 0, representing code points that cannot be encoded rather than discarding the export. Given the surrogate is genuine data recovered from the artefact, escaping it preserves the information; aborting loses everything else.

**Debug output/tracebacks:**

```
Traceback (most recent call last):
  File "/usr/bin/psort.py", line 8, in <module>
    sys.exit(Main())
             ~~~~^^
  File "/usr/lib/python3/dist-packages/plaso/scripts/psort.py", line 81, in Main
    tool.ProcessStorage()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3/dist-packages/plaso/cli/psort_tool.py", line 551, in ProcessStorage
    output_engine.ExportEvents(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^
        storage_reader,
        ^^^^^^^^^^^^^^^
    ...<6 lines>...
        use_time_slicer=self._use_time_slicer,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3/dist-packages/plaso/multi_process/output_engine.py", line 567, in ExportEvents
    self._ExportEvents(
    ~~~~~~~~~~~~~~~~~~^
        storage_reader,
        ^^^^^^^^^^^^^^^
    ...<4 lines>...
        use_time_slicer=use_time_slicer,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3/dist-packages/plaso/multi_process/output_engine.py", line 336, in _ExportEvents
    self._FlushExportBuffer(storage_reader, output_module)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/plaso/multi_process/output_engine.py", line 419, in _FlushExportBuffer
    output_module.WriteFieldValuesOfMACBGroup(self._output_mediator, macb_group)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/plaso/output/l2t_csv.py", line 351, in WriteFieldValuesOfMACBGroup
    self.WriteLine(output_text)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/plaso/output/text_file.py", line 135, in WriteLine
    self._file_object.write(f"{text:s}\n")
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
UnicodeEncodeError: 'utf-8' codec can't encode character '\udab3' in position 115: surrogates not allowed
```

**Additional context**

*Suggested fix.* Add an overridable error handler and thread it into the `open()` call:

```diff
--- a/plaso/output/text_file.py
+++ b/plaso/output/text_file.py
@@ -65,6 +65,7 @@
     WRITES_OUTPUT_FILE = True
 
     _ENCODING = "utf-8"
+    _ERRORS = "backslashreplace"
 
     def __init__(self):
         """Initializes an output module that writes to a text file."""
@@ -114,7 +115,8 @@
             )
 
         # pylint: disable=consider-using-with
-        self._file_object = open(path, "wt", encoding=self._ENCODING)
+        self._file_object = open(
+            path, "wt", encoding=self._ENCODING, errors=self._ERRORS)
```

`backslashreplace` rather than `replace`, for three reasons:

- It is lossless. `\udab3` in the CSV still identifies the original code point, whereas `replace` collapses everything to U+FFFD.
- It is already the convention in this codebase: 32 occurrences across 18 files, and zero uses of `errors="replace"`.
- `plaso/parsers/ntfs.py:443` is a near-exact precedent - same decode, same handler, plus a warning:

  ```python
  name_string = ucs2_stream.decode("utf-16-le", errors="backslashreplace")
  parser_mediator.ProduceWarning(
      f"unable to decode USN record name string with error: "
      f"{exception!s}. Unsupported code points are escaped.")
  ```

With the patch applied, `psort.py` exits 0, the full timeline is written, and the value appears as:

```
ExpectedValue: [REG_SZ] PLASO-5199-\udab3-END
```

Emitting a warning when the handler engages, mirroring `ntfs.py`, would make the escaping visible rather than silent, though that is secondary to not discarding the timeline.

*How common this is.* A high surrogate is any code unit in D800-DBFF, i.e. 4/256 of arbitrary 16-bit values, and it is unpaired unless the next unit falls in DC00-DFFF - so p ≈ 0.0154 per code unit, roughly 1 in 65. For a ~1200-byte DER certificate read as UTF-16LE that is ~9 expected lone surrogates, and P(at least one) is effectively 1. Empirically, 2000 random 1200-byte blobs gave a mean of 9.2 lone surrogates and 100% contained at least one.

Field data agrees. On a Windows 10 triage collection processed with `--parsers win7,!filestat`, of 1,528,416 extracted `event_data` containers exactly 3 contained a lone surrogate - all three `winreg/winreg_default` on the same `SOFTWARE` hive, in three adjacent NodeCache nodes, carrying 10, 5 and 9 surrogates respectively. Three malformed registry values out of 1.5 million events discarded the entire timeline.

Isolation was confirmed three independent ways on that collection: running each of the 28 parsers in the `win7` preset alone (only `winreg` fails, 27 exit 0); running each of the 11 registry hives alone (only `SOFTWARE` fails, 10 exit 0); and scanning all 1,528,416 `event_data` containers in the storage file for surrogates by parser and attribute.

Happy to open a PR with the change and a test if that would help.

J-a-s-o-n-P:
Fixes #5205.

## What this fixes

`psort` aborts with `UnicodeEncodeError: 'utf-8' codec can't encode character '\udab3' ... surrogates not allowed` and produces no usable timeline whenever an event field contains a lone UTF-16 surrogate. Several Windows string types are UCS-2 rather than UTF-16, so unpaired surrogates are legitimate values, for example a `REG_SZ` under the MDM provisioning `NodeCache` that actually holds a DER certificate. `pyregf` decodes such a value to a Python string containing a lone surrogate, and strict UTF-8 cannot encode it, so the whole export dies on the first affected row. The full reproducer is in #5205.

## Root cause

`TextFileOutputModule` in `plaso/output/text_file.py` opens its output file with `encoding="utf-8"` and the default `errors="strict"`, so the first un-encodable code point in any field raises and takes down the entire write. This class is the shared base for the `l2tcsv`, `tln`, `dynamic` and `kml` output modules, so switching text output format does not avoid it.

## The fix

Add an overridable `_ERRORS = "backslashreplace"` on `TextFileOutputModule` and pass `errors=self._ERRORS` to the `open()` call. An un-encodable code point is escaped to its `\uXXXX` form instead of aborting the write. `backslashreplace` is lossless (the original value is recoverable from the escape), it is the handler the plaso style guide recommends for this case, and it matches existing use in the codebase (for example `plaso/parsers/ntfs.py`). It is deliberately not `errors="replace"`, which is lossy. Behaviour for data that already encodes cleanly is unchanged.

## What was tested to prove it

Unit:

- New regression test `testWriteLineWithUnpairedSurrogate` in `tests/output/text_file.py`. It writes `"x\udab3"` through the module's real `Open()` / `WriteLine` path (against a real file, so it exercises the strict-encoding write that used to raise) and asserts the file reads back `x\udab3\n`. It fails on current main with the `UnicodeEncodeError` and passes with this change.
- Full `tests/output` suite: 90 passed, 9 skipped. `pylint 4.0.5 --rcfile=.pylintrc`, `black --check` and `docformatter --check` are all clean.

End to end, against a real Windows `SOFTWARE` registry hive carrying the trigger value from the #5205 reproducer, same codebase before and after:

- Unpatched main: `psort -o l2tcsv` exits 1 with the `UnicodeEncodeError` and writes a truncated CSV (235,572 rows), stopping at the un-encodable value.
- With this change: `psort` exits 0 and writes the complete timeline (235,575 rows), with the offending value preserved as a backslash escape and recoverable from it.
- `log2timeline` exits 0 in both runs; the crash and the fix are entirely in the psort output-writing step.


joachimmetz (approved):
LGTM

joachimmetz:
Thanks for making these changes

PR #5207, opened 2026-09-12
Changes to Linux tagging rules to also match the sshd-session reporter #5183

kev365:
Follows #5193, for #5183.

OpenSSH 9.8 split the server into a listener binary, `sshd`, and a per-session binary,
`sshd-session`, which writes the session and authentication messages. #5193 recognises both names
in the syslog plugins. In that change I expected the four rules in `plaso/data/tag_linux.txt` that
select on a reporter of `sshd` to apply again as a result. They select on the reporter value stored
on the event, which is the name as logged, so the rules need the per-session name as well. This
adds it to the four rules:

| Label | Message text the rule selects on |
|---|---|
| `login` | `session opened`, `Starting session` |
| `login_failed` | `uthentication fail` |
| `login_failed` | `Access denied for user`, `not allowed because` |
| `logout` | `session closed`, `Close session` |

`tests/data/tag_linux.py` now covers both reporter names for each rule. The third rule had no test;
one is added.

### Measured effect

Events extracted by the syslog plugin from the two `/var/log/auth.log` files measured in #5193,
tagged with `tag_linux.txt`:

| | Label | before | after |
|---|---|---|---|
| Ubuntu 26.04, OpenSSH 10.2p1 | `login` | 5 | 245 |
| | `login_failed` | 8 | 14 |
| | `logout` | 241 | 479 |
| Ubuntu 24.04, OpenSSH 9.6p1 | `login` | 18 | 18 |
| | `login_failed` | 6 | 6 |
| | `logout` | 32 | 32 |

The additional events on the 26.04 host are all `sshd-session` records: 240 `session opened`, 6
`authentication failure` and 238 `session closed`. The 24.04 host, where the reporter is `sshd`, is
unchanged, as are the other labels on both hosts.

### Backward compatibility

Only the four rules are extended; each still matches everything it matched before. No parser, data
type or attribute is involved.

## Checklist:
* [x] Reviewer assigned.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plaso/data/tag_linux.txt
+++ b/plaso/data/tag_linux.txt
@@ -14,7 +14,7 @@ login
   data_type is 'linux:utmp:event' AND login_type == 7
   data_type is 'selinux:line' AND audit_type is 'LOGIN'
   reporter is 'login' AND (message_body contains 'logged in' OR message_body contains 'ROOT LOGIN' OR message_body contains 'session opened')
-  reporter is 'sshd' AND (message_body contains 'session opened' OR message_body contains 'Starting session')
+  (reporter is 'sshd' OR reporter is 'sshd-session') AND (message_body contains 'session opened' OR message_body contains 'Starting session')
   reporter is 'dovecot' AND message_body contains 'imap-login: Login:'
   reporter is 'postfix/submission/smtpd' AND message_body contains 'sasl_'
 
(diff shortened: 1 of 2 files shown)


joachimmetz (approved):
LGTM

joachimmetz:

PR #5209, opened 2026-09-12
Added support for AVC records to the SELinux log file parser #5156

kev365:
Adds support for AVC records to `text_plugins/selinux.py`, as proposed in #5156.

### Why

An SELinux AVC record produced no structured attributes at all. Its message body starts with an
access vector decision rather than a field, so the key value parsing stopped at the first token and
nothing was extracted, not even `pid` or `comm`. From `test_data/selinux.log`:

```
type=AVC msg=audit(1243332701.744:101): avc: denied { getattr } for pid=2714 comm="ls" path="/usr/lib/locale/locale-archive" dev=dm-0 ino=353593 scontext=system_u:object_r:unlabeled_t:s0 tcontext=system_u:object_r:locale_t:s0 tclass=file
```

AppArmor records of the same type were unaffected, since those start with `apparmor="DENIED"` and
parse as fields. That record now yields the process, the file, both security contexts, the object
class, the requested permission and the decision.

The record layout is the kernel's, in
[`security/selinux/avc.c`](https://github.com/torvalds/linux/blob/master/security/selinux/avc.c):
`avc:  <denied|granted> `, the requested permissions in `{ … }`, the fields, and, on denials only,
`permissive=<0|1>`. Records from kernels that predate the `permissive` field, such as the one in
`test_data/selinux.log`, leave `permissive_mode` unset.

### Added attributes

| attribute | source | note |
|---|---|---|
| `access_granted` (bool) | `denied` / `granted` | the access vector decision |
| `access_permissions` (list[str]) | `{ … }` | the permissions that were requested |
| `permissive_mode` (bool) | `permissive` | the access was allowed even though the policy denied it |
| `target_object_class` (str) | `tclass` | such as `file`, `process` or `capability` |
| `target_security_context` (str) | `tcontext` | security context of the object |

`scontext` is stored in the existing `security_context`, since it holds the security context of the
process, the same as `subj` on other record types. `path` is stored in the existing `file_path`,
alongside the `name` field that was already used.

`permissive_mode` is included because it changes what a `denied` record means: with the policy in
permissive mode the operation went ahead. Without it, a log from such a system reads as though every
denial was enforced.

### Notes

- The decision and the permissions are parsed with a small grammar for the
  `avc: <decision> { <permissions> } for` prefix; the remainder of the body is parsed as fields by
  the existing code, so no other record type is affected. A body that starts with `avc:` but cannot
  be parsed as a decision falls back to being parsed as fields, and is covered by a test.
- Only `denied` decisions and single permissions appear in the logs I captured, and I found no
  `granted` records, since those require an `auditallow` rule. The permissions are stored as a list
  so that a record with several is handled, but I have no specimen that proves it.
- The AppArmor `apparmor="DENIED"` records use a different vocabulary and are left as they are.

### Backward compatibility

Five attributes are added to the existing `selinux:line` data type; no attribute is removed or
renamed and no data type is added. The formatter and timeliner entries are unchanged. Records of
every other `type=` parse exactly as before, and the verification grammar is untouched, so the plugin
claims the same files.

### Test data

New `test_data/audit_avc.log` holds six real AVC records captured from a Fedora 44 system with
SELinux in enforcing mode, covering `tclass` of `file`, `process` and `capability`, both
`permissive=0` and `permissive=1`, and both the `name` and `path` fields. The existing AVC record in
`test_data/selinux.log` is now asserted as well, and a truncated AVC record was added to
`test_data/audit_corrupted.log`.

Verified against current `main` with GIFT-PPA dependencies matching CI: unit tests, `psteal` with no
extraction warnings, `tests.cli.psteal_tool`, and the full end-to-end suite including the
`ext4_with_binaries` over-claim guard.

## Checklist:
* [x] Reviewer assigned.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


joachimmetz (approved):
LGTM

joachimmetz:

